PR #2, opened 2022-01-11
Update README.md for mispelling

debussyman:


paulgb:
Good catch, thank you Ryan!

PR #19, opened 2022-03-29
add spawn_owner to Updating SLAB state message

rolyatmax:


crates/core/src/lib.rs
```diff
@@ -431,8 +431,15 @@ impl SessionLivedBackend {
             duration.num_milliseconds().to_string()
         }).unwrap_or("".to_string());
 
+        let spawn_owner: String = self.metadata.labels
```
paulgb, line 434: It's probably worth exposing this as a `.owner()` method on the `SessionLivedBackend` itself. It would return an `Option<String>`, which would also simplify all the `unwrap_or()` chains because you could use the `?` operator. (Plus we will probably want to use it in other places, so it makes that easy)

paulgb: (There would still be one `unwrap_or` in this call, because it would still be returning an option and we want to log a string.)

rolyatmax: Ohh good call!

rolyatmax: Updated!

crates/core/src/lib.rs
```diff
@@ -284,6 +284,11 @@ impl SessionLivedBackend {
         }
     }
 
+    pub fn owner(&self) -> Option<String> {
+        let labels = self.metadata.labels.as_ref()?;
+        labels.get("spawn-owner").map(|l| l.to_string())
```
paulgb, line 289: If the type of `l` is `&String` (I'm not 100% but I think it is), the `.map` call could be replaced by `.clone()`. Otherwise, looks good!

rolyatmax: Updated

paulgb (approved):

PR #26, opened 2022-04-06
Refactor to move API library into core

paulgb:


crates/controller/src/main.rs
```diff
@@ -193,8 +189,8 @@ async fn reconcile(
                         SessionLivedBackendStatus::default(),
                     )
                     .await?;
-                },
-                _ => ()
+                }
```
rolyatmax, line 192: Oh interesting - how does this behave without the `,`?

crates/controller/src/main.rs
```diff
@@ -91,7 +91,7 @@ fn get_application_state(pod: &Pod) -> Option<ContainerState> {
 
             if let Some(terminated) = state.terminated.as_ref() {
                 if terminated.reason.as_deref() == Some("Error") {
-                    return Some(ContainerState::Failed)
+                    return Some(ContainerState::Failed);
```
rolyatmax, line 94: Just curious: does this produce a change in behavior or are semicolons optional on `return` expressions?

paulgb: They're optional, but the `cargo clippy` and `cargo fmt` prefer the semicolon in this case.

crates/controller/src/main.rs
paulgb: The `,` in a `match` statement can be omitted after a `}` or after the last element. It doesn't change the meaning. In this case, `cargo fmt` decided to remove it.

rolyatmax (approved):
LGTM!

PR #51, opened 2022-06-14
config to switch linker to mold for linux target

pretentious7:
PROBLEM:
dev hot-rebuilds currently take ~10s on a 4c8t intel laptop
from circa 2017. (rustc 1.61.0)

SOLUTION:
After switching to mold, this time drops to ~3.5s.

FURTHER INFORMATION:
Profiling via rustc --self-profile reveals that most of the time
for a hot reload comes from linking. mold is a faster linker than
lld or ld.

POSSIBLE IMPROVEMENTS:
Switching to rust nightly (currently 1.63.0) yield further 20-30%
improvements.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,3 @@
+[target.x86_64-unknown-linux-gnu]
+linker = "/usr/bin/clang"
+rustflags = ["-C", "link-arg=--ld-path=/usr/bin/mold"]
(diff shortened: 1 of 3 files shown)


.cargo/config.toml
```diff
@@ -0,0 +1,3 @@
+[target.x86_64-unknown-linux-gnu]
+linker = "/usr/bin/clang"
+rustflags = ["-C", "link-arg=--ld-path=/usr/local/bin/mold"]
```
paulgb, line 3: Should this be:

```
rustflags = ["-C", "link-arg=-fuse-ld=/path/to/mold"]
```

pretentious7: yea, it's deprecated, but is fine for now. esp considering stuff like this
https://github.com/rust-lang/rust/issues/94347

pretentious7:
OK that should be good. In the future, or if we decide to switch to nightly, something like this might be worthwhile: https://github.com/dimensionhq/fleet

PR #60, opened 2022-06-15
Fix fast build

pretentious7:
PROBLEM:
as mentioned in 86054, hot-reloads take around 10s, mostly due to
linker.

SOLUTION:
use mold, this time via a fast_build script that sets the necessary
cargo env vars, ensures that mold and clang are installed, and runs
cargo build with the correct path for the mold linker.

POSSIBLE IMPROVEMENTS:
may be a good idea to switch the shell script for a zx script. Also
would be nice if `cargo build` automatically ran fast_build.sh as a
wrapper if the host and profile are correct (ie. linux host, with
dev profile)

### DIFF
--- a/fast_build.sh
+++ b/fast_build.sh
@@ -1,0 +1,28 @@
+#!/bin/env bash
+set -reuo pipefail
+IFS=$'\n\t'
+trap '__error_handling__' ERR
+moldPath=1
+clangPath=1
+
+function __error_handling__() {
+  if [[ $moldPath == "" ]]
+  then
+    >&2 echo "mold linker is not installed, please install it"
+  fi
+
+  if [[ $clangPath == "" ]]
+  then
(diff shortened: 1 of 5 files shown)


fast_build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if [[ $HALT -ne 0 ]]
+then
+  exit 1
+fi
+
+export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_LINKER="clang"
+export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUSTFLAGS="-C link-arg=-fuse-ld=${moldPath}"
+
+cargo build
```
paulgb, line 29: Let’s pass —features=full here too

paulgb (commented):
Left one comment, can you also remove .cargo from .dockerignore? Otherwise looks good

PR #69, opened 2022-06-18
feat: basic tail on backend logs

lalo:
Not sure about the specifics but played around with #48. First time using rust so I might be doing some non-idiomatic things. Feel free to decline.

- add optional local docker unit test (hidden via flag since docker might not be installed)
- add simple unit test using hello-world docker image
- add get_last_n_log_lines
- add Destroy EventType
- stop container only if its running
- return actual container name/id via run_container(..)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,3 +37,4 @@ required-features = ["full"]
 
 [features]
 full = []
+docker-unit-test = []
--- a/src/drone/agent/docker.rs
+++ b/src/drone/agent/docker.rs
@@ -2,7 +2,9 @@ use super::DockerOptions;
 use anyhow::{anyhow, Result};
 use bollard::{
     auth::DockerCredentials,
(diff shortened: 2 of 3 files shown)


paulgb:
Thanks Eduardo, I appreciate the initiative.

A couple of high-level comments:

- This approach is similar to the approach we took for logs in the Kubernetes version, and it's a valid approach, but I've been thinking that since we now have NATS Jetstream in the stack, we could just stream all logs to Jetstream and then use NATS to access them as needed. One upside to this is that we can use Jetstream to persist the logs temporarily even after the backend is removed. The tricky part is how we resume log collection for a given backend after a drone restart, without losing or double-counting log entries. Maybe we could just use Jetstream deduplication and a large replay window for this?

- We've been using separate Typescript tests in tests/ for tests that are not self-contained, because some of them require logic around how resources are spun up/down. I'd rather not fragment those tests between the rust/typescript codebase unless/until we have a reason not to.

PR #75, opened 2022-06-29
pass in nats subject when init_tracing_with_nats

rolyatmax:
Passing in the NATS subject to publish logs to allows applications using `init_tracing_with_nats` (like the Jamsocket platform) to publish to something like `logs.platform` instead of `logs.drone`.

### DIFF
--- a/src/logging.rs
+++ b/src/logging.rs
@@ -14,9 +14,9 @@ use tracing_subscriber::{layer::Context, util::SubscriberInitExt, EnvFilter, Lay
 const TRACE_STACKDRIVER: &str = "TRACE_STACKDRIVER";
 const LOG_DEFAULT: &str = "info,sqlx=warn";
 
-async fn do_logs(nc: &TypedNats, mut recv: tokio::sync::mpsc::Receiver<LogMessage>) {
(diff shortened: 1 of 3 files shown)


src/logging.rs
```diff
@@ -38,7 +38,7 @@ pub fn init_tracing() -> Result<()> {
     Ok(())
 }
 
-pub fn init_tracing_with_nats(nats: TypedNats) -> Result<()> {
+pub fn init_tracing_with_nats(nats: TypedNats, subject: String) -> Result<()> {
```
paulgb, line 41: nit: &str rather than String.

paulgb (approved):

src/logging.rs
rolyatmax: I did a `String` here so that I could avoid lifetime issues in the `tokio::spawn` below. I could also stick `let subject = subject.to_string();` at the top of this function, too, though? Are there other ways of dealing with this?

paulgb: Ah, right. This is fine, then. LGTM.

paulgb (approved):

PR #76, opened 2022-06-30
Log backend metadata

paulgb:
This logs fields provided in the `metadata` of a backend, both to stdout logs and over NATS.

### DIFF
--- a/src/drone/agent/executor.rs
+++ b/src/drone/agent/executor.rs
@@ -9,6 +9,7 @@ use crate::{
 use anyhow::{anyhow, Result};
 use chrono::Utc;
 use dashmap::DashMap;
+use serde_json::json;
 use std::{fmt::Debug, net::IpAddr, sync::Arc};
 use tokio::{
     sync::mpsc::{channel, Sender},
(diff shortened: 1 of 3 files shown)


rolyatmax (approved):

src/drone/agent/executor.rs
```diff
@@ -89,7 +90,8 @@ impl Executor {
         let _span = tracing::span!(
             Level::INFO,
             "run_backend",
-            backend_id = spawn_request.backend_id.id()
+            backend_id = spawn_request.backend_id.id(),
+            metadata = %json!(spawn_request.metadata),
```
rolyatmax, line 94: We need to pass this to the `Executing state.` log below, too.

paulgb: Executing state runs while we have entered this span, so it is automatically added.

rolyatmax: According to [these docs](https://docs.rs/tracing/latest/tracing/struct.Span.html#in-asynchronous-code) it sounds like this span might be open while other async tasks are running, causing that metadata to incorrectly get applied to other log messages. Am I understanding that correctly?

paulgb:
Re. elapsed time, I agree with the db strategy. This is related to being able to resume backends when a drone restarts, so I'm working on the drone resume first.

PR #78, opened 2022-06-30
Pass metadata directly in the event rather than using a span.

paulgb:


rolyatmax (approved):
Last commit looks good to me! (I think the commits from #77 accidentally slipped in here - we should rip those out before landing.)

rolyatmax (approved):

PR #82, opened 2022-07-06
Use a mutex to guard initial DB migration

paulgb:
It's possible for two paths to try creating the db at the same time, which seems to be a source of test flakiness. This PR applies the same approach we use for sharing the NATS connection to sharing the database pool.


pretentious7 (approved):
makes sense :+1:

PR #94, opened 2022-07-15
Update async nats

paulgb:


src/drone/cli.rs
```diff
@@ -257,9 +253,7 @@ impl From<Opts> for DronePlan {
                     None
                 };
 
-                if proxy_options.is_none() && agent_options.is_none() {
-                    panic!("Expected at least one of --proxy, --agent, --certloop if `serve` is provided explicitly.");
-                }
+                assert!(!(proxy_options.is_none() && agent_options.is_none()), "Expected at least one of --proxy, --agent, --certloop if `serve` is provided explicitly.");
```
rolyatmax, line 256: Nit: would this be better rewritten as `proxy_options.is_some() || agent_options.is_some()`?

rolyatmax (approved):
LGTM aside from minor nit

src/drone/cli.rs
paulgb: Good call.

PR #96, opened 2022-07-18
Add log context to certificate refresh code

paulgb:


rolyatmax (approved):
Aside from the questions I left, this LGTM.

src/drone/proxy/certs.rs
```diff
@@ -75,7 +75,9 @@ impl CertRefresher {
         };
 
         for path in key_cert_path_pair.parent_paths() {
-            watcher.watch(&path, RecursiveMode::NonRecursive)?;
+            watcher
+                .watch(&path, RecursiveMode::NonRecursive)
+                .context("Error installing watcher.")?;
```
rolyatmax, line 80: Where exactly do these `context` strings go when the `?` "fails"?

src/drone/proxy/certs.rs
```diff
@@ -1,5 +1,5 @@
 use crate::keys::{load_certs, load_private_key, KeyCertPathPair};
-use anyhow::Result;
+use anyhow::{Context, Result};
```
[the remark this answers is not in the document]
paulgb: Yes, in Rust you need to import a trait to use its methods (in this case `context`).

src/drone/proxy/certs.rs
paulgb: When an error is reported (technically, when its debug formatter is called), the context message is added to the output along with the original error.

PR #97, opened 2022-07-23
Abhi/dis 428 spawner pipe docker stats output to nats

pretentious7:
Realized I didn't put this in as a pr yet, modeled it off the way logs work. Very unsure about both exact design and the stats being passed, as well as the test. But its a start!

### DIFF
--- a/src/drone/agent/executor.rs
+++ b/src/drone/agent/executor.rs
@@ -2,7 +2,9 @@ use super::docker::{ContainerEventType, DockerInterface};
 use crate::{
     database::{Backend, DroneDatabase},
     drone::agent::wait_port_ready,
-    messages::agent::{BackendState, BackendStateMessage, DroneLogMessage, SpawnRequest},
+    messages::agent::{
+        BackendState, BackendStateMessage, DroneLogMessage, DroneStatsMessage, SpawnRequest,
+    },
     nats::TypedNats,
     types::BackendId,
 };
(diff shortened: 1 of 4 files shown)


tests/src/test-agent.ts
```diff
@@ -146,14 +146,55 @@ test("Spawn with agent", async (t) => {
...
+    metadata: {},
+  }
+  expectResponse(t, nats, "drone.1.spawn", request, true)
+
+  const statsStatusSubsription = 
+    new NatsMessageIterator<unknown>(
+      nats.subscribe(`backend.${backendId}.stats`, { timeout : 10000 })
+  )
+  t.deepEqual(["cpu_used", "mem_used", "disk_used"],
+    Object.keys((await statsStatusSubsription.next())[0]))
```
paulgb, line 193: This enforces a key order which is stricter than we care about (for the purposes of our interfaces, objects should not have ordered keys).

Maybe we should individually test that each of these are numbers strictly greater than 0 and less than 1?

paulgb (commented):
This looks good structurally, left a few comments.

src/drone/agent/executor.rs
```diff
@@ -195,6 +232,7 @@ impl Executor {
 
                     if state.running() {
                         self.start_log_loop(&spawn_request.backend_id);
+                        self.start_stats_loop(&spawn_request.backend_id)
```
paulgb, line 235: nit: add semicolon at the end.

This is one of the few places where Rust doesn't care if there's a semicolon, but I prefer to use semicolons in those cases because a reader can look at a line in isolation and see that the return value is discarded.

paulgb (approved):
One nit, otherwise looks good. We will probably want to take a deeper look at the calculation (handling of missing values, but also whether the calculations make sense for us).

PR #99, opened 2022-07-29
add backend to state update messages

rolyatmax:


src/messages/agent.rs
```diff
@@ -229,16 +229,20 @@ pub struct BackendStateMessage {
     /// The new state.
     pub state: BackendState,
 
+    /// The backend id.
+    pub backend: BackendId,
```
rolyatmax, line 233: Is it better to borrow this?

paulgb: No, it's fine like this. It's usually better for structs to own their members instead of holding references to them.

paulgb (approved):

src/messages/agent.rs
rolyatmax: I figured

PR #102, opened 2022-08-09
Ensure that containers are removed

paulgb:
Instead of just stopping containers, remove them. Should also cut down on spurious warnings about containers not existing.


rolyatmax (approved):
LGTM!

PR #103, opened 2022-08-09
ensure that container stats collection is stopped when container stat…

pretentious7:
…s no longer being reported

fixes bug with 'cpu_usage:null' reports in nats for swept backends

### DIFF
--- a/src/drone/agent/executor.rs
+++ b/src/drone/agent/executor.rs
@@ -179,12 +179,18 @@ impl Executor {
 
                     while let Some(v) = stream.next().await {
                         match v {
-                            Ok(v) => {
-                                if let Some(message) = DroneStatsMessage::from_stats_message(&v) {
+                            Ok(v) => match DroneStatsMessage::from_stats_message(&v) {
+                                Some(message) => {
                                     nc.publish(&DroneStatsMessage::subject(&backend_id), &message)
                                         .await?;
                                 }
-                            }
+                                None => {
+                                    let message =
+                                        "failed to get stats (container may have been swept)";
+                                    tracing::warn!(message);
+                                    return Err(anyhow::Error::msg(message));
+                                }
+                            },
                             Err(error) => {
                                 tracing::warn!(?error, "Error encountered sending stats.")
                             }
--- a/src/drone/agent/executor.rs
+++ b/src/drone/agent/executor.rs
@@ -187,7 +187,7 @@ impl Executor {
                                 None => {
                                     let message =
                                         "failed to get stats (container may have been swept)";
-                                    tracing::warn!(message);
+                                    tracing::info!(%backend_id, message);
                                     return Err(anyhow::Error::msg(message));
                                 }
                             },
--- a/src/messages/agent.rs
+++ b/src/messages/agent.rs
@@ -92,12 +92,15 @@ impl DroneStatsMessage {
         //cpu
         let cpu_stats = &stats_message.cpu_stats;
         let precpu_stats = &stats_message.precpu_stats;
-        let cpu_delta = cpu_stats.cpu_usage.total_usage - precpu_stats.cpu_usage.total_usage;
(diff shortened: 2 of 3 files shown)


paulgb:
We can close #101 thanks to this now right?

pretentious7:
ye, closed

tests/src/test-agent.ts
```diff
@@ -187,15 +187,66 @@ test("stats are acquired", async (t) => {
...
+  const statsStatusSubsription =
+    new NatsMessageIterator<unknown>(
+      nats.subscribe(`backend.${backendId}.stats`, { timeout: 1000000 })
+    )
+
+  try {
+    while (true) {
+      //NOTE: this sleep is strongly coupled to 
+      //DEFAULT_DOCKER_THROTTLED_STATS_INTERVAL_SEC in agent/docker.rs
+      await Promise.race([statsStatusSubsription.next(), sleep(11000)])
```
paulgb, line 240: I don't like using exceptions for happy-path control flow.

I think this should work?

```
let success = await Promise.race([statsStatusSubsription.next(), sleep(11000)])
if (!success) {
  break;
}
```

And then remove the outer `try / catch`

paulgb (commented):
Thanks, looks good besides a comment about the tests.

src/drone/agent/executor.rs
```diff
@@ -179,12 +179,18 @@ impl Executor {
...
+                            Ok(v) => match DroneStatsMessage::from_stats_message(&v) {
+                                Some(message) => {
                                     nc.publish(&DroneStatsMessage::subject(&backend_id), &message)
                                         .await?;
                                 }
-                            }
+                                None => {
+                                    let message =
+                                        "failed to get stats (container may have been swept)";
+                                    tracing::warn!(message);
```
rolyatmax, line 190: Might be useful to log the backend id here as well

rolyatmax: Also, maybe we should make this an `info` instead of a `warn`? We expect this to happen once for every backend after they've stopped, right?

src/messages/agent.rs
```diff
@@ -92,12 +92,15 @@ impl DroneStatsMessage {
         //cpu
         let cpu_stats = &stats_message.cpu_stats;
         let precpu_stats = &stats_message.precpu_stats;
-        let cpu_delta = cpu_stats.cpu_usage.total_usage - precpu_stats.cpu_usage.total_usage;
-        let sys_cpu_delta = cpu_stats.system_cpu_usage.unwrap_or_default()
-            - precpu_stats.system_cpu_usage.unwrap_or_default();
+        let cpu_delta =
```
rolyatmax, line 95: From offline convo: add a comment here about what `cpu_usage` means (incrementing clock ticks) and link to the docs from whence came these formulas: https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerStats

pretentious7: done

tests/src/test-agent.ts
```diff
@@ -186,15 +186,68 @@ test("stats are acquired", async (t) => {
...
+  expectResponse(t, nats, "drone.1.spawn", request, true)
+
+  await sleep(100)
+
+  const statsStatusSubsription =
+    new NatsMessageIterator<unknown>(
+      nats.subscribe(`backend.${backendId}.stats`, { timeout: 1000000 })
+    )
+
+  //TODO: figure out a way to write a test such that
```
paulgb, line 235: Does this TODO not describe what's implemented below? We can probably just remove it, or move it into a linear issue if it's something you want to do in the next few cycles.

paulgb (approved):

tests/src/test-agent.ts
pretentious7: yea, what I'd like is a nice test.eventuallyEnds(nats.subscription, TIMEOUT), but I'll stick that into linear backlog.

PR #105, opened 2022-08-12
Abhi/dis 486 allow certs via eab enables other cert

pretentious7:


src/drone/cert/mod.rs
```diff
@@ -36,7 +40,21 @@ pub async fn get_certificate(
...
 
     let mut builder = AccountBuilder::new(dir);
-    builder.contact(vec!["mailto:paul@driftingin.space".to_string()]);
+    builder.contact(vec![format!("mailto:{}", mailto_email)]);
+    if let Some(eab_key_b64) = acme_eab_key {
+        if let Some(kid) = acme_eab_kid {
+            let eab_key = {
+                let value_b64 = eab_key_b64;
+                let value = base64::decode(&value_b64).expect("cannot decode base64 value");
+                /*
```
paulgb, line 49: remove this

src/drone/cli.rs
```diff
@@ -46,6 +46,18 @@ pub struct Opts {
     #[clap(long, action)]
     pub acme_server: Option<String>,
 
+    //email to use as mailto for cert issuance
```
paulgb, line 49: Use a proper docstring so that this appears in the `--help` message of the CLI:

```
/// Email to use as mailto: for ACME certificate issuance.
```

Note the three `///`; without this, `clap` won't pick this up as cli documentation.

src/drone/cli.rs
```diff
@@ -46,6 +46,18 @@ pub struct Opts {
...
+    #[clap(long, action)]
+    pub cert_email: Option<String>,
+
+    /// EAB ID
```
paulgb, line 53: Replace `EAB` with `ACME External Account Binding` for clarity, here and also below.

Might also be good to explain that this is only used for ACME providers other than Let's Encrypt.

tests/src/test-cert-refresh.ts
```diff
@@ -49,3 +49,103 @@ test("Generate certificate", async (t) => {
...
 })
+
+
+test("Generate cert with EAB credentials", async (t) => {
+  const natsPort = await t.context.docker.runNats()
+  const isEab = true
+  const pebble = await t.context.docker.runPebble(isEab)
+
+  await sleep(500)
+  //t.timeout(5000, "Connecting to NATS")
```
paulgb, line 60: remove

tests/src/test-cert-refresh.ts
```diff
@@ -49,3 +49,103 @@ test("Generate certificate", async (t) => {
...
+  await certRefreshPromise
+  t.assert(validateCertificateKeyPair(keyPair))
+})
+
+test.only("incorrect eab credentials cause panic", async (t) => {
+  const natsPort = await t.context.docker.runNats()
+  const isEab = true
+  const pebble = await t.context.docker.runPebble(isEab)
+  await sleep(500)
+  //needed else NatsError
```
paulgb, line 101: ?

tests/src/test-cert-refresh.ts
```diff
@@ -49,3 +49,103 @@ test("Generate certificate", async (t) => {
...
+    t.true(e instanceof Error, "spawner does not error out when acme_kid is invalid")
+  }
+
+  try {
+    const certRefreshPromise = t.context.runner.certRefresh(
+      keyPair,
+      natsPort,
+      pebble,
+      isEab,
+      //note last letter in key is different (hence invalid)
```
paulgb, line 138: nit: instead of a comment explaining a subtle change like this, I'd rather just make the change obvious (while still forcing the same code path, i.e. still valid base64). So something like:

` { kid: 'kid-1', key: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" }`

tests/src/test-cert-refresh.ts
```diff
@@ -49,3 +49,103 @@ test("Generate certificate", async (t) => {
...
+    t.context.tempdir.path("keys/cert.key"),
+    t.context.tempdir.path("keys/cert.pem")
+  )
+  const sub = new NatsMessageIterator<DnsMessage>(
+    nats.subscribe("acme.set_dns_record")
+  )
+
+
+  let error_free_cert_refreshes = 0;
+  try {
```
paulgb, line 117: Instead of `try` / `catch` here (and below), we can use [`throwsAsync`](https://github.com/avajs/ava/blob/main/docs/03-assertions.md#throwsasyncthrower-expectation-message) from ava.

tests/src/util/runner.ts
```diff
@@ -56,11 +56,32 @@ export class DroneRunner implements DropHandler {
   async certRefresh(
     certs: KeyCertPair,
     natsPort: number,
-    pebble: PebbleResult
+    pebble: PebbleResult,
+    eab = false,
```
paulgb, line 60: Instead of passing an `eab` boolean, let's just assume `eab` is used if `eabOptions` is not `undefined`.

tests/src/util/docker.ts
```diff
@@ -56,13 +56,13 @@ export class Docker implements DropHandler {
...
 
-  async runPebble(): Promise<PebbleResult> {
+  async runPebble(eab = false): Promise<PebbleResult> {
     const port = await getPort()
     const tempdir = mkdtempSync(join(tmpdir(), "spawner-pebble-config-"))
     const certs = await generateCertificates()
     const imageName = "docker.io/letsencrypt/pebble:latest"
 
-    const pebbleConfig = {
+    const pebbleConfig = eab ? {
```
paulgb, line 65: rather than do this (which creates the possibility of accidental drift between the duplicated fields, like `httpPort`), I'd rather have one `pebbleConfig` definition that [spread syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) in the eab configuration. For example:

```
const eabConfig = {};
if (eab) {
    eabConfig = { 
        externalAccountBindingRequired: true,
        // etc.
    }
}
const pebbleConfig = {
    pebble: {
        listenAddress: "blah",
        // etc.
        ...eabConfig
    }
}
```

tests/src/util/runner.ts
```diff
@@ -56,11 +56,32 @@ export class DroneRunner implements DropHandler {
...
+      "--acme-eab-key",
+      eabOptions?.key,
+      "--acme-eab-kid",
+      eabOptions?.kid,
+      "cert",
+    ]
     const proc = spawn(
       SPAWNER_PATH,
-      [
+      eab ? eab_cli_options : [
```
paulgb, line 84: This is another place where spread notation is preferable to defining two separate args lists.

paulgb (commented):
Looks good, a few comments.

tests/src/test-cert-refresh.ts
pretentious7: clarified this, just wanted to note that the nats server is necessary because otherwise spawner wouldn't get to the code paths being tested.

src/drone/cli.rs
pretentious7: kinda ironic, considering the IETF came up with it for them :laughing: . Noted.

paulgb (approved):
Thanks for the fixes, LGTM.

PR #109, opened 2022-08-16
Refactor EAB data flow

paulgb:
This is a light refactoring that simplifies data flow of EAB kid/keys. The main thing I am aiming for here are:
- Fuse together EAB key and kid to simplify data flow inside the codebase.
- Handle missing key/kid early so that they are an error at the parse stage.
- Handle base64-decoding early so that it is an error at the parse stage.

I also removed the failure test. In general I like tests of failure cases, but in this case the failure isn't particularly interesting (we're effectively just testing for a non-zero exit code), and the test was causing me other problems with flakiness.

### DIFF
--- a/drone/src/drone/cli.rs
+++ b/drone/src/drone/cli.rs
@@ -87,6 +87,12 @@ pub struct Opts {
     command: Option<Command>,
 }
 
+#[derive(Debug, PartialEq)]
+pub struct EabKeypair {
+    pub eab_kid: String,
+    pub eab_key: Vec<u8>,
+}
+
 #[derive(Subcommand)]
 enum Command {
     /// Migrate the database, and then exit.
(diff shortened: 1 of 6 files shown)


drone/src/drone/cli.rs
```diff
@@ -196,6 +201,19 @@ impl From<Opts> for DronePlan {
 
         let db = opts.db_path.map(DatabaseConnection::new);
 
+        let acme_eab_keypair = match (opts.acme_eab_key, opts.acme_eab_kid) {
```
pretentious7, line 204: Ahhh, I was trying to do something like this. Thanks, this is very helpful.

pretentious7:
what flakiness issues was the failure case test having? For me the value there was mostly being able to see the ServerError's in the logs from pebble, but I couldn't figure out a simple way to parse for those. But then again the cert generation tests cover communication with pebble, so I suppose its not necessary. Was very useful while debugging during dev though.

pretentious7 (approved):

rolyatmax (approved):

paulgb (commented):
bump timeout

PR #110, opened 2022-08-17
First pass of scheduler

paulgb:


paulgb:
This needs more tests, but in the spirit of minimal(ish) PRs, I'd like to push this for now and then work on #132.

controller/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "dis-spawner-controller"
```
rolyatmax, line 2: Just curious: why the name `controller` instead of `scheduler`?

drone/src/drone/agent/mod.rs
```diff
@@ -136,18 +139,11 @@ async fn ready_loop(nc: TypedNats, drone_id: DroneId, cluster: String) -> Result
 pub async fn run_agent(agent_opts: AgentOptions) -> Result<()> {
     let nats = agent_opts.nats.connection().await?;
 
-    // Ensure that status stream exists.
-    nats.add_jetstream_stream(
-        &BackendStateMessage::stream_name(),
-        BackendStateMessage::wildcard_subject(),
-    )
-    .await?;
```
rolyatmax, line 144: I was thinking we would still need this? Isn't there a chance that the jetstream stream won't exist before the drone tries to publish to it?

core/src/messages/agent.rs
```diff
@@ -140,7 +142,7 @@ impl BackendStatsMessage {
 #[derive(Serialize, Deserialize, Debug)]
 pub struct DroneStatusMessage {
     pub drone_id: DroneId,
-    pub cluster: String,
+    pub cluster: ClusterId,
     pub capacity: u32,
```
pretentious7, line 146: I've been thinking about this capacity idea. Do we get anything out of having the drone report what it thinks its capacity is? rather than deriving it scheduler side from the metrics stream? 

I recall that we thought it might help with the case where a lot of backends spawn and the scheduler overcommits them to a drone based on stale metrics, but isn't that also an issue with the capacity approach?

pretentious7 (approved):
looks good! can't wait to write and benchmark scheduler algorithms.

controller/Cargo.toml
paulgb: My thought was that both would live in the same "controller" binary, similar to how the proxy, agent, and cert-refresher live in the "drone" binary. I'm not sure yet whether it should be possible to run both in the same process (which simplifies local test deployment) or not, but even if not, we could build them into the same container and switch between them as a runtime flag.

I'm also tempted to combine the drone and controller into one binary so that we just have one image that does everything, and then we run it with different flags to get the different components.

controller/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+        )
+        .await,
+    );
+
+    loop {
+        select! {
+            status_msg = status_sub.next() => {
+                let status_msg = status_msg.unwrap();
+
+                scheduler.update_status(Utc::now(), &status_msg);
```
rolyatmax, line 31: I wonder if there's a way to make this somewhat resistant to delays in message delivery 🤔  I suppose we could include a timestamp in the `DroneStatusMessage`, but then we'd be relying on the drone's clock and the controller's clock being relatively in sync.

controller/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+            },
+
+            spawn_request = spawn_request_sub.next() => {
+                let spawn_request = spawn_request.unwrap().unwrap();
+
+                let result = match scheduler.schedule() {
+                    Ok(drone_id) => {
+                        match nats.request(&spawn_request.value.schedule(drone_id)).await {
+                            Ok(false) | Err(_) => ScheduleResponse::NoDroneAvailable,
```
rolyatmax, line 40: Do we want to differentiate between NoDroneAvailable and other types of errors?

core/src/messages/scheduler.rs
```diff
@@ -0,0 +1,94 @@
...
+impl Display for ClusterId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        self.hostname.fmt(f)
+    }
+}
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct ScheduleRequest {
+    pub cluster: ClusterId,
```
rolyatmax, line 39: The idea is that there would be a single scheduler across multiple clusters?

core/src/messages/agent.rs
paulgb: We don't really do anything with this capacity value as of now. I imagine it being surpassed by the stats you're reporting.

I'll leave this in until we decide what capacity-based scheduling looks like, but I agree that it's probably not this.

controller/src/lib.rs
paulgb: My thinking is that `NoDroneAvailable` will be a user-exposed error, where everything else is an internal error. The justification is that `NoDroneAvailable` could be the fault of the user who runs the cluster, whereas other errors are our fault.

core/src/messages/scheduler.rs
paulgb: Yes.

PR #111, opened 2022-08-17
Rewrite tests in Rust

paulgb:


dev/src/container.rs
```diff
@@ -0,0 +1,197 @@
...
+            Some(LogsOptions::<&str> {
+                follow: true,
+                stderr: true,
+                stdout: true,
+                timestamps: true,
+                ..LogsOptions::default()
+            }),
+        );
+
+        let mut log_file = File::create(scratch_dir("logs").join(format!("{}.txt", spec.name)))?;
```
pretentious7, line 129: it'd be nice to be able to pass in a filehandle to ContainerResource (use case here is that we could pass in stdout, which would be nice for just seeing the result)

Alternatively, would be nice to have an option to tee stdout?

coverage.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cargo llvm-cov nextest -j 10 --lcov --output-path lcov.info
```
pretentious7, line 3: are we using this file for anything at present?

paulgb: The [Coverage Gutters](https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters) vscode extension will pick it up if it exists.

dev/src/container.rs
pretentious7: my use case here is that I quite often write a test then run `ls * | entr -s :TEST_COMMAND`, which reruns that test whenever something changes, and its nice to be able to see the output in that terminal.

paulgb: I don't think the `ContainerResource` is the right level to do that at, maybe it could be done at the test level. I'm struggling to come up with a use case where it's more useful than just running `tail -f` on the file as the test is running, though.

paulgb: Ah, I didn't see your second comment until after mine, I'll give that some thought.

pretentious7:
This is great :D.

I'd like the option to be able to see the logs in stdout (and being able to see ALL the logs, not just from spawner but from nats and pebble would be very nice). But I can always write a utility that tails all new files in the scratch directory, so its not a blocker.

LGTM!

pretentious7 (approved):

PR #121, opened 2022-08-22
Log IP on proxy requests

paulgb:


drone/Cargo.lock
```diff
@@ -1,2745 +0,0 @@
-# This file is automatically @generated by Cargo.
```
rolyatmax, line 1: Was deleting this file intentional?

dev/tests/proxy.rs
```diff
@@ -110,6 +110,12 @@ impl Proxy {
             .resolve(&hostname, self.bind_address)
             .build()?;
 
+        let path = if let Some(path) = path.strip_prefix("/") {
+            path
+        } else {
+            path
+        };
+
```
pretentious7, line 118: I don't think I understand this, why does it return path in both branches?

pretentious7 (approved):
Lgtm, just a bit confused about that if let

dev/tests/proxy.rs
paulgb: In the first branch, `path` is shadowed by the `if let Some(path)` assignment. So it's the same variable name in both branches, but they refer to a different value. I probably could have used `unwrap_or` here instead, I'll clean that up next time I see it.

PR #123, opened 2022-08-24
Add backend.:id.terminate nats endpoint

pretentious7:
Draft pr to add backend.:id.terminate. 

I have not yet added the typed-nats checks to ensure the terminate command body refers to the backend_id in the address. (but I don't think that should substantially change anything in what's been written thus far,  just return an error either when `sub.next()` is called or when `msg.value` is accessed

### DIFF
--- a/dev/tests/agent.rs
+++ b/dev/tests/agent.rs
@@ -504,6 +504,7 @@ async fn handle_termination_request() -> Result<()> {
     let mut controller_mock = MockController::new(nats.clone()).await?;
 
     let mut request = base_spawn_request();
+    request.max_idle_secs = Duration::from_secs(10000);
 
     controller_mock
         .expect_handshake(request.drone_id, agent.ip)
@@ -520,11 +521,14 @@ async fn handle_termination_request() -> Result<()> {
         .wait_for_state(BackendState::Ready, 60_000)
         .await?;
 
+    /*
     let proxy_route = agent
         .db
         .get_proxy_route(request.backend_id.id())
         .await?
         .expect("Expected proxy route.");
+    */
+    /*
     tokio::spawn(async move {
         loop {
             let result = reqwest::get(format!("http://{}/", proxy_route)).await;
@@ -532,6 +536,7 @@ async fn handle_termination_request() -> Result<()> {
             std::thread::sleep(std::time::Duration::from_secs(1));
         }
     });
+    */
 
     let termination_request = TerminationRequest {
         backend_id: request.backend_id.clone(),
--- a/dev/tests/agent.rs
+++ b/dev/tests/agent.rs
@@ -9,7 +9,7 @@ use dev::{
 use dis_spawner::{
     messages::agent::{
         BackendState, BackendStateMessage, BackendStatsMessage, DroneConnectRequest,
-        DroneConnectResponse, DroneStatusMessage, SpawnRequest,
+        DroneConnectResponse, DroneStatusMessage, SpawnRequest, TerminationRequest,
     },
     nats::{TypedNats, TypedSubscription},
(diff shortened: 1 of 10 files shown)


paulgb (commented):
Thanks for this. Left some comments.

pretentious7:
I've rebased onto main (post the TypedMessage refactor) and addressed the comments.

dev/tests/agent.rs
```diff
@@ -488,3 +495,53 @@ async fn handle_agent_restart() -> Result<()> {
...
+    state_subscription
+        .wait_for_state(BackendState::Ready, 60_000)
+        .await?;
+
+    let proxy_route = agent
+        .db
+        .get_proxy_route(request.backend_id.id())
+        .await?
+        .expect("Expected proxy route.");
+    tokio::spawn(async move {
```
paulgb, line 528: I guess this is just to keep it alive? Instead, let's just use a long grace period.

I've also created an issue (#128) to allow more flexible termination conditions, including no maximum idle time, which we could use here once implemented.

drone/src/drone/agent/executor.rs
```diff
@@ -104,6 +105,16 @@ impl Executor {
         self.run_backend(spawn_request, BackendState::Loading).await
     }
 
+    pub async fn kill_backend(
+        &self,
+        termination_request: &TerminationRequest,
+    ) -> Result<(), anyhow::Error> {
+        tracing::info!("Trying to terminate!");
```
paulgb, line 112: Add the backend ID here, e.g.

```
tracing::info!(backend_id=%termination_request.backend_id, "Trying to terminate backend.");
```

drone/src/drone/agent/mod.rs
```diff
@@ -132,24 +154,19 @@ pub async fn run_agent(agent_opts: AgentOptions) -> Result<()> {
...
 
     match result {
         DroneConnectResponse::Success { drone_id } => {
-            {
-                let nats = nats.clone();
-                let cluster = cluster.clone();
-                tokio::spawn(ready_loop(nats, drone_id, cluster));
-            }
-
+            let executor = Arc::new(Executor::new(docker, db, nats.clone()));
```
paulgb, line 162: Executor is designed to be cloned (the underlying types use `Arc`), so wrapping it in an `Arc` creates another layer of indirection. Better to just create an executor and clone it.

drone/src/drone/agent/mod.rs
```diff
@@ -132,24 +154,19 @@ pub async fn run_agent(agent_opts: AgentOptions) -> Result<()> {
...
+            let result = tokio::try_join!(
+                ready_loop(nats.clone(), drone_id, cluster.clone()),
+                listen_for_spawn_requests(drone_id, executor.clone(), nats.clone()),
+                listen_for_termination_requests(executor.clone(), nats.clone())
+            );
             tracing::info!("Listening for spawn requests.");
```
paulgb, line 168: This won't be run until after the `try_join`, should move it above.

dev/tests/agent.rs
pretentious7: hehe yes.

drone/src/drone/agent/mod.rs
pretentious7: needed to do a small refactor here and there to get his to work, but done!

drone/src/drone/agent/mod.rs
pretentious7: moved it into listen_for_spawn_requests. might be a good idea to put these Listening for {{X}} traces into the nats subscribe call.

paulgb (approved):
LGTM, thanks!

PR #125, opened 2022-08-26
Refactor Typed NATS wrapper

paulgb:
NATS publish permissioning works at the subject glob level: a user can be given rights to publish to subjects that match a specific glob. If the subject refers to a particular resource (e.g., a particular backend or drone), the ID of that resource can be encoded into the subject to limit who has access to that resource.

A listener who is listening on a wildcard needs to ensure that an incoming message refers to the resource in its subject. Otherwise, a writer with permission to publish to any resource could effectively create messages that refer to *any* subject, and have them treated the same by the wildcard subscriber.

One way to do this is for the subscriber to parse the subject, and then check that the value in the message matches the value in the subject. Another way (which avoids parsing) is to standardize the way subjects are generated for each message type. When a wildcard subscriber receives a message, it regenerates the subject and ensures that it matches the subject over which the message was sent. The advantage of this is that it can be done directly in the typed NATS wrapper, so that it is opaque to subscribers.

This takes the latter approach, implemented as follows:
- A `TypedMessage` trait is introduced. Any message that can be sent to a regular subject (as opposed to a reply inbox) gets this trait. It encodes both a reply type, and a method called `subject` that returns the canonical `Subject` for a particular instance of that type.
- `Subject` and related traits are simplified by referring to a single `TypedMessage` instead of a pair of (request, reply) message types. Because the `TypedMessage` already encodes its reply type, no type safety is lost with this change, but code is simplified.

### DIFF
--- a/core/src/logging.rs
+++ b/core/src/logging.rs
@@ -18,9 +18,9 @@ use tracing_subscriber::{layer::SubscriberExt, registry::LookupSpan};
 const TRACE_STACKDRIVER: &str = "TRACE_STACKDRIVER";
 const LOG_DEFAULT: &str = "info,sqlx=warn";
 
-async fn do_logs(nc: &TypedNats, mut recv: tokio::sync::mpsc::Receiver<LogMessage>, subject: &str) {
+async fn do_logs(nc: &TypedNats, mut recv: tokio::sync::mpsc::Receiver<LogMessage>) {
     while let Some(msg) = recv.recv().await {
-        if let Err(err) = nc.publish(&LogMessage::subject(subject), &msg).await {
+        if let Err(err) = nc.publish(&msg).await {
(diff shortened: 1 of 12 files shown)


pretentious7 (approved):
Took me a while to understand how this worked with wildcard_subject, but I think I get it. neat!

rolyatmax (approved):

pretentious7:
Might be nice to have a test which sends a purposely badly created message? for example a spawn_request which has the wrong drone_id in the message or something like that?

paulgb:
> Might be nice to have a test which sends a purposely badly created message? for example a spawn_request which has the wrong drone_id in the message or something like that?

This PR doesn't actually implement the check, it just refactors the code in anticipation of that check. I agree that a unit test like that should be added once we add the check.

PR #127, opened 2022-08-29
Expose some subject names for use outside of Spawner

paulgb:


pretentious7:
so this adds an explicit log_channel instead of needing to know whatever the nats id is to subscribe?

pretentious7 (approved):

paulgb:
> so this adds an explicit log_channel instead of needing to know whatever the nats id is to subscribe?

Yes, the main difference now is just that it flows through the `LogMessage` object, which allows us to maintain the property that subject is deterministically derived from object.

Issue #132, opened 2022-09-01
Refactor NATS wrapper


paulgb:
The NATS wrapper has evolved a bit and could use some conceptual cleanup.

- [x] Get rid of `Subject`. `subject()` should return a `String`.
- [x] Get rid of `StreamName` type and replace it with `HasJetstream`. This trait should include a `stream_config() -> Config` method that returns a JetStream configuration.
- [x] Jetstream methods on a `TypedNats` should automatically create the jetstream if it does not exist the first time it is used. The TypedNats object should include an `Arc<RwLock<HashSet<String>>>` to keep track of whether a particular stream has been created.

PR #135, opened 2022-09-01
Get rid of Subject type (per #132)

paulgb:
This is the first of several refactors under #132 to make the NATS wrapper nicer/safer to work with.

**Remove `Subject`**: `Subject` was only used to describe subjects to publish to, and `SubscribeSubject` was used for subscriptions. It's still useful to have a typed handle to a (possibly wildcard) subject to use for subscriptions, but we recently added a rule that a message should always be sent to a subject deterministically derived from the message itself. That means we no longer have a need for a typed `Subject`; when we want to send a message we just pass the message itself.

**Remove `StreamName`**: `StreamName` was used as a typed handle to the JetStream stream associated with an object. Rather than passing this handle around, we implement a `JetStreamable` trait instead that returns a config. This has the additional advantage of allowing us to add per-stream configuration rules.

### DIFF
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -16,7 +16,8 @@ pub async fn run_scheduler(nats: TypedNats) -> Result<()> {
     let mut spawn_request_sub = nats.subscribe(ScheduleRequest::subscribe_subject()).await?;
 
     let mut status_sub = Box::pin(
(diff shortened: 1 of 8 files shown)


pretentious7 (approved):
LGTM!

PR #136, opened 2022-09-01
Lazily create jetstreams (per #132)

paulgb:


core/src/nats.rs
```diff
@@ -163,6 +165,7 @@ impl<T> NatsResultExt<T> for std::result::Result<T, async_nats::Error> {
 pub struct TypedNats {
     nc: Client,
     jetstream: Context,
+    jetstream_created_streams: Arc<DashSet<String>>,
```
pretentious7, line 168: I might be misunderstanding something here, but why does this DashSet need to be in an Arc? Shouldn't it be cloneable by itself?

paulgb: If we clone a `DashSet`, we're working with a clone of the data, so there's a possibility that we would try to create the same stream N times if we have N clones. When wrapped in `Arc`, every clone shares the same set and will only create the stream once (notwithstanding race conditions, which are nbd since it's an idempotent operation).

You could still create multiple `TypedNats` instances in the same problem and each will have their own set, but we avoid doing that anyway because then each has its own NATS connection.

rolyatmax (approved):
LGTM

PR #142, opened 2022-09-02
Use `publish_jetstream` for backend status messages.

paulgb:


drone/src/drone/agent/executor.rs
```diff
@@ -254,7 +254,7 @@ impl Executor {
                         .await
                         .log_error();
                     self.nc
-                        .publish(&BackendStateMessage::new(
+                        .publish_jetstream(&BackendStateMessage::new(
```
rolyatmax, line 257: Do we need to publish the backend logs to Jetstream as well?

rolyatmax (approved):

drone/src/drone/agent/executor.rs
paulgb: They still get published to jetstream even if they aren't explicitly published with `publish_jetstream`, as long as they are published to a subject associated with the jetstream. The difference is that `publish_jetstream` waits for an ack from the Jetstream server.

paulgb: (This is a bit confusing, but it's consistent with NATS' own semantics)

Issue #133, opened 2022-09-01
Allow configuration by config file


paulgb:
We've outgrown command-line only configuration.

When it comes to configuration files there's a tricky balance; on one hand we don't want to stray too far from expected practices; on the other hand, common practices are fractured and leave a lot to be desired.

Some things to consider:

- It's common to have a fairly static configuration but want to set some values at runtime. How should this be accomplished?
  - [Caddy approach](https://caddyserver.com/docs/caddyfile-tutorial#environment-variables): env var templating in configuration file
  - [Nginx approach](https://hub.docker.com/_/nginx/) doesn't have built-in env var substitution, but the docker image implements it based on [envsubst](https://www.gnu.org/software/gettext/manual/html_node/envsubst-Invocation.html)
  - [docker registry](https://docs.docker.com/registry/configuration/#override-specific-configuration-options) allows overriding individual yaml leaves by turning the full path to that leaf into a variable name. This is good for overriding string values, but is awkward for list/array/set values.
- It's important to be able to automatically generate configuration from popular languages, so it should be a relatively common format.
- It's important to be able to add comments to configuration, so if we were to go with something like JSON, we'd want to use a flavor that supported comments. And trailing commas.

Some options:
- `TOML`: Rust projects tend to favor it.
- `YAML`: Popular with Go/CNCF projects, but [No Body Wants To Write YAML](https://noyaml.com/). Superficially a superset of JSON, making it very easy to generate universally, but [not without footguns](https://john-millikin.com/json-is-not-a-yaml-subset).
- `JSON`: Good for generating but not for human-generated configuration.
- `Dhall`: Has some nice neat ideas but overcomplicated for our use case and still pretty obscure.
- `CUE`: Exposes less complexity than Dhall although it has a similarly thoughtful design. It could be a nice user experience with the right tooling, but I don't think the tooling is mature enough yet. Mostly targets the Go ecosystem currently.
- [`Nickel`](https://nickel-lang.org/): Cool but too programming-languagey

paulgb:
There are Rust libraries that allow us to read from multiple formats and also to allow environment overrides.

https://github.com/SergioBenitez/Figment
https://crates.io/crates/config

A minimal feature set for us probably includes:
- Ability to use toml (for human-created) and json (for machine-created)
- Reasonable errors on schema validation
- Ability to override with env or cli options
- Serde, ideally

paulgb:
[...]

PR #149, opened 2022-09-04
Replace CLI-based configuration with file-based (#133)

paulgb:
This replaces the previous CLI-based configuration with a layered configuration:
- If a positional argument is passed to the drone on startup, configuration is loaded from there.
- If environment variables matching `SPAWNER_*` are present, they override the configuration.

Additionall, a `--dump-config` flag can be used to dump the configuration as a JSON blob instead of running the drone. This is useful for validation and testing of env overrides.

The only required argument is `cluster_domain`, so this works:

    SPAWNER_CLUSTER_DOMAIN=spawner.test cargo run --bin spawner-drone -- --dump-config

However, to do anything useful, at least one of the optional `[acme]`, `[proxy]`, or `[agent]` sections must be provided.

A more complete example is provided in the `sample_config.toml` file in root. It can be validated with this command:

    cargo run --bin spawner-drone -- sample_config.toml --dump-config


sample_config.toml
```diff
@@ -0,0 +1,74 @@
+# This file contains some examples of Drone configuration.
+# You can run the drone with --dump-config to validate a configuration.
```
pretentious7, line 2: this is not implemented yet?

pretentious7 (approved):
lgtm!

ah this is going to make life a lot nicer.

sample_config.toml
paulgb: It's here: https://github.com/drifting-in-space/spawner/blob/a3dc709adb38451dfb73fb9a8c419aee868fefea/drone/src/drone/mod.rs#L50

pretentious7: ah my bad i keep forgetting the - to _ thing while grepping

Issue #141, opened 2022-09-02
Never panic in `subscribe_jetstream`


paulgb:
We currently have some remaining `unwrap()`s in `subscribe_jetstream` because it's messy to propagate them another way in a stream. We should fix that.

https://github.com/drifting-in-space/spawner/blob/main/core/src/nats.rs#L319

paulgb:
Fixed by #151

PR #151, opened 2022-09-06
Update deps + misc fixes

paulgb:
- Update dependencies, including to `async-nats` 0.19.0.
- Remove panics from `subscribe_jetstream` (#141) 
- Rewrote `LivenessGuard` in the test framework to panic immediately if the future it's guarding terminates.
- Remove a few dependencies made unnecessary by the rewrite of `JetstreamSubscription`.

This is ready for review, but will have some conflicts with #149. Given that this should be an easier review, I'd like to get it in first and rebase 149 over it.

### DIFF
--- a/dev/src/timeout.rs
+++ b/dev/src/timeout.rs
@@ -1,7 +1,10 @@
 use anyhow::Result;
+use futures::FutureExt;
 use std::{
-    fmt::Display,
+    fmt::{Debug, Display},
     future::Future,
+    pin::Pin,
+    task::Poll,
     time::{Duration, SystemTime},
 };
 use tokio::task::JoinHandle;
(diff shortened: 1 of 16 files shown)


dev/src/timeout.rs
```diff
@@ -95,31 +98,68 @@ where
     })
 }
 
-pub struct LivenessGuard {
-    #[allow(unused)]
-    handle: JoinHandle<Result<()>>,
+struct LivenessGuardFuture<D: Debug> {
+    inner: Pin<Box<dyn Future<Output = D> + Send>>,
```
pretentious7, line 102: Just to make sure I understand, this says that the struct LivenessGuardFuture has an inner field on the heap which can't be moved and implements a Future trait that's thread safe?

pretentious7 (approved):

dev/src/timeout.rs
paulgb: Yep, you've got it.

Technically, it's a bit more specific: `Send` means that ownership can move between threads, but not a borrow to it. To share a borrow between threads, it would need `Sync`, too.

PR #152, opened 2022-09-07
Abhi/fix ulimits

pretentious7:
This PR brings back the cpu ulimit, cpu_period, and cpu_quota.
To do so, I've added an optional ResourceLimits field to SpawnRequest of the form:
```rust
ResourceLimits {
    cpu_period: Duration,
    cpu_period_percent: u8,
    cpu_time_limit: Duration
}
```
(The defaults for this field are a 30 min ulimit and a 100 ms cpu_period with 100% cpu_period_percent)

This PR is still in draft, I'll add tests tomorrow, and run manually on staging to ensure no regressions re: jupyterlab etc. just wanted thoughts on the message schema.

### DIFF
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -6,7 +6,7 @@ use bollard::{auth::DockerCredentials, container::LogOutput, container::Stats};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-use serde_with::{DurationMicroSeconds, DurationSeconds};
+use serde_with::DurationSeconds;
 use std::{collections::HashMap, net::IpAddr, str::FromStr, time::Duration};
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -243,7 +243,8 @@ pub struct SpawnRequest {
     pub credentials: Option<DockerCredentials>,
 
     /// Resource limits
-    pub resource_limits: Option<ResourceLimits>,
+    #[serde(default = "ResourceLimits::default")]
+    pub resource_limits: ResourceLimits,
 }
 
 // eventually, this will be generic over executors
@@ -252,24 +253,23 @@ pub struct SpawnRequest {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct ResourceLimits {
     /// period of cpu time, serializes as microseconds
-    #[serde_as(as = "DurationMicroSeconds")]
-    pub cpu_period: Duration,
+    #[serde_as(as = "Option<DurationSeconds>")]
+    pub cpu_period: Option<Duration>,
 
     /// proportion of period used by container
-    pub cpu_period_percent: u8,
+    pub cpu_period_percent: Option<u8>,
 
     /// total cpu time allocated to container    
-
-    #[serde_as(as = "DurationSeconds")]
-    pub cpu_time_limit: Duration,
+    #[serde_as(as = "Option<DurationSeconds>")]
+    pub cpu_time_limit: Option<Duration>,
 }
 
 impl Default for ResourceLimits {
     fn default() -> ResourceLimits {
         ResourceLimits {
-            cpu_period: Duration::from_millis(100),
-            cpu_period_percent: 100,
-            cpu_time_limit: Duration::from_secs(30 * 60),
+            cpu_period: None,
+            cpu_period_percent: None,
+            cpu_time_limit: None,
         }
     }
 }
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -264,6 +264,16 @@ pub struct ResourceLimits {
     pub cpu_time_limit: Duration,
 }
 
+impl Default for ResourceLimits {
+    fn default() -> ResourceLimits {
+        ResourceLimits {
+            cpu_period: Duration::from_millis(100),
+            cpu_period_percent: 100,
+            cpu_time_limit: Duration::from_secs(30),
+        }
+    }
+}
+
 impl TypedMessage for SpawnRequest {
     type Response = bool;
 
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -6,7 +6,7 @@ use bollard::{auth::DockerCredentials, container::LogOutput, container::Stats};
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-use serde_with::DurationSeconds;
+use serde_with::{DurationMicroSeconds, DurationSeconds};
 use std::{collections::HashMap, net::IpAddr, str::FromStr, time::Duration};
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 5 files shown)


core/src/messages/agent.rs
```diff
@@ -241,6 +241,37 @@ pub struct SpawnRequest {
...
+    #[serde_as(as = "DurationSeconds")]
+    pub cpu_time_limit: Duration,
+}
+
+impl Default for ResourceLimits {
+    fn default() -> ResourceLimits {
+        ResourceLimits {
+            cpu_period: Duration::from_millis(100),
+            cpu_period_percent: 100,
+            cpu_time_limit: Duration::from_secs(30 * 60),
```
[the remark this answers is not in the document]
paulgb: I agree with this. Spawner shouldn't be opinionated about its workloads; if these limits make sense in the context of how we employ Spawner we should set them in p2.

core/src/messages/agent.rs
```diff
@@ -241,6 +241,37 @@ pub struct SpawnRequest {
 
     /// Credentials used to fetch the image.
     pub credentials: Option<DockerCredentials>,
+
+    /// Resource limits
+    pub resource_limits: Option<ResourceLimits>,
```
paulgb, line 246: Rather than make this optional, is there a disadvantage to using

```
#[serde(default = "ResourceLimits::default")]
pub resource_limits: ResourceLimits,
```

This only makes sense if the defaults do not apply limits, which I think is the desired behavior.

drone/src/drone/agent/docker.rs
```diff
@@ -309,13 +316,22 @@ impl DockerInterface {
...
                     ),
                     runtime: self.runtime.clone(),
-                    // cpu_period: Some(DEFAULT_CPU_PERIOD),
-                    // cpu_quota: Some(DEFAULT_CPU_QUOTA),
-                    // ulimits: Some(vec![ResourcesUlimits {
-                    //     name: Some("cpu".to_string()),
-                    //     soft: Some(DEFAULT_CPU_TIME_ULIMIT),
-                    //     hard: Some(DEFAULT_CPU_TIME_ULIMIT),
-                    // }]),
+                    //cpu_period: Some(DEFAULT_CPU_PERIOD),
```
paulgb, line 319: get rid of commented code

drone/src/drone/agent/docker.rs
```diff
@@ -309,13 +316,22 @@ impl DockerInterface {
...
+                    cpu_period: Some(resource_limits.cpu_period.as_micros() as i64),
+                    //cpu_quota: Some(DEFAULT_CPU_QUOTA),
```
paulgb, line 321: get rid of commented code

core/src/messages/agent.rs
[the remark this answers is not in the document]
paulgb: I guess this is a general instance of “move optionals to the leaves of the data structure” which I think is generally a good idea. The exception is where the sub-data-structure being missing represents different semantics to when the leaves are missing.

E.g. if we have a data structure

```
struct Foobar {
    foo: Option<String>,
    bar: Option<String>,
}
```

We should only use an `Option<Foobar>` if a value of `None` conveys different semantics to `Foobar {foo: None, bar: None}`. Otherwise, we have multiple ways of representing the same thing, which is not ideal.

core/src/messages/agent.rs
pretentious7: I didn't know I could do that :D

drone/src/drone/agent/docker.rs
```diff
@@ -309,13 +316,29 @@ impl DockerInterface {
...
-                    // cpu_period: Some(DEFAULT_CPU_PERIOD),
-                    // cpu_quota: Some(DEFAULT_CPU_QUOTA),
-                    // ulimits: Some(vec![ResourcesUlimits {
-                    //     name: Some("cpu".to_string()),
-                    //     soft: Some(DEFAULT_CPU_TIME_ULIMIT),
-                    //     hard: Some(DEFAULT_CPU_TIME_ULIMIT),
-                    // }]),
+                    cpu_period: resource_limits
+                        .cpu_period
+                        .and_then(|cpu_period| Some(cpu_period.as_micros() as i64)),
```
paulgb, line 321: When the closure always returns a `Some`, `map()` is idiomatic because you don't have to `Some`-wrap the result, e.g.

```
cpu_period: resource_limits
                        .cpu_period
                        .map(|cpu_period| cpu_period.as_micros() as i64),
```

drone/src/drone/agent/docker.rs
```diff
@@ -309,13 +316,29 @@ impl DockerInterface {
...
+                    cpu_quota: resource_limits
+                        .cpu_period_percent
+                        .and_then(|cpu_period_percent| {
+                            let cpu_period = resource_limits
+                                .cpu_period
+                                .unwrap_or(std::time::Duration::from_millis(100));
+                            cpu_period
+                                .saturating_mul(cpu_period_percent as u32)
+                                .checked_div(100)
+                                .and_then(
```
paulgb, line 331: `map`

paulgb (approved):
This looks great. I made a nit-picky comment but otherwise this is good to go.

dev/tests/agent.rs
```diff
@@ -271,6 +271,82 @@ async fn spawn_with_agent() -> Result<()> {
...
+    }
+
+    async fn connection(self) -> TypedNats {
+        self.nats
+            .connection()
+            .await
+            .expect("expected nats connection")
+    }
+
+    async fn state_subscription(Arc(self)) -> BackendStateSubscription {
```
paulgb, line 296: `self` is syntax sugar for `self: Self`, but I don't think it works through an `Arc`; you need to de-sugar it to `self: Arc<Self>`

dev/tests/agent.rs
```diff
@@ -271,6 +271,82 @@ async fn spawn_with_agent() -> Result<()> {
...
 }
 
+struct SpawnTestContext {
+    nats: Nats,
+    agent: Agent,
+    req: SpawnRequest,
+}
+
+impl SpawnTestContext {
+    async fn proxy_route(self) -> Option<String> {
```
paulgb, line 281: This can be `&self`

dev/tests/agent.rs
```diff
@@ -271,6 +271,82 @@ async fn spawn_with_agent() -> Result<()> {
...
+        self.agent
+            .db
+            .get_proxy_route(self.req.backend_id.id())
+            .await
+            .expect("Expected proxy route.")
+    }
+
+    async fn connection(self) -> TypedNats {
```
paulgb, line 289: This can be `&self`

paulgb (commented):
Although I usually like to pair implementation and tests into one code merge, I'd be comfortable here merging the [98688cc](https://github.com/drifting-in-space/spawner/pull/152/commits/98688cc4d36f92b852f8f5da18c0fe273dcafb3f) commit and then having you pick up tests as a separate PR. The tests that we currently have are sufficient to ensure that at least the case where default values are passed works, and breaking it into two PRs will help with the discussion around test refactoring.

PR #158, opened 2022-09-18
Remove drone handshake and make drone ID a string

paulgb:


core/src/messages/scheduler.rs
```diff
@@ -36,9 +36,9 @@ pub struct ScheduleRequest {
 }
 
 impl ScheduleRequest {
-    pub fn schedule(&self, drone_id: DroneId) -> SpawnRequest {
+    pub fn schedule(&self, drone_id: &DroneId) -> SpawnRequest {
```
pretentious7, line 39: So, this is a rust idiom right? to generally borrow in function params as much as possible? why?

dev/tests/cert.rs
```diff
@@ -5,7 +5,9 @@ use dev::{
     timeout::{spawn_timeout, timeout},
 };
 use dis_spawner::{messages::cert::SetAcmeDnsRecord, nats::TypedNats, types::ClusterName};
-use dis_spawner_drone::{cert::acme::AcmeEabConfiguration, cert::CertOptions, keys::KeyCertPathPair};
+use dis_spawner_drone::{
+    cert::acme::AcmeEabConfiguration, cert::CertOptions, keys::KeyCertPathPair,
+};
```
pretentious7, line 10: I think I'm the one with a misconfigured rustfmt probably, but might be a good idea to put a rustfmt toml in the root dir.

core/src/messages/scheduler.rs
paulgb: Yes. Before, `DroneId` wrapped a `u32`, which is cheap to copy, so `DroneId` implemented `Copy`, and the compiler would make copies of it freely. That way, you can pass it to functions like this without the caller losing ownership of the value.

With this update, `DroneId` wraps a `String`, so it does not implement `Copy`. Our choice here would either be to keep this function signature as-is and call `.clone()` on every `DroneId` we pass in (except where we really do want to transfer ownership), or do what I do here and take a reference.

Generally, my decision between the two is:
- If an argument implements `Copy`, the signature should receive it by value rather than as a reference.
- If an argument does not implement `Copy` and the function does not make a copy of it, take the value by reference.
- If an argument does not implement `Copy` and the function needs to make a copy of it (e.g. to put it into a data structure and return it), it's a judgement call; usually I receive it by value and allow the caller to `clone()` if needed, but if I know that the caller will always need to `clone()`, I might take a reference and bake the cloning behavior into the function.

pretentious7 (approved):

dev/tests/cert.rs
paulgb: What config are you using? I am just using defaults but if you are opinionated about the style I'm open to different settings.

core/src/messages/scheduler.rs
pretentious7: ahhh that makes a lot of sense :+1:

Issue #137, opened 2022-09-01
Generalize job spec


paulgb:
Currently, the spawn message bakes in assumptions related to Docker. We would like to support other execution backends down the road, so we should:

- [x] Move executor-related options into a subfield of `SpawnRequest`/`ScheduleRequest`.
- [x] Make the options into an enum, initially with `DockerContainer` as its only option.
- [x] Allow configuring ulimits etc. through Docker config.

Issue #119, opened 2022-08-20
Implement DNS


paulgb:
The server itself is done, it just needs to be moved into the repo, tests added, and code reviewed.

paulgb:
The DNS server listens for route messages that map backend ids to IPs.

Design decision: who is responsible for emitting these messages, the agent or the proxy?

paulgb:
This will be implemented as part of #186

PR #162, opened 2022-09-22
add SetDnsRecord message

pretentious7:
This PR adds a SetDnsRecord message type, conforming to the address scheme proposed here

I'm not happy with the LocalDnsRecord dance, but seems to be needed to deal with the fact that source can't be serialized (even though it has a serde_cfg serialize something but idk)

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -20,5 +20,6 @@ tokio-stream = "0.1.9"
 tracing = "0.1.36"
 tracing-stackdriver = "0.5.0"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
+trust-dns-server = "0.22.0"
 url = "2.2.2"
 uuid = { version = "1.1.2", features = ["v4"] }
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -20,6 +20,5 @@ tokio-stream = "0.1.9"
 tracing = "0.1.36"
 tracing-stackdriver = "0.5.0"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
-trust-dns-server = "0.22.0"
 url = "2.2.2"
 uuid = { version = "1.1.2", features = ["v4"] }
--- a/core/src/messages/dns.rs
+++ b/core/src/messages/dns.rs
@@ -5,7 +5,7 @@ use crate::{
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Debug)]
-pub enum LocalDnsRecordType {
+pub enum DnsRecordType {
     A,
     TXT,
     AAAA,
@@ -14,7 +14,7 @@ pub enum LocalDnsRecordType {
 #[derive(Serialize, Deserialize, Debug)]
 pub struct SetDnsRecord {
     pub cluster: ClusterName,
-    pub kind: LocalDnsRecordType,
+    pub kind: DnsRecordType,
     pub name: String,
     pub value: String,
 }
--- a/core/src/messages/dns.rs
+++ b/core/src/messages/dns.rs
@@ -3,31 +3,19 @@ use crate::{
     types::ClusterName,
 };
 use serde::{Deserialize, Serialize};
-use trust_dns_server::client::rr::RecordType;
 
 #[derive(Serialize, Deserialize, Debug)]
 pub enum LocalDnsRecordType {
     A,
     TXT,
     AAAA,
-    Unknown(u16),
-}
-
-impl LocalDnsRecordType {
-    fn as_trust_record_type(&self) -> RecordType {
-        match self {
-            LocalDnsRecordType::A => RecordType::A,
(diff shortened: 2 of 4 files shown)


core/Cargo.toml
```diff
@@ -20,5 +20,6 @@ tokio-stream = "0.1.9"
 tracing = "0.1.36"
 tracing-stackdriver = "0.5.0"
 tracing-subscriber = { version = "0.3.15", features = ["env-filter"] }
+trust-dns-server = "0.22.0"
```
paulgb, line 23: Rather than add this dependency here, let's just do anything that uses trust dns types in the DNS server itself. The advantages are:
- It reduces the number of dependencies that need to be fetched and built by consumers of this library who just need the message schemas.
- It avoids tying us to a particular library at the Rust interface level, making the choice of a particular library an implementation decision.

It was a mistake on my part to make this crate depend on bollard, and I'd like to refactor that dependency out as part of #137.

core/src/messages/dns.rs
```diff
@@ -0,0 +1,50 @@
...
+};
+use serde::{Deserialize, Serialize};
+use trust_dns_server::client::rr::RecordType;
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum LocalDnsRecordType {
+    A,
+    TXT,
+    AAAA,
+    Unknown(u16),
```
paulgb, line 13: For other record types, I think we can just make them unrepresentable. We don't need this to be a full DNS server, just enough to route traffic and get certs.

paulgb (commented):
Looks good, left a few comments.

core/Cargo.toml
pretentious7: I was thinking this, but didn't feel like implementing display (for format) and figured it'd be fine since bollard was in there too. yea makes sense to me I'll drop it.

paulgb: Yeah, I appreciate that you tried to do it in a way that was consistent with how we do it for bollard. It was the right design decision given the context you had; my bad for not giving you the context that I want to remove bollard.

core/src/messages/dns.rs
```diff
@@ -0,0 +1,50 @@
...
+            LocalDnsRecordType::TXT => RecordType::TXT,
+            LocalDnsRecordType::AAAA => RecordType::AAAA,
+            LocalDnsRecordType::Unknown(id) => RecordType::Unknown(*id),
+        }
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub struct SetDnsRecord {
+    pub cluster: ClusterName,
```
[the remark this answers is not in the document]
pretentious7: I like that!

core/src/messages/dns.rs
pretentious7: So just have those three in the enum? Sounds good.

paulgb: Yep exactly.

core/Cargo.toml
paulgb: rm this now that it's not used

core/src/messages/dns.rs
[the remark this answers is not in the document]
paulgb: I think this should be a `String`, because in the case of a `TXT` record its value is `_acme-challenge` which is not a `ClusterName`. We could create a "dummy" `ClusterName` for this but that would violate the newtype pattern.

core/src/messages/dns.rs
```diff
@@ -0,0 +1,38 @@
+use crate::{
+    nats::{SubscribeSubject, TypedMessage},
+    types::ClusterName,
+};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum LocalDnsRecordType {
```
paulgb, line 8: What does `Local` refer to in this context? Can we just drop it and call this `DnsRecordType`?

paulgb (commented):
Looking better

core/src/messages/dns.rs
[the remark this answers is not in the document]
pretentious7: ah I completely misunderstood something. does the DNS server itself know what domain its running on?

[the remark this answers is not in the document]
pretentious7: like the record request for acme is _acme_challenge.<cluster> right? where cluster is the domain associated with a cluster? hmmm I think im misunderstanding something very simple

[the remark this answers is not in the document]
pretentious7: and it'll need a ClusterName for the nats subject (for TXT records)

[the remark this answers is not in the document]
paulgb: Ahh sorry ignore me ignore me

I was confusing name and cluster_name. We still need this, and you added name as a string, which is right.

paulgb (approved):
LGTM, thanks!

paulgb:
Related to #119

PR #163, opened 2022-09-23
Fix spawn request logging

paulgb:
This results in a nicer spawn request log message.

### DIFF
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -37,9 +37,9 @@ pub async fn run_scheduler(nats: TypedNats) -> NeverResult {
             },
 
             spawn_request = spawn_request_sub.next() => {
-                tracing::info!(?spawn_request, "Got spawn request");
                 match spawn_request {
                     Ok(Some(spawn_request)) => {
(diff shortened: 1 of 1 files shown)


pretentious7:
wait, why is the message any different?

paulgb:
Oops, all I did was unwrap the `Ok(Some(..))`, I meant to unwrap the actual `.value` from it as well. Just pushed a change that does that.

controller/src/lib.rs
```diff
@@ -31,15 +31,14 @@ pub async fn run_scheduler(nats: TypedNats) -> NeverResult {
...
-                    tracing::info!("here22");
                     return Err(anyhow!("status_sub.next() returned None."));
                 }
             },
 
             spawn_request = spawn_request_sub.next() => {
-                tracing::info!(?spawn_request, "Got spawn request");
                 match spawn_request {
                     Ok(Some(spawn_request)) => {
+                        tracing::info!(spawn_request=?spawn_request.value, "Got spawn request");
```
pretentious7, line 41: hows this work? what's the =? doing?

PR #169, opened 2022-09-27
do not multiply cpu_percent by num_cpus

pretentious7:


pretentious7:
ref https://github.com/drifting-in-space/spawner/issues/164

rolyatmax (approved):
This looks good to me. I say we let @paulgb take a look at that discussion again before we land it.

paulgb (approved):
LGTM.

PR #171, opened 2022-09-27
Minimal functional docker-compose config

### LINKED ISSUE #170: Create docker-compose file
This is a necessary prereq to go live with, and would also help with debugging things.

The compose file should expose NATS on a host port, which the CLI could use by default.

paulgb:
Closes #170

### DIFF
--- a/sample-config/controller.toml
+++ b/sample-config/controller.toml
@@ -1,0 +1,7 @@
+[nats]
+hosts = ["nats"]
+
+# An optional token is allowed for token authentication.
+auth = { token = "my-secret-token" }
+
+[scheduler]
--- a/sample-config/controller.toml
+++ b/sample-config/controller.toml
@@ -2,6 +2,6 @@
 hosts = ["nats"]
 
 # An optional token is allowed for token authentication.
-auth = { token = "my-secret-token" }
+# auth = { token = "my-secret-token" }
 
 [scheduler]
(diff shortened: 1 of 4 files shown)


sample-config/drone.toml
```diff
@@ -58,7 +58,7 @@ ip = "123.123.123.123"
 # Optional Docker settings for the agent.
 [agent.docker]
 # The runtime to use (defaults to "runc")
-runtime = "runsc"
+runtime = "runc"
```
pretentious7, line 61: does this need to be in the sample config?

sample-config/controller.toml
```diff
@@ -0,0 +1,7 @@
+[nats]
+hosts = ["nats"]
+
+# An optional token is allowed for token authentication.
+auth = { token = "my-secret-token" }
```
pretentious7, line 5: rename to "nats-secret-token" here and in docker-compose? to clarify that they have to be the same?

pretentious7: or even a comment line saying "note this must be set on server", (its obvious I know, but I speak from experience setting up synapse and stuff while knowing nothing, we all start somewhere!)

pretentious7 (approved):
couple of small comments to improve novice friendliness, looks good otherwise.

sample-config/drone.toml
paulgb: Functionally, it doesn't need to be because it's the default, but I'd like for this to double as documentation of the toml format until we have better docs.

sample-config/controller.toml
paulgb: hmm fair, I might just comment this out and not set a token on the demo server so that when we have a CLI it can default to not using token auth

PR #174, opened 2022-09-28
Initialize tracing before planning

paulgb:
Currently, if a drone ID is not passed in via configuration, one is generated during the planning stage. Enabling tracing requires that the drone ID be known, so we currently defer enabling tracing until after planning. This means that log messages during planning do not appear.

This PR instead ensures that a drone ID exists in the config that gets passed into planning, generating one if necessary. This allows us to enable tracing before planning begins.

### DIFF
--- a/drone/src/run.rs
+++ b/drone/src/run.rs
@@ -10,6 +10,7 @@ use dis_spawner::cli::init_cli;
 use dis_spawner::logging::TracingHandle;
 use dis_spawner::messages::logging::Component;
 use dis_spawner::retry::do_with_retry;
+use dis_spawner::types::DroneId;
 use dis_spawner::NeverResult;
 use futures::future::try_join_all;
 use futures::Future;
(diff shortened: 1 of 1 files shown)


drone/src/run.rs
```diff
@@ -10,26 +10,38 @@ use dis_spawner::cli::init_cli;
...
 
 async fn drone_main() -> NeverResult {
-    let config: DroneConfig = init_cli()?;
-    let plan = DronePlan::from_drone_config(config).await?;
+    let mut config: DroneConfig = init_cli()?;
+
+    // Extract drone ID, or generate one if necessary.
+    // DronePlan::from_drone_config will do this if we don't do it here,
+    // but we want to initialize tracing now, so we ensure the value
+    // exists before calling from_drone_config.
```
rolyatmax, line 26: Should we remove that logic from `from_drone_config` so that it's clear that `drone_id` is always determined here?

rolyatmax (approved):
Left a question there. Don't want to block anything though, so ✅

drone/src/run.rs
paulgb: I think not, because then it becomes part of the unwritten contract we have with `from_drone_config` if we (or someone else) ever calls it from elsewhere.

PR #177, opened 2022-09-28
move from streaming stats api to polling

pretentious7:
this PR moves from the streaming docker_stats api to a polling API, this solves the currently buggy backend_stats messages (throttling doesn't drop messages) and also cleans up the code a bunch since get_stats no longer need return an Option.

### DIFF
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -137,8 +137,9 @@ impl BackendStatsMessage {
         //      is what proportion of total cpu resource is consumed, and not knowing
         //      the top bound makes that impossible
         let cpu_use_percent = (cpu_delta as f64 / sys_cpu_delta) * 100.0;
-        //disk
-        //TODO: stream https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerInspect
+
+        //TODO: implement disk stats from
+        //      stream at https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerInspect
 
         BackendStatsMessage {
             backend_id: backend_id.clone(),
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -104,7 +104,7 @@ impl BackendStatsMessage {
 }
 
 impl BackendStatsMessage {
-    pub fn from_stats_message(
+    pub fn from_stats_messages(
         backend_id: &BackendId,
         prev_stats_message: &Stats,
         cur_stats_message: &Stats,
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -107,10 +107,12 @@ impl BackendStatsMessage {
         backend_id: &BackendId,
         prev_stats_message: &Stats,
         cur_stats_message: &Stats,
-    ) -> Option<BackendStatsMessage> {
+    ) -> BackendStatsMessage {
         // based on docs here: https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerStats
 
         //memory
+        tracing::warn!(?prev_stats_message, "yeet");
+        tracing::warn!(?cur_stats_message, "deet");
         let mem_naive_usage = cur_stats_message.memory_stats.usage.unwrap_or_default();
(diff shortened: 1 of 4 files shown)


core/src/messages/agent.rs
```diff
@@ -105,14 +105,15 @@ impl BackendStatsMessage {
...
+        cur_stats_message: &Stats,
+    ) -> BackendStatsMessage {
         // based on docs here: https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerStats
 
         //memory
-        let mem_naive_usage = stats_message.memory_stats.usage.unwrap_or_default();
-        let mem_available = stats_message.memory_stats.limit.unwrap_or(u64::MAX);
-        let mem_stats = stats_message.memory_stats.stats;
+        let mem_naive_usage = cur_stats_message.memory_stats.usage.unwrap_or_default();
+        let mem_available = cur_stats_message.memory_stats.limit.unwrap_or(u64::MAX);
```
paulgb, line 115: Rather than using default values (and returning an incorrect result), if these are `None`, we should return an error.

Usually I'd say let's make that a follow-up PR since the problem pre-existed this PR, but since this PR changes the return type of `from_stats_message`, I suggest the following as part of this PR:

- rename `from_stats_message` to `from_stats_messages`
- change the return type to `Result<BackendStatsMessage>`
- if any field is missing, return an anyhow error stating which field is missing

drone/src/agent/docker.rs
```diff
@@ -12,13 +12,13 @@ use bollard::{
...
-use std::{collections::HashMap, net::IpAddr};
-use tokio_stream::{Stream, StreamExt};
+use std::{collections::HashMap, net::IpAddr, time::Duration};
+use tokio_stream::{wrappers::IntervalStream, Stream, StreamExt};
 
 /// The port in the container which is exposed.
 const CONTAINER_PORT: u16 = 8080;
 const DEFAULT_DOCKER_TIMEOUT_SECONDS: u64 = 30;
-const DEFAULT_DOCKER_THROTTLED_STATS_INTERVAL_SECS: u64 = 10;
+const DEFAULT_DOCKER_STATS_INTERVAL_SECONDS: u64 = 10;
```
paulgb, line 21: Add a docstring. If I understand, something like:

```
/// Interval between reporting stats of a running backend.
```

core/src/messages/agent.rs
```diff
@@ -125,13 +126,10 @@ impl BackendStatsMessage {
 
         //REF: https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerStats
         //cpu
-        let cpu_stats = &stats_message.cpu_stats;
-        let precpu_stats = &stats_message.precpu_stats;
+        let cpu_stats = &cur_stats_message.cpu_stats;
+        let precpu_stats = &prev_stats_message.cpu_stats;
```
paulgb, line 130: Let's rename this to avoid confusion with the `precpu` fields that Docker defines.

core/src/messages/agent.rs
```diff
@@ -142,11 +140,11 @@ impl BackendStatsMessage {
         //disk
```
paulgb, line 140: Is this meant to be a todo? If so, turn into a proper todo comment, if not, rm it

drone/src/agent/docker.rs
```diff
@@ -193,19 +193,31 @@ impl DockerInterface {
...
+            stream: false,
+            one_shot: true,
         };
-        self.docker
-            .stats(container_name, Some(options))
-            .throttle(std::time::Duration::from_secs(
-                DEFAULT_DOCKER_THROTTLED_STATS_INTERVAL_SECS,
-            ))
+
+        IntervalStream::new({
```
pretentious7, line 207: also, I was wondering, how does rust know to drop this stream? does it do it when  self gets dropped?

paulgb (commented):
A few comments, otherwise looks good

drone/src/agent/docker.rs
pretentious7: Interesting, is there any way to define a 'container lifetime that'll warn about stuff like this?

paulgb (approved):
LGTM, thanks for the fixes

PR #181, opened 2022-09-29
Initial commit of CLI (lists drones, spawn backends)

paulgb:


core/src/nats.rs
```diff
@@ -300,24 +296,46 @@ impl TypedNats {
...
+            let mut messages = consumer.fetch().messages().await.to_anyhow()?;
+            let mut done = true;
+
+            while let Some(v) = messages.next().await {
+                let v = v.to_anyhow()?;
+                done = false;
+
+                result.push(serde_json::from_slice(&v.payload)?);
+
+                if v.info().to_anyhow()?.pending == 0 {
```
pretentious7, line 318: so pending is the number of messages the server thinks haven't been .next()ed yet right? or is it messages that the server has but the client hasn't got in its queue yet?

guess what I'm asking is, is it possible for pending to be 0 while messages.next() != None

core/src/nats.rs
```diff
@@ -300,24 +296,46 @@ impl TypedNats {
...
+                    done = true;
+                    break
+                }                
+            }
+
+            if done {
+                break;
```
pretentious7, line 325: If a subject is having messages continually delivered to it, and I call get_all, will it ever terminate?

core/src/nats.rs
paulgb: Pending is the number remaining not counting the current message. Unfortunately, when no messages remain, `messages.next().await` returns a `Some(Err(_))` rather than `None`. Because `async_nats` uses boxed errors, and the error itself only indicates the type of error (no messages on server) through a string, checking the error message directly would be pretty ugly. The approach here is to try to avoid encountering the error in the first place, by only calling `.next()` a second time if we know there are messages.

That said, looking at this again, I think it will cause an undesired errors if there are no messages at all on the server, since it will never have a pending value to read. Ugh. I'll revisit this, good catch.

core/src/nats.rs
```diff
@@ -300,24 +296,46 @@ impl TypedNats {
...
+        }
+
+        Ok(result)
+    }
+
+    pub async fn get_latest<T>(&self, subject: &SubscribeSubject<T>) -> Result<Option<T>>
+    where
+        T: TypedMessage<Response = NoReply> + JetStreamable,
+    {
+        let mut result = self.get_all(subject, DeliverPolicy::Last).await?;
```
pretentious7, line 336: with deliverpolicy::last can we just call consumer.messages().next()?

core/src/nats.rs
pretentious7: :/, thats ew, could you take_while() pending != 0, and do the map_err thing to stop reading if there's any errors?

pretentious7: like `messages.take_while(async |res| res.ok_or(false).map(|msg| msg.info().pending > 0)` with whatever prayers it takes for the compiler to accept it.

paulgb: The problem is I don't want to treat any old error the same as the error that happens when there are not messages in the stream. My hack for now is to string match to see if the error is the right one (since async_nats doesn't yet have concrete errors, this is the only way.)

core/src/nats.rs
paulgb: Hmm good call, I guess we could.

core/src/nats.rs
paulgb: Good point, I think it would not terminate.

I guess we could look at the first "pending" count, and use that. I'll rewrite again.

pretentious7: That solves the error issue too no?

core/src/nats.rs
paulgb: Turns out this isn't used anywhere and I have doubts about whether it's useful as-is, so I've just removed it.

core/src/nats.rs
paulgb: Not entirely, because pending is an attribute of messages themselves. If there are no messages in the stream the error occurs before we have a chance to read a `.pending` value.

I've decided that this is something I'd like to look at once I have my head around streams a bit more, so I've replaced it with a TODO. For our use case the messages are low-frequency enough for this not to happen.

pretentious7 (approved):

PR #185, opened 2022-09-30
Initial docs

paulgb:


docs/docusaurus.config.js
```diff
@@ -0,0 +1,111 @@
...
+    /** @type {import('@docusaurus/preset-classic').ThemeConfig} */
+    ({
+      navbar: {
+        title: 'Spawner',
+        logo: {
+          alt: 'Spawner Logo',
+          src: 'img/logo.svg',
+        },
+        // items: [
+        //   {
```
pretentious7, line 59: are these going to be uncommented later (with tutorial links and such?)

pretentious7 (approved):

docs/README.md
```diff
@@ -0,0 +1,41 @@
+# Website
+
+This website is built using [Docusaurus 2](https://docusaurus.io/), a modern static website generator.
+
+### Installation
+
+```
+$ yarn
```
rolyatmax, line 8: Thoughts on using `npm` instead of `yarn` - just for consistency across our codebases?

docs/docs/concepts.md
```diff
@@ -0,0 +1,32 @@
...
+a unique subdomain under that cluster’s hostname (e.g. `302d996a-6b3a-433b-a1ad-c0ad47ba1575.spawner.dev`).
+
+Spawner is capable of supporting multiple clusters at once, but servers that run backends (**drones**,
+see below) are only capable of belonging to one cluster at a time.
+
+## Controller and Drone
+
+Functionally, Spawner is divided into two parts: the **controller**, and the **drone**.
+
+The controler is the dispatch center, responsible for accepting external requests for backends and
```
rolyatmax, line 26: typo: controler

docs/docs/concepts.md
```diff
@@ -0,0 +1,32 @@
...
+deciding which drone to run them on. The controller is also responsible for routing traffic to the
+appropriate drone by serving DNS, and vouching for drones when they need to request HTTPS certificates.
+
+The drones are the workers that run the actual backends. The Spawner drone codebase runs on each
```
rolyatmax, line 30: Might make sense to swap the drone definition here with the controller definition above. That way you define drones before you reference them in the controller definition.

docs/README.md
paulgb: Yeah, I've been using npm to deploy it, this is just a readme created by the template. I can update it.

docs/docs/concepts.md
paulgb: done

docs/docs/concepts.md
paulgb: fixed

docs/docusaurus.config.js
paulgb: I think so. Usually I don't like checking in commented code, but I'm just trying to move quickly here and this is a good way of documenting the structure until we need it.

Issue #180, opened 2022-09-29
Consolidate `backend_to` maps into one map.


pretentious7:
This will be useful to ensure we don't leak resources after a container is swept.

paulgb:
Rust lifetimes themselves don't change program behaviour, they only place limits on the behaviour that is allowed, so a good way of approaching problems like this is to ask “what behaviour would ensure that all container resources are cleaned up when a container is swept”.

A nice way to do this is to create a `struct` which takes ownership of all container-bound threads/resources, and cleans them up in its drop handler (either explicitly, if we implement `drop`, or implicitly, if we rely on `drop` automatically being called on each member).

This way we would replace all of the `backend_to` [maps](https://github.com/drifting-in-space/spawner/blob/df570118e1d06f6b54a61e7dd8537184ad31d654/drone/src/agent/executor.rs#L46-L50) with a map from backend ID to the new struct.

Note that the drop handler of a [`JoinHandle`](https://docs.rs/tokio/latest/tokio/task/struct.JoinHandle.html) does **not** automatically abort the task, but it's easy to write a wrapper around a `JoinHandle` that would do so.

Some open design questions:
- Should we also combine all container-bound event loops (stats, logs) into one event loop with a select? I'm not persuaded that the result is cleaner, but it's something to consider.
- Should we consolidate the code for (re)connecting to a container into this new `struct`? That way, both initial container creation as well as resuming a connection to an existing container after a restart could use the same code, e.g. `BackendMonitor::for_container(&container_name)`.

paulgb:
I've updated the title to reflect the more specific change that is needed here.

paulgb:
This is now implemented in #186

PR #186, opened 2022-10-01
Local spawner development environment

paulgb:
- [x] Add network so that the drone can access backends when run through docker-compose.
- [x] Add self-signed SSL certs and pass them to drone
- [x] Emit periodic messages for each backend from the agent
- [x] Implement `list-dns` command in CLI
- [x] Add IP API service so that controller can route to the right drone IP
- [x] Implement DNS server
- [x] Configure FF to use local DNS server
- [x] Configure FF to accept self-signed certs


paulgb:
image

Huzzah, it works!

paulgb:
This will close #180, #178, #119.

I'll make a follow-up task for supporting SOA records in DNS.

controller/src/dns/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+                                    );
+                                    continue;
+                                }
+                            };
+                            let value = RData::A(ip);
+                            name_map.insert(key, value);
+                        }
+                        DnsRecordType::TXT => {
+                            let key = (v.cluster.clone(), v.name.clone(), DnsRecordType::TXT);
+                            let value = RData::TXT(TXT::new(vec![v.value]));
```
pretentious7, line 71: if I understand correctly this sets the latest TXT record from the stream right? Would this not cause race conditions with multiple drones? It'd be good to write a test with multiple drones on the same cluster to check at what point this is an issue.

controller/src/dns/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    let sock = UdpSocket::bind(ip_port_pair)
+        .await
+        .context("Binding UDP port for DNS server.")?;
+    fut.register_socket(sock);
+
+    let listener = TcpListener::bind(ip_port_pair)
+        .await
+        .context("Binding TCP port for DNS server.")?;
+    fut.register_listener(listener, Duration::from_secs(10));
```
pretentious7, line 209: the 10 here should be in a more descriptive const (something like MAX_TCP_CONNECTION_DURATION_SECS)

core/src/messages/dns.rs
```diff
@@ -20,17 +27,32 @@ pub struct SetDnsRecord {
...
         )
     }
 }
 
+impl JetStreamable for SetDnsRecord {
+    fn config() -> async_nats::jetstream::stream::Config {
+        async_nats::jetstream::stream::Config {
+            name: Self::stream_name().into(),
+            subjects: vec!["cluster.*.dns.*".into()],
+            max_age: Duration::from_secs(60),
```
pretentious7, line 46: this 60 is the minimum rate at which this message can be emitted without periods in which no dns is set, correct? if so, we should have a way to make sure we don't accidentally cause that, or a test to check for it.

drone/src/agent/backend.rs
```diff
@@ -0,0 +1,145 @@
...
+        dns::{DnsRecordType, SetDnsRecord},
+    },
+    nats::TypedNats,
+    types::{BackendId, ClusterName},
+};
+use tokio::{task::JoinHandle, time::sleep};
+use tokio_stream::StreamExt;
+
+/// JoinHandle does not abort when it is dropped; this wrapper does.
+struct AbortOnDrop<T>(JoinHandle<T>);
```
pretentious7, line 17: ooooooooooooo i like

drone/src/agent/backend.rs
```diff
@@ -0,0 +1,145 @@
...
+            loop {
+                nc.publish_jetstream(&SetDnsRecord {
+                    cluster: cluster.clone(),
+                    kind: DnsRecordType::A,
+                    name: backend_id.to_string(),
+                    value: ip.to_string(),
+                })
+                .await?;
+
+                sleep(Duration::from_secs(50)).await;
```
pretentious7, line 70: why 50?

drone/src/agent/backend.rs
```diff
@@ -0,0 +1,145 @@
...
+        nc: &TypedNats,
+    ) -> Self {
+        let log_loop = Self::log_loop(backend_id, docker, nc);
+        let stats_loop = Self::stats_loop(backend_id, docker, nc);
+        let dns_loop = Self::dns_loop(backend_id, ip, nc, cluster);
+
+        BackendMonitor {
+            _log_loop: AbortOnDrop(log_loop),
+            _stats_loop: AbortOnDrop(stats_loop),
+            _dns_loop: AbortOnDrop(dns_loop),
```
pretentious7, line 46: So, I understand this, but how do you know that these event loops get moved instead of copied? Here it's obvious, but I'm not entirely clear on the mechanism for this. Is it that unless .clone() is called, a non borrow is always a move?

drone/src/agent/docker.rs
```diff
@@ -284,16 +286,40 @@ impl DockerInterface {
...
+                return Ok(ip_addr.parse()?);
+            }
+        }
 
-        let port = inspect.network_settings?.ip_address?;
+        let networks = network_settings
+            .networks
+            .ok_or_else(|| anyhow!("Inspect did not return an IP or networks."))?;
+        if networks.len() != 1 {
+            return Err(anyhow!(
```
pretentious7, line 306: so if Inspect does return an Ip, then having multiple networks does not throw an error. is that intended?

drone/src/agent/executor.rs
```diff
@@ -63,9 +68,10 @@ impl Executor {
             database,
             nc,
             _container_events_handle: Arc::new(container_events_handle),
+            backend_to_monitor: Arc::default(),
```
pretentious7, line 71: So, the idea here is that backend_to_monitor lives as long as someone has a ref to it right? Shouldn't we be able to rework this into a non refcounted type which drops when container reaches a nonliving state?

No change needed for this PR, I'm just wondering.

pretentious7 (changes requested):
Great Stuff :D, I've left some comments regarding unexplained constants and a concern about TXT records, rest looks good.

I really like the BackendMonitor idea.

drone/src/agent/backend.rs
paulgb: Yes, if a value does not implement `Copy`, it has to be cloned explicitly by calling `.clone()`. `JoinHandle`, the underlying type which holds the threads, not only doesn't implement `Copy`, it also does not implement `Clone`. This means that _can't_ be cloned, even explicitly. So if we own a `JoinHandle`, we own the only handle to that task.

(If we wanted to access it in multiple places, we could wrap it in an `Arc` or `Rc`.)

drone/src/agent/backend.rs
paulgb: It hast to be at most 60 so that there's always a value within the ttl of 60 seconds, but we go a bit shorter in case the network is anomalously slow. I'm open to revisiting this approach, but it works for now.

drone/src/agent/docker.rs
paulgb: The underlying assumption here is that if a container has an IP associated with it (meaning that it has an IP on the default bridge network), that IP should be accessible from the drone at that IP. Depending on the setup, this may require that the drone runs on the same network, but overall it seems like a reasonable default assumption.

If the container does not have an IP, we take a more conservative approach and use the IP associated with a network if and only if it's the _only_ network. We could go through this list and look for the “right” network, but if someone is attaching a container to multiple networks, there's various things they might be trying to do, none of which I've particularly thought through at this point, so rather than inferring their desired outcome I'd rather just give an error and have them discuss the use case.

pretentious7: ye that makes sense!

drone/src/agent/executor.rs
paulgb: Yes, it lives as long as there is an `Executor`. It's necessary here because we clone the Executor, but we never actually access `backend_to_monitor` on those clones, so we could refactor to break `Executor` up into a smaller struct that gets passed around and remove this `Arc`.

We could design the control loops to exit when they reach a terminal state, in which case we could just spawn them and leave them unattached (we would drop the `JoinHandle`, which does not abort the underlying task), and not have a problem. The advantages to doing it this way are:

- We ensure that only one of each control loop is running per container. All control loops are owned by a `BackendMonitor`, and a `BackendMonitor` is only kept around by putting it into the "slot" belonging to that backend in `backend_to_monitor`. If we create a second `BackendMonitor` for the same backend, as soon as we insert it into `backend_to_monitor`, the old `BackendMonitor` is dropped. Because it is implemented to abort tasks on dropping, those redundant control loops get aborted.
- It makes it harder to create leaks where control loops aren't terminated after a container exits, because instead of verifying that each control loop self-aborts, we just need to verify that `BackendMonitor` gets dropped.

controller/src/dns/mod.rs
```diff
@@ -0,0 +1,218 @@
...
+                        }
+                    }
+                }
+
+                Ok(())
+            })
+        };
+
+        ClusterDnsServer {
+            name_map,
```
rolyatmax, line 82: Will this `name_map` have any of the updates from the `while` loop above? Aren't those inserts only happening on the cloned `name_map` inside the handle? (Maybe I'm not not understanding `name_map`'s scope correctly?)

paulgb: `name_map` is an `Arc` (**a**tomic **r**eference **c**ounted), which means that every "clone" is really just an additional (reference-counted) pointer to the same underlying map.

core/src/messages/dns.rs
paulgb: I will defer this to #187

drone/src/agent/executor.rs
paulgb: I'm not sure what you mean by having a futures queue in the executor? Do you mean flattening it into a state machine to avoid calling `tokio::spawn` entirely?

drone/src/agent/backend.rs
paulgb: yeah, good call, I've cleaned this up and made the values live in `SetDnsRecord`.

controller/src/dns/mod.rs
paulgb: Fixed.

controller/src/ttl_store/ttl_map.rs
```diff
@@ -26,48 +26,46 @@ impl<K: Hash + Eq + Clone, V> TtlMap<K, V> {
...
         }
-        let expiry = time
-            .checked_add_signed(self.ttl)
-            .expect("Adding ttl should never fail.");
+        let expiry = time + self.ttl;
         self.inner_map.insert(key.clone(), (expiry, value));
         self.queue.push_back((expiry, key));
     }
 
     pub fn get_or_insert<F: FnOnce() -> V>(&mut self, key: K, func: F) -> &mut V {
```
pretentious7, line 46: this takes a function, so should be called get_or_insert_with, no?

controller/src/ttl_store/ttl_map.rs
```diff
@@ -26,48 +26,46 @@ impl<K: Hash + Eq + Clone, V> TtlMap<K, V> {
             ttl,
             inner_map: HashMap::new(),
             queue: VecDeque::new(),
-            last_time: DateTime::<Utc>::MIN_UTC,
+            last_time: SystemTime::UNIX_EPOCH,
```
pretentious7, line 29: what's this for? is `last_time` supposed to be the last_time of access? its not updated anywhere though.

paulgb: Ah good catch, it's supposed to be the time of the last compaction, but I wasn't setting it. I've renamed it `last_compaction` and ensured that it is set.

drone/src/agent/executor.rs
pretentious7: yes but no state machine, just spawn_queue: Queue<tokio::Futures> field in the Executor struct, so start_backend and kill_backend instead of returning futures, return immediately after calling Executor.push_future_onto_queue(docker.kill_backend(...))
kinda like io_uring. But then it'll need callbacks for errors and such so maybe not. There's probably a more ergonomic rust way to achieve that.

pretentious7 (approved):
LGTM!!

PR #189, opened 2022-10-04
rename Spawner -> Plane

paulgb:
The time has come at last!

Let's not be overly concerned with non-production things that may break with this (e.g. the doc pipeline) because there are some things that can't be fixed until the merge is completed. Tests should still work though.


pretentious7 (approved):
LGTM!

PR #195, opened 2022-10-06
Add arm64 arch to docker builds

paulgb:
This probably works, but it's an action so we won't know until we try it.


paulgb:
Hmm, given that the test failed, maybe this is more involved than I thought. Thoughts, @pretentious7?

paulgb:
LGMT! (just realized I can't review because I opened the PR, but feel free to merge)

PR #199, opened 2022-10-06
Add linux-docker-compose file which starts firefox with X11

pretentious7:


pretentious7:
https://docs.docker.com/compose/extends/

I just learned that one can do multiple docker-compose files and extend configs. This will be a lower maintenance of adding this, so I'll submit this PR again after doing that.

paulgb:
I think the approach we should take here is:
- Stick all compose files in a new subdirectory called `sample-config/compose/`
- Put a compose file that runs only Plane and its dependencies in `plane.yml`
- Create compose files for `firefox-guac.yaml`, `firefox-x11.yaml`, and `firefox-xquartz.yaml`
- Add a shell script in `sample-config` that detects the OS and whether an X11 server is running, and chooses which mix-in to apply accordingly.

We should probably implement these CLI args:
- `--no-browser` (don't run the browser container at all)
- `--open` (open a browser to localhost:3000; gives an error if not using guac mix-in)
- `--no-x11` (default to guac even if x11 is available)
- `--build` (build the Plane containers from the repo source instead of pulling `:latest`)

OTOH, I worry that this breaks the docker compose workflow, e.g. being able to run `docker compose build firefox`, because the user will still have to make compose aware of which mix-ins are used.

I realize that this expands the scope of this PR quite a bit. I'm open to keeping this PR limited in scope and figuring out the CLI in future PRs.

Issue #118, opened 2022-08-20
Allow cookie/bearer token-based authentication to individual backends


paulgb:
Since Spawner includes a Layer 7 proxy, it can refuse requests to a backend if they do not contain a bearer token as a cookie or as an `Authorization` header.

A possible implementation would be:

- Add a field to `SpawnRequest` that indicates that a bearer token should be used.
- Add a field to `SpawnRequestResponse` that contains a randomly-generated bearer token.
- Pass the token to the client, which hits an endpoint on the backend (like `__cookie`) that takes a bearer token and returns it as a `Set-Cookie` header.

We could also allow a token to be provided as a `GET` parameter, so that we could generate URLs that can be shared and grant access to a backend. The proxy would strip that parameter, validate it, and return it as a `Set-Cookie` on the response.

paulgb:
Closed by #302

PR #200, opened 2022-10-06
Refactor SpawnRequest to be composed of RouterConfig and ExecutorConfig

pretentious7:
https://github.com/drifting-in-space/plane/issues/161

I also tried a bunch of things with trying an Executor trait with a trait object, but that felt bad,. However I think this is a good start

### DIFF
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -255,6 +255,39 @@ pub struct SpawnRequest {
     pub resource_limits: ResourceLimits,
 }
 
+// enum of supported executors
+pub enum Executors {
+    Docker(Box<dyn Executor>),
+}
+
+pub struct DockerExecutableConfig {
+    pub backend_id: BackendId,
+    pub max_idle_secs: Duration,
+    pub env: HashMap<String, String>,
+    pub metadata: HashMap<String, String>,
+    pub credentials: Option<DockerCredentials>,
+    pub resource_limits: ResourceLimits,
+}
+
+pub trait Executor {
+    fn initialize(&self) -> Result<(), anyhow::Error>;
+}
+
+pub trait Executable<A: Executor> {}
+
+pub struct RouterConfig {
+    pub drone_id: DroneId,
+    pub backend_id: BackendId,
+}
+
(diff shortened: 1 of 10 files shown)


core/src/messages/agent.rs
```diff
@@ -254,6 +255,49 @@ pub struct SpawnRequest {
...
+    pub credentials: Option<DockerCredentials>,
+
+    /// Resource limits
+    #[serde(default = "ResourceLimits::default")]
+    pub resource_limits: ResourceLimits,
+}
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct RouterConfig {
```
paulgb, line 283: I'm inclined to call this `ProxyConfig`, and use it for things that are specific to the proxy. Currently, that's just the `backend_id`, but will include auth tokens (#118) eventually as well.

We could also just leave these fields in the `SpawnRequest` itself -- unlike the executor, I don't see a clear advantage to bundling them together at this point.

paulgb (commented):
Thanks for this, left a couple comments.

core/src/messages/agent.rs
```diff
@@ -254,6 +255,49 @@ pub struct SpawnRequest {
...
+
+#[serde_as]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct DockerExecutableConfig {
+    /// The container image to run.
+    pub image: String,
+
+    /// The timeout after which the drone is shut down if no connections are made.
+    #[serde_as(as = "DurationSeconds")]
+    pub max_idle_secs: Duration,
```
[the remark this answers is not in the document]
pretentious7: ah that's right, yes will do

core/src/messages/agent.rs
pretentious7: I think that's sensible yes, will do (leaving the other fields in the SpawnRequest itself). The current main motivation after all is to have `DockerExecutableConfig` be seperately serializeable.

paulgb (approved):
LGTM, thanks for the changes.

PR #201, opened 2022-10-07
Abhi/compose macos

pretentious7:


pretentious7:
`/opt/X11/bin/Xquartz :0 -listen tcp`

needs this in wrapper shell script

PR #202, opened 2022-10-11
test: add web socket test

### LINKED ISSUE #112: Add WebSocket tests
The Ava implementation used a WebSocket client/server library to test this, we should do the same with the new Rust tests. Tricky part may be finding a good simple WebSocket implementation that also supports passing our own TLS certificates and custom DNS resolution.

lalo:
Please squash commits when merging.

Closes #112


paulgb (approved):
Thanks Eduardo, LGTM!

PR #212, opened 2022-10-12
add subscribe jetstream ordered

pretentious7:


core/src/nats.rs
```diff
@@ -374,7 +374,38 @@ impl TypedNats {
...
+            stream,
+            _ph: PhantomData::default(),
+        })
+    }
+
+    pub async fn subscribe_jetstream_ordered<T: JetStreamable>(
+        &self,
+        subject: SubscribeSubject<T>,
+    ) -> Result<JetstreamSubscription<T>> {
+        //TODO: use ordered consumers for this
```
pretentious7, line 389: I was trying to use one of these: https://docs.rs/async-nats/latest/async_nats/jetstream/consumer/type.OrderedPushConsumer.html#
with a wrapper enum implementing .next(), but I had an absolutely miserable time and couldn't figure out how to get the compiler to stop complaining about lifetimes, so I've kept it this way.
another option is to have an ordered: Bool key that toggles the max_ack_pending instead of another function, wdyt @paulgb

core/src/nats.rs
```diff
@@ -369,12 +370,15 @@ impl TypedNats {
                 deliver_policy: DeliverPolicy::All,
                 filter_subject: subject,
                 deliver_subject,
-                ..async_nats::jetstream::consumer::push::Config::default()
+                max_ack_pending: 1, // NOTE: IF YOU REMOVE THIS OR CHANGE THE VALUE,
```
paulgb, line 373: This is important but doesn't need to be all-caps

paulgb (commented):
LGTM, just use regular casing and this is ready

rolyatmax (approved):

PR #215, opened 2022-10-18
Report drone version in liveness ping

paulgb:
`CARGO_PKG_VERSION` is [set by cargo at compile time](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates). The Rust [`env` macro](https://doc.rust-lang.org/std/macro.env.html) will read a value at compile time and insert it into the source.

This also removes the `capacity` value from the liveness ping. It was a placeholder that we never used, and probably does not represent how we want to do scheduling.

### DIFF
--- a/controller/src/scheduler.rs
+++ b/controller/src/scheduler.rs
@@ -77,6 +77,7 @@ impl Scheduler {
 #[cfg(test)]
 mod tests {
     use super::*;
+    const PLANE_VERSION: &str = env!("CARGO_PKG_VERSION");
 
     fn date(date: &str) -> DateTime<Utc> {
         DateTime::parse_from_rfc3339(date).unwrap().into()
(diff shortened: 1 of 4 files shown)


controller/src/scheduler.rs
```diff
@@ -77,6 +77,7 @@ impl Scheduler {
 #[cfg(test)]
 mod tests {
     use super::*;
+    const PLANE_VERSION: &str = env!("CARGO_PKG_VERSION");
```
rolyatmax, line 80: Is the CARGO_PKG_VERSION the same as the commit hash?

paulgb: No, it is the version string from the [cargo toml](https://github.com/drifting-in-space/plane/blob/main/drone/Cargo.toml#L3).

I'd like to move away from deploying based on the sha hash, and instead to use numbered versions. I'm not ready to do that yet, but this will help when we do.

PR #217, opened 2022-10-19
Abhi/compose cli

pretentious7:
This adds the cli proposed in https://github.com/drifting-in-space/plane/pull/199

I'm still working on this (need to add in certs and make the guac-firefox default open up to plane.dev)
but I wanted some comments on the cli, the way it works is:
`./plane-dev -[lmxg]` where
-l : linux
-m : mac
-x : x11
-g : guac

I'd have nice long options, but can't do that without GNU getopts, which isn't on macs by default, so I'd rather not.
what'd you guys think?


paulgb:
Do we need getopts for this? If there are only four options, can't we just do string comparisons?

I'd rather support only long options than only short options, especially since these are all pretty short.

pretentious7:
well I also wanted to add a flag for compose up vs down, and I think it makes the script more readable, but I suppose it makes the *command* less readable, so ye I'll just match on the longopts (does remove the option to do -xl tho, which is nice and short)

paulgb:
One thing I'm a bit worried about here is that we break `docker compose down`, `docker compose build`, etc functionality by doing this. One solution is to:

- Have a default `docker-compose.yml` file that includes the guac mix-in
- Have the CLI overwrite the `docker-compose.yml` with the appropriate mix-in
- Instruct the user to run `docker compose up` after doing so

In particular, it's important to me that the [getting started](https://plane.dev/docs/getting-started) guide not introduce any additional complexity with the CLI, which is why I like keeping a default compose file with the guac mix-in around. Then we can have another section on local development that talks about x11.

pretentious7:
I think I know how to deal with that, looking at xquartz docs, but I'll have to make sure.

pretentious7:
hm yea i think that works perfect, I'll have the firefox service be defaulted to guac in the compose file, and -f firefox-x11.yml to override it in the x11 cases. this keeps `docker-compose up / docker-compose down` useful.

pretentious7:
In that case, if this is to be slightly off the beaten track, can we keep the current getopts handling of cli args? I'll make the `usage` function nicer.

pretentious7:
ahh I think this is ready to be merged, just needs some testing on macos first. (I'll add in additional docs in a future pr, the current workflow is unaffected aside from CD'ing to a different dir, which I changed in the docs)

could one of you try it on mac and make sure I didn't break anything?  try `./cli/plane-dev.sh -mg` from the sample-config dir or just `docker compose up` in the plane/sample-config/compose dir

paulgb:
When I run that, I get:

```
cli/plane-dev.sh: line 63: xdg-open: command not found
```

Once it starts, I get this:

```
Error response from daemon: Invalid address 172.16.238.10: It does not belong to any of this network's subnets
```

sample-config/x11-firefox/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM debian:11-slim
+RUN apt-get -y update && apt-get -y install firefox-esr ca-certificates
+
+#RUN echo "firefox https://plane.dev" > defaults/autostart
```
paulgb, line 4: rm dead code

sample-config/x11-firefox/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Override default preferences.
+COPY firefox-common-config/firstrun.js /usr/lib/firefox/browser/defaults/preferences/firstrun.js
+
+# Copy init file, which starts the DNS server.
+#COPY guac-firefox/custom-cont-init.d /custom-cont-init.d
```
paulgb, line 10: rm dead code

sample-config/guac-firefox/policies.json
```diff
@@ -4,6 +4,7 @@
         "Certificates": {
             "Install": [
                 "plane.crt"
+                "usr/local/share/ca-certificates/plane.crt"
```
paulgb, line 7: do we need both?

sample-config/guac-firefox/Dockerfile
```diff
@@ -1,16 +1,17 @@
...
 # Set home page.
 RUN echo "firefox https://plane.dev" > defaults/autostart
 
 # Override default preferences.
-COPY firefox/firstrun.js /usr/lib/firefox/browser/defaults/preferences/firstrun.js
+COPY firefox-common-config/firstrun.js /usr/lib/firefox/browser/defaults/preferences/firstrun.js
 
 # Copy init file, which starts the DNS server.
-COPY firefox/custom-cont-init.d /custom-cont-init.d
+# COPY guac-firefox/custom-cont-init.d /custom-cont-init.d
```
paulgb, line 14: rm dead code

sample-config/guac-firefox/Dockerfile
```diff
@@ -1,16 +1,17 @@
 FROM lscr.io/linuxserver/firefox:latest
 
-RUN apk add bind-tools
-RUN apk add dnsmasq
+#RUN apk update
```
paulgb, line 3: rm dead code

sample-config/guac-firefox/policies.json
```diff
@@ -4,6 +4,7 @@
         "Certificates": {
```
paulgb, line 4: should this be a symlink to the other policy json?

sample-config/compose/plane.yml
```diff
@@ -0,0 +1,84 @@
+version: "3.9"
+services:
+  drone:
+    container_name: plane-drone
+    image: ghcr.io/drifting-in-space/plane-drone:latest
```
paulgb, line 5: the `image` declared here seems to be ignored in favor of building, we probably want to use the image by default.

The ideal behavior is to use the image unless `docker compose build` is explicitly called, but I'm not sure if `docker compose` allows that.

sample-config/compose/plane.yml
```diff
@@ -0,0 +1,84 @@
...
+      /etc/plane/controller.toml
+    networks:
+      plane:
+        ipv4_address: 172.16.238.11
+
+  nats:
+    container_name: plane-nats
+    image: nats:latest
+    command:
+      "-DV --jetstream"
```
paulgb, line 34: `-DV` is super chatty and makes it impossible to read anything else in the logs, let's not use it as the default.

paulgb (commented):
I like this as long as we can make it 100% seamless on Mac. I do worry about the support burden though, there's a lot of moving parts and unknowns.

pretentious7:
> When I run that, I get:
> 
> ```
> cli/plane-dev.sh: line 63: xdg-open: command not found
> ```
> 
> Once it starts, I get this:
> 
> ```
> Error response from daemon: Invalid address 172.16.238.10: It does not belong to any of this network's subnets
> ```

Ah can you try running it after a `docker compose down`? Looks like the invalid address issue there comes from it using the wrong network.

I also totally forgot the freedesktop stuff wasn't POSIX. Will fix.

pretentious7:
I'll add in the x11 support for mac once I have the aws VM set up, I think this should be good to merge for now though.

paulgb:
This still has the issue we discussed where it wants to build the drone locally when I don't have the image already, and we don't have a workaround for that without causing the other images not to build. I'm inclined to just go back to commenting out the build section of the yaml, I think.

docs/docs/getting-started.md
```diff
@@ -37,7 +37,7 @@ Use docker-compose to start a local instance of Plane:
 
 ```bash
 cd plane/sample-config/compose
-docker compose up
+docker compose pull && docker compose up
```
paulgb, line 40: nit: let's just do these on separate lines, i.e.

```bash
cd plane/sample-config/compose
docker compose pull
docker compose up
```

That way it's easier to debug because the output is directly associated with the command that caused it.

paulgb (commented):
One small doc change but otherwise looks good!

PR #218, opened 2022-10-19
Add SOA record support on DNS

paulgb:


controller/src/plan.rs
```diff
@@ -20,10 +24,28 @@ impl ControllerPlan {
         let nats = config.nats.connect_with_retry().await?;
 
         let scheduler_plan = config.scheduler.map(|_| SchedulerPlan);
-        let dns_plan = config.dns.map(|options| DnsPlan {
-            options,
-            nc: nats.clone(),
-        });
+        let dns_plan = if let Some(options) = config.dns {
```
rolyatmax, line 27: Minor: would using a `map` here make it slightly easier to read? Only one if/else instead of nested if/elses?

paulgb: Good instinct! But in this case the `if/else` is more readable because I use the `?` operator within it. `?` provides a short-circuit return for the function it is run inside, so if you use `?` within a closure passed to a map, it doesn't short-circuit return for the outer method, just the closure, which you then need to pop up again. It gets to be kind of messy and, though a bit more verbose, I prefer just using `if let`/`else` in those situations.

controller/src/plan.rs
```diff
@@ -20,10 +24,28 @@ impl ControllerPlan {
...
+            let soa_email = if let Some(soa_email) = options.soa_email {
+                let soa_email = format_rname(&soa_email).context(
```
pretentious7, line 29: might it be better to sanitize the configuration in a separate function? or pass what this does to serde somehow?

pretentious7 (commented):
LGTM!

controller/src/plan.rs
paulgb: `format_rname` does a basic check (ensures that there is an `@`, basically). Full email validity checking is [famously crazy](http://www.ex-parrot.com/~pdw/Mail-RFC822-Address.html) and not really necessary here.

PR #219, opened 2022-10-20
subscribe to drone status message as a regular subscription - not a jetstream

rolyatmax:
I noticed that Plane was creating a Jetstream here when [the equivalent code in the platform](https://github.com/drifting-in-space/p2/blob/main/jamsocket/src/platform/mod.rs#L32-L34) is just using a regular nats subscription. Could this be related?


paulgb:
Here's [where we publish to that stream](https://github.com/drifting-in-space/plane/blob/main/drone/src/agent/mod.rs#L106) - we publish as a jetstream (basically, we wait for an ack).

Can you be more descriptive about what the bug here is?

paulgb (approved):
LGTM as discussed offline.

PR #221, opened 2022-10-24
Fix route duplication issue

paulgb:
We should be synchronizing the `route` table with drone liveness. I'll follow up with a PR that does that, and tests it. For now, this fixes a correctness issue by ensuring that only one route can point to the same local IP at any given time.


rolyatmax (approved):
LGTM

PR #223, opened 2022-10-27
Abhi/add scheduler tests for nats restart

pretentious7:


pretentious7:
I think restarting nats during the tests tests nats more than it tests our code, so doesn't need to be here.

PR #224, opened 2022-10-28
fix bug and notif for other unhandled errors

pretentious7:
fixes https://github.com/drifting-in-space/plane/issues/193


paulgb (approved):
Thanks, this will be a good bug to close.

pretentious7:
I just changed it to update the state variable so it can remove the monitor and listeners, going to give it another quick test then merge.

PR #226, opened 2022-10-30
Remove SetAcmeDnsRecord in favor of SetDnsRecord

paulgb:
This updates the certificate creation logic to use the message created in #162.

This technically breaks the rust/NATS API, but I'm not marking it as such because the relevant APIs were never used. There was nothing to receive those messages.


drone/src/cert/mod.rs
```diff
@@ -76,17 +75,15 @@ pub async fn get_certificate(
             .ok_or_else(|| anyhow!("No authorization value."))?;
 
         tracing::info!("Requesting TXT record from platform.");
-        let result = nats
-            .request(&SetAcmeDnsRecord {
+        nats
+            .publish_jetstream(&SetDnsRecord {
```
pretentious7, line 79: why switch from req/res here?

pretentious7 (approved):
just a little question about req/res vs pub/sub here, otherwise lgtm

drone/src/cert/mod.rs
paulgb: `SetAcmeDnsRecord` dates from the days where the source-of-truth for DNS was the database. After writing to the database, the DNS server would return true and all DNS servers would see the update (from the DB).

Now, Jetstream itself is the source of truth, so instead of waiting for confirmation from a DNS server, we wait for confirmation from the Jetstream server. It's very similar in principle, the difference is that async_nats is aware of the Jetstream protocol's ack and has a special API for waiting for it.

PR #230, opened 2022-10-30
Add timings for certain startup tasks

paulgb:
This creates a `Timer` struct, and uses it to record timings of a number of things:
- Scheduling a backend
- Pulling an image
- Creating a container
- Starting a container

Ideally I'd like to be able to generate a full flamegraph from scheduler request to backend ready, but this is a start in that direction.

### DIFF
--- a/drone/src/agent/docker.rs
+++ b/drone/src/agent/docker.rs
@@ -11,7 +11,7 @@ use bollard::{
     system::EventsOptions,
     Docker, API_DEFAULT_VERSION,
 };
-use plane_core::messages::agent::ResourceLimits;
+use plane_core::{messages::agent::ResourceLimits, timing::Timer};
 use std::{collections::HashMap, net::IpAddr, time::Duration};
 use tokio_stream::{wrappers::IntervalStream, Stream, StreamExt};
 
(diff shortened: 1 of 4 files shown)


drone/src/agent/docker.rs
```diff
@@ -231,12 +231,12 @@ impl DockerInterface {
         })
     }
 
-    #[allow(unused)]
     pub async fn pull_image(
         &self,
         image: &str,
         credentials: &Option<DockerCredentials>,
     ) -> Result<()> {
+        let timer = Timer::new();
```
pretentious7, line 239: I really like the idea of this, is there any way to implement it without the extra noise? would an attribute macro work?

pretentious7:
I like this, it's going to be useful enough with the perf tuning that I think we can merge it as is, but it'd be cool to be able to do #[timed] attribute or something like that.

pretentious7 (approved):

drone/src/agent/docker.rs
paulgb: I wonder if any timing info is built into tracing's built-in [attribute macro](https://docs.rs/tracing-attributes/latest/tracing_attributes/attr.instrument.html). Eventually we should probably instrument the code with that as it's an ergonomic way to add spans that work across await calls.

PR #232, opened 2022-11-02
CORS header for 404

paulgb:
When a backend is running, the backend can set its own CORS headers. However, after a backend is shut down, it defaults back to the strict default CORS settings. This change sets the CORS header for the 404 page to accept all origins.

Security-wise, this doesn't expose anything that wasn't exposed before. The 404 page returned doesn't look at authentication headers or cookies, it simply reveals that a backend no longer exists, which any internet-connected device could already determine.


rolyatmax (approved):
Good catch

PR #234, opened 2022-11-02
Abhi/caching

pretentious7:
adds caching for the docker-compose files, for now only S3 is supported; eventually plan to add local caching support for compose.

### DIFF
--- a/drone/Dockerfile
+++ b/drone/Dockerfile
@@ -1,4 +1,4 @@
-FROM rust:latest as build
+FROM rust:buster as build
 
 WORKDIR /work
 
--- a/drone/Dockerfile
+++ b/drone/Dockerfile
@@ -1,10 +1,24 @@
-FROM rust:buster as build
+FROM rust:latest as build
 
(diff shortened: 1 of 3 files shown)


drone/Dockerfile
```diff
@@ -1,10 +1,24 @@
-FROM rust:buster as build
+FROM rust:latest as build
```
paulgb, line 1: Why do this in the drone but not controller?

pretentious7: oops oversight

paulgb:
Have you managed to get this to work in `docker compose`?

I set the environment variables, ran `docker compose build`, then modified a file by adding a newline and ran again, and the build still took 5 mins. Can we speed that up?

paulgb:
D'oh deleted a comment because I was using the wrong vars.

paulgb:
Cool, confirmed that this speeds up the docker compose by 5x. Nicely done!

paulgb (approved):

Issue #129, opened 2022-08-31
Scheduler should support "draining" drones


paulgb:
We need a way to indicate that a particular drone should not be scheduled on.

paulgb:
Closed by #228

PR #240, opened 2022-11-07
Support drone draining (#129)

paulgb:
This implements a basic version of drone draining.

- Adds a `DrainDrone` message type. The response of `DrainDrone` is a unit type (`()`), used to acknowledge that the message was received by the drone it was intended for.
- Adds a listener for `DrainDrone` messages to the agent.
- Adds a `watch` channel that the listener can use to send changes to the readiness value to the `ready_loop`.
- Removes drones from the scheduler's view when they indicate that they are not ready.

Additionally, I made some changes in order to test it:

- `timeout` returns a `Result<T>` instead of `T`, and returns an `Err()` instead of panicking.
- `expect_status_message` takes an additional argument which is asserted to be the readiness status of the drone.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -22,7 +22,7 @@ jobs:
         cache-on-failure: true
 
     - name: Install cargo-nextest
-      run: cargo install cargo-nextest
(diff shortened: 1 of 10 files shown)


paulgb:
Good idea, done.

PR #243, opened 2022-11-07
move everything over to 16 core machines!

pretentious7:


.github/workflows/tests.yml
```diff
@@ -13,7 +13,7 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest-16-cores
+    runs-on: ubuntu-latest-16-cores-16-cores
```
paulgb, line 16: We can't afford 256 cores ;)

paulgb (commented):
One typo, rest looks good

.github/workflows/tests.yml
pretentious7: My sed skills have failed me :laughing:

pretentious7:
O thanks

paulgb (approved):

paulgb:
Is there any measurable difference here? I wonder if builds are so network IO-bound (especially now that we have caching) that it's a wash?

pretentious7:
for some reason these last builds didn't have anything in the cache.

PR #247, opened 2022-11-08
Drone reports number of running backends

paulgb:


drone/src/agent/mod.rs
```diff
@@ -159,7 +163,7 @@ pub async fn run_agent(agent_opts: AgentOptions) -> NeverResult {
 
     nats.publish(&request).await?;
 
-    let executor = Executor::new(docker, db, nats.clone(), ip, cluster.clone());
+    let executor = Executor::new(docker, db.clone(), nats.clone(), ip, cluster.clone());
```
rolyatmax, line 166: Is it possible (or even preferable?) to pass `db` by reference here instead of cloning?

paulgb: The clone here is cheap, because the underlying data is an `Arc<...>`. We could probably technically get away with a reference here because we don't need a mutable version of the data and only need it in this function, but I'm not sure.

rolyatmax (approved):
LGTM

rolyatmax:
Do we need to bump the plane version with this?

paulgb:
Version bumped, good call.

PR #248, opened 2022-11-09
Do not use sccache when S3 bucket url not set in $PLANE_SCCACHE_BUCKET (fixes slow docker build times when that variable is unset)

pretentious7:
this change just removes the use of sccache when $PLANE_SCCACHE_BUCKET is not set (its slow to not have this set as sccache makes http requests and waits for them to fail)


paulgb:
Please give the PR a more descriptive name. When you create a PR from more than one commit, GitHub tries to infer a name from the branch and it's usually not a great name.

Is this a standard way of doing a bash conditional? I'm not familiar with it, maybe a standard `if` / `else` would be clearer to read?

pretentious7:
I generally use this cause it puts it onto one line (otherwise one has to use the EOF thing, which breaks syntax highlighting), && happens when exit code is 0 || when exit code is non zero (from [ which is alias for /usr/bin/test))

I can switch it to the EOF style if that's preferable

PR #249, opened 2022-11-10
 use test command explicitly instead of [ ] alias in docker run and remove "" as default arg in compose

pretentious7:
using [] does not work as docker interprets [ as the start of an "exec" form. 
"" as default arg in compose does not work because compose interprets "" literally and passes in "" instead of the null string.


paulgb (approved):
Good find!

PR #253, opened 2022-11-12
Executor refactoring

### LINKED ISSUE #144: Backend termination should go through the regular state machine step
Currently, we terminate nodes by killing the container. This is picked up by the state machine step, since it listens for changes to the container, but it's not ideal.

Each backend running on a drone [has a channel](https://github.com/drifting-in-space/spawner/blob/86f1f720e51d67339adb6e9d110ec685414d8010/drone/src/drone/agent/executor.rs#L214) through which we can interrupt the current state machine step. It's used to alert the state machine to (potential) changes that happen externally (currently, this is only events from Docker). Maybe we want to reuse this channel.

paulgb:
- Creates `Engine` trait, which abstracts over the functionality we get from Docker such that we could swap in another executor.
- Adds `Terminated` state for backends which differentiates backends that are terminated via the API from backends that self-terminate. (NB. this affects the external API; closes #144)
- Version bump to `0.3.4`

Minor changes:
- Disables default dependencies for `chrono` for slightly faster build times.
- `integration_test` macro now expects a function which returns nothing, instead of returning a result. Instead, the test function itself should panic. This gives us more useful tracebacks.
- Peppered some context around tests that were flakey. The flakiness turned out to be a result of my environment and the amount of parallelism I was using (resource exhaustion), but I've kept it in because it's harmless and potentially useful.
- Modified devcontainer to add `cargo-workspaces` crate.


drone/src/agent/engines/docker/mod.rs
```diff
@@ -316,4 +307,56 @@ impl Engine for DockerInterface {
...
+
+        let running = state
+            .running
+            .ok_or_else(|| anyhow!("State found but no running field for container."))?;
+
+        if running {
+            Ok(EngineBackendStatus::Running)
+        } else {
+            match state.exit_code {
+                None => Ok(EngineBackendStatus::Terminated),
```
rolyatmax: Is this correct? Don't terminated processes still have exit codes?

drone/src/agent/backend.rs
```diff
@@ -83,21 +83,11 @@ impl BackendMonitor {
...
 
             while let Some(v) = stream.next().await {
-                match v {
-                    Ok(v) => {
-                        if let Some(message) = DroneLogMessage::from_log_message(&backend_id, &v) {
-                            nc.publish(&message).await?;
-                        }
-                    }
-                    Err(error) => {
-                        tracing::warn!(?error, "Error encountered forwarding log.");
```
rolyatmax, line 99: Did we lose this warning?

drone/src/agent/engine.rs
```diff
@@ -16,13 +16,13 @@ pub enum EngineBackendStatus {
...
     /// The backend exited on its own without indicating failure.
-    Finished,
+    Exited,
 
     /// The backend exited on its own with a failure state.
     Failed,
 
     /// The backend was terminated by external forces.
-    Terminated,
+    Swept,
```
rolyatmax, line 25: Why this change? To me, "Terminated" feels like the right terminology for the engine level, while "Swept" feels more Plane-specific.

paulgb: I agree, it was an attempt at consistency but now that you mention it, I think it's more confusing to use “swept” here.

drone/src/agent/backend.rs
paulgb: It will result in “Log loop terminated” instead (the `log_stream` terminates if it encounters a parse error). My thinking is that we can add more logging if we start encountering that message.

drone/src/agent/engines/docker/mod.rs
paulgb: I haven't tested this, but my understanding is:

- If a process receives `SIGTERM` (graceful termination) and then exits, it returns an exit code.
- If a process is killed by `SIGKILL`, it does not have an exit code.
- Docker attempts to kill a process first gracefully by sending a `SIGTERM`, then forcefully by sending a `SIGKILL` after ~10s.

The underlying problem here is that from Docker state alone, we can't infer _why_ a container was killed anyway. Fortunately, we only use this method when we are either in the `Starting` and `Ready` states; if a container is swept or terminated by Plane, it has already advanced to another state and we don't need to inquire about its status from Docker.

rolyatmax: It looks like the first answer here might be correct: https://stackoverflow.com/questions/31297616/what-is-the-authoritative-list-of-docker-run-exit-codes

When I tested this, `docker inspect` returned an exit code of `0` while it was still running (I guess it's a non-nullable value?) and when I ran `docker kill`, the `docker inspect` command showed the exit code as `137`.

drone/src/agent/backend.rs
```diff
@@ -107,38 +92,20 @@ impl BackendMonitor {
...
-                        )?)
-                        .await?;
-                        prev_stats = cur_stats;
-                    }
-                    Err(error) => {
-                        tracing::warn!(?error, "Error encountered sending stats.")
-                    }
-                }
+
+            while let Some(stats) = stream.next().await {
```
pretentious7, line 107: what exactly happens if stream.next() gives None?

drone/src/agent/backend.rs
```diff
@@ -107,38 +92,20 @@ impl BackendMonitor {
...
+                nc.publish(&stats).await?;
```
pretentious7, line 108: so this error is in a spawn that we don't handle, so does it just quietly drop any error?

drone/src/agent/engine.rs
pretentious7: +1 for terminated

drone/src/agent/engines/docker/mod.rs
paulgb: Good find. I think you're right that it's not a nullable value; bollard is generated from the OpenAPI spec and I don't know OpenAPI that well, but it doesn't seem to indicate that it's a required field. It seems to be a case of the field always being populated but not required by the API spec.

Unfortunately it means that a container that was killed is indistinguishable from one that exited successfully, unless that's represented elsewhere.

So `None` here is probably an impossible scenario, but we need to handle it somehow to satisfy the typechecker, and I don't think it makes sense to panic since:
- Getting a `None` here isn't particularly alarming
- Treating it as terminated seems like a reasonable thing to do

drone/src/agent/executor.rs
```diff
@@ -37,34 +37,73 @@ impl<T, E: Debug> LogError for Result<T, E> {
...
 
-#[derive(Clone)]
-pub struct Executor {
-    docker: DockerInterface,
+#[derive(Debug, PartialEq, Eq)]
+enum Signal {
+    /// Tells the executor to interrupt current step to recapture an external status
+    /// change. This signal is sent after an engine detects an external status change,
+    /// e.g. a backend has terminated itself with an error.
+    Interrupt,
```
pretentious7, line 45: I like this idea, but doesn't it require that the state changes be atomic? for example if the interrupt happens after some non undoable side effect, could we run into incosistent state?

paulgb: That's a good point. For now there's no correctness issue, because interrupts are only fired if a container has died anyway, so the executor immediately discovers that fact and jumps to the appropriate terminal state. But the interrupt itself technically doesn't convey to the executor that the backend is dead, it just nudges the executor to discover that on its own, so `Terminate` doesn't seem appropriate here.

(The reason the signal doesn't convey that the backend has shut down is that we also want to pick up that fact if the drone is resumed and a container has died while the drone was not receiving events from Docker.)

drone/src/agent/backend.rs
paulgb: Changed to log it.

drone/src/agent/backend.rs
paulgb: Changed to log it.

drone/src/agent/engines/docker/mod.rs
rolyatmax: Might make sense to add a comment or something, as it implies to the reader that this is how we can see if an container has terminated. Not a blocker though.

rolyatmax (approved):

pretentious7:
LGTM!

PR #266, opened 2022-11-20
Update async_nats version

paulgb:


rolyatmax:
What changes does this end up pulling in?

rolyatmax (approved):

paulgb:
> What changes does this end up pulling in?

The big ongoing change in nats_async is a move towards concrete error types, which will allow us to refactor error handling to simplify it (e.g. remove the to_anyhow method). This only introduces one concrete error that we care about, but it’s a start.

PR #269, opened 2022-11-22
Abhi/add pull policy docker

pretentious7:
working on tests (also need to remove the nats cluster commits, leaked in accidentally)

### DIFF
--- a/drone/src/agent/docker.rs
+++ b/drone/src/agent/docker.rs
@@ -40,11 +40,15 @@ impl<T> AllowNotFound for Result<T, bollard::errors::Error> {
         match self {
             Ok(_) => Ok(()),
             Err(bollard::errors::Error::DockerResponseServerError {
-                status_code: 404, message,
+                status_code: 404,
+                message,
             }) => {
-                tracing::warn!(?message, "Received 404 error from docker, possibly expected.");
+                tracing::warn!(
+                    ?message,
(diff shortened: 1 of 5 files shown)


drone/src/agent/docker.rs
```diff
@@ -327,6 +331,15 @@ impl DockerInterface {
         Ok(ip.parse()?)
     }
 
+    /// returns true if image exists, returns false in all other cases
+    pub async fn image_exists(&self, image: &str) -> bool {
+        //not sure if it's faster to do this or check if image in docker.list_images
```
paulgb, line 336: This seems fine to me, and has the advantage of not being linear in the number of images.

drone/src/agent/executor.rs
```diff
@@ -262,16 +262,30 @@ impl Executor {
...
-                        &spawn_request.executable.image,
-                        &spawn_request
-                            .executable
-                            .credentials
-                            .as_ref()
-                            .map(|d| d.into()),
-                    )
-                    .await?;
+                let should_pull_image;
+                match spawn_request.executable.pull_policy {
```
paulgb, line 266: match expressions can return a value! Instead of declaring should_pull_image and then setting it, you can set it to the value of the match expression

paulgb (commented):
I’ll wait until you resolve the git stuff to approve, but lgtm so far!

PR #274, opened 2022-11-29
State replication refactoring

paulgb:
Motivation, design, and broader plan discussed in #278.

This PR introduces a concept of a "view", which is a data structure that is synchronized so that it mirrors the state of the overall system.

`core/src/views/mod.rs` contains three views implemented in this PR, arranged as a three-level tree.

image

- `SystemView` represents the entire state of a multi-tenant system. 
- `ClusterView` represents the state of a cluster, consisting of a map from drone ID to `DroneView`, and from backend ID to the `DroneId` of the drone it is running on. The latter data structure is used for routing.
- `DroneView` represents the state of an individual drone, including the state of running backends.

The views themselves are not connected with the outside world. The only external mutator is `SystemView::update_state`, which takes a `StateUpdate` message and updates itself, dispatching down the tree as appropriate.

`UpdateState` messages are an `enum`, currently consisting of `BackendStatus` (to represent backend status changes), and `DroneStatus` (to represent drone heartbeats and status changes).

In addition to the views, a `SystemViewReplica` struct is implemented in `core/src/views/replica.rs`. This combines a (shared) instance of `SystemView` and a NATS subscription, and updates the shared instance when updates come in over NATS.

`SystemViewReplica` also provides a `snapshot` method, which is an `async` method that goes through the same initialization process of a `SystemViewReplica`, but stops as soon as it has an up-to-date version of all subjects and returns a `SystemView`. This `SystemView` is a frozen "snapshot" of the system state; the NATS subscription used to create it is dropped and no further updates are applied to it.

As of this PR, the controller does not yet use a `SystemView`; scheduling and DNS use the previous message patterns, which are preserved. But the drone has been modified to emit both the old message types and the new `UpdateState` messages, and the CLI has been rewritten to use the `SystemView` for `list-dns` and `list-drones` commands.

### DIFF
--- a/core/src/messages/state.rs
+++ b/core/src/messages/state.rs
@@ -4,8 +4,8 @@ use crate::{
     types::{BackendId, ClusterName, DroneId},
 };
 use serde::{Deserialize, Serialize};
+use uuid::Uuid;
 use std::net::IpAddr;
-use time::OffsetDateTime;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub enum StateUpdate {
@@ -29,14 +29,15 @@ pub enum StateUpdate {
         backend: BackendId,
         state: BackendState,
     },
-    /// A message sent by a consumer who thinks a drone may be dead. If the
-    /// message comes back and the last timestamp of the drone in question
-    /// is still the last heartbeat on record, the drone can be assumed dead.
-    MaybeDead {
-        cluster: ClusterName,
-        drone: DroneId,
-        last_timestamp: OffsetDateTime,
-    },
+    Landmark {
+        uuid: uuid::Uuid,
+    }
+}
+
+impl StateUpdate {
+    pub fn landmark() -> StateUpdate {
+        Self::Landmark { uuid: Uuid::new_v4() }
+    }
 }
 
 impl TypedMessage for StateUpdate {
@@ -60,11 +61,7 @@ impl TypedMessage for StateUpdate {
                 drone.id(),
                 backend.id()
             ),
-            StateUpdate::MaybeDead { cluster, drone, .. } => format!(
-                "cluster.{}.drone.{}.sm.dead",
-                cluster.subject_name(),
-                drone.id()
-            ),
+            StateUpdate::Landmark { .. } => "state_update.landmark".into(),
         }
     }
 }
@@ -78,18 +75,18 @@ impl JetStreamable for StateUpdate {
         async_nats::jetstream::stream::Config {
             name: Self::stream_name().into(),
             max_messages_per_subject: 1,
-            subjects: vec!["cluster.*.drone.*.sm.>".into()],
+            subjects: vec!["cluster.*.drone.*.sm.>".into(), "state_update.landmark".into()],
             ..async_nats::jetstream::stream::Config::default()
         }
     }
 }
 
 impl StateUpdate {
-    pub fn cluster(&self) -> &ClusterName {
+    pub fn cluster(&self) -> Option<&ClusterName> {
         match self {
-            StateUpdate::DroneStatus { cluster, .. } => cluster,
-            StateUpdate::BackendStatus { cluster, .. } => cluster,
-            StateUpdate::MaybeDead { cluster, .. } => cluster,
+            StateUpdate::DroneStatus { cluster, .. } => Some(cluster),
+            StateUpdate::BackendStatus { cluster, .. } => Some(cluster),
+            StateUpdate::Landmark { .. } => None,
         }
     }
 }
--- a/core/src/messages/state.rs
+++ b/core/src/messages/state.rs
@@ -64,10 +64,7 @@ impl JetStreamable for StateUpdate {
         async_nats::jetstream::stream::Config {
             name: Self::stream_name().into(),
             max_messages_per_subject: 1,
-            subjects: vec![
-                "cluster.*.drone.*.sm.>".into(),
-                "state_update.landmark".into(),
-            ],
+            subjects: vec!["cluster.*.drone.*.sm.>".into()],
             ..async_nats::jetstream::stream::Config::default()
         }
     }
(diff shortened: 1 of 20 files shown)


core/src/views/mod.rs
```diff
@@ -0,0 +1,288 @@
...
+
+    pub fn num_backends(&self) -> usize {
+        self.backends.len()
+    }
+}
+
+#[derive(Default)]
+pub struct ClusterView {
+    /// Maps active backends to the drone they run on.
+    pub routes: BTreeMap<BackendId, DroneId>,
```
pretentious7, line 42: why Btrees instead of hashmaps?

paulgb: I wanted to preserve the (recent) behaviour that `list-dns` and `list-drones` return drones/backends in alphabetical order.

I suppose it would be slightly more performant to use `HashMap` and sort only in the CLI after they are retrieved, but this is simple for now and I don't think it will be a noticeable performance difference.

core/src/nats.rs
```diff
@@ -12,19 +13,37 @@ use async_nats::{Client, Message, Subscriber};
...
+}
+
+impl TryFrom<jetstream::Message> for MessageMeta {
+    type Error = anyhow::Error;
+
+    fn try_from(value: jetstream::Message) -> Result<Self, Self::Error> {
+        let info = value.info().to_anyhow()?;
+
+        Ok(MessageMeta {
+            timestamp: info.published,
```
pretentious7, line 40: whose time is this exactly? is it the time of the nats node at which the message is seen?

core/src/views/mod.rs
pretentious7: my worry here is that lookups are no longer O(1), which shouldn't be an issue, but is another thing to think about.

api doesn't really change though if we find we need to go to a hashmap.

core/src/nats.rs
pretentious7: what happens if the server republishes an old message?

core/src/views/mod.rs
paulgb: Yeah, I might give this some thought and revisit this. I'm fine with this for now though.

drone/src/agent/mod.rs
```diff
@@ -173,7 +198,7 @@ pub async fn run_agent(agent_opts: AgentOptions) -> NeverResult {
 
     let executor = Executor::new(docker, db.clone(), nats.clone(), ip, cluster.clone());
 
-    let (send_ready, recv_ready) = watch::channel(true);
+    let (send_ready, recv_ready) = watch::channel(DroneState::Ready);
```
rolyatmax, line 201: Should these names become `send_state` / `recv_state`?

drone/src/agent/mod.rs
```diff
@@ -106,13 +112,15 @@ async fn ready_loop(
     nc: TypedNats,
     drone_id: &DroneId,
     cluster: ClusterName,
-    recv_ready: Receiver<bool>,
+    recv_ready: Receiver<DroneState>,
```
rolyatmax, line 115: `recv_state`?

drone/src/agent/mod.rs
```diff
@@ -145,8 +164,14 @@ async fn listen_for_drain(
         tracing::info!(req=?req.message(), "Received request to drain drone.");
         req.respond(&()).await?;
 
-        send_ready
-            .send(!req.value.drain)
+        let state = if req.value.drain {
+            DroneState::Draining
+        } else {
+            DroneState::Ready
```
rolyatmax, line 170: I wonder if it could become an issue that sending a `drain: false` message sets a drone state to `Ready`? I suppose we could make sure we only set `DroneState::Ready` if `req.value.drain == false` _and_ the current state is `Draining`? Likewise, we'd only set `DroneState::Draining` if `req.value.drain == true` _and_ the current state is `Ready`. Is that too defensive? Thoughts?

core/src/views/replica.rs
```diff
@@ -0,0 +1,67 @@
...
+        let mut sub = nats.subscribe_jetstream::<StateUpdate>().await?;
+
+        if sub.has_pending {
+            while let Some((update, meta)) = sub.next().await {
+                view.update_state(update, meta.timestamp);
+
+                if !sub.has_pending {
+                    break;
+                }
+            }
```
rolyatmax, line 27: Does this read through all messages ever published to this stream to build the state?

core/src/nats.rs
paulgb: (Admittedly it's a little weird that this metadata comes in the form of the reply subject, but that's also where the API clients get it.)

core/src/messages/state.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        /// State for scheduling purposes.
+        state: DroneState,
+
+        /// Public IP of the drone.
+        ip: IpAddr,
+        drone_version: String,
+    },
+    /// A backend state change. Sent immediately when a backend state changes.
+    BackendStatus {
```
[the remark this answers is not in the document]
paulgb: Good catch.

core/src/views/replica.rs
paulgb: It reads through the most recent message published to each subject. It could also be constructed from a filtered stream (e.g. if we know in advance that we only want info about a specific drone), but I haven't implemented that yet.

drone/src/agent/mod.rs
paulgb: Good catch. I'm ok with this for now because setting drain to false is something we do manually, but I'll give it some thought.

rolyatmax (approved):

core/src/views/replica.rs
rolyatmax: There's a subject for each backend, right? Does that mean this grows unbounded?

Issue #145, opened 2022-09-04
Standardize NATS subjects


paulgb:
We should standardize subjects as follows:

`cluster.<cluster_name>.*`: cluster-level messages
`cluster.<cluster_name>.drone.<drone_id>.*`: drone-level messages
`cluster.<cluster_name>.drone.<drone_id>.backend.<backend_id>.*` backend-level messages

This allows us to permission pub/sub to various levels: cluster-wide, drone-wide, backend-only. It also means that all backend-level messages will need a cluster and drone field, and all drone-level messages will need a cluster field (because of our rule that the subject must be derived from the message itself).

paulgb:
Proposed changes:

|  | Current | New | Usage |
| --- | --- | --- | --- |
| **agent.rs** |  |  |  |
| `DroneLogMessage` | backend.{backend_id}.log | cluster.{cluster_id}.backend.{backend_id}.log | Report a log message that comes from a backend. backend_log stream. |
| `BackendStatsMessage` | backend.{backend_id}.stats | cluster.{cluster_id}.backend.{backend_id}.stats | Report current backend usage. Used for metering. |
| `DroneStatusMessage` | drone.{drone_id}.status | cluster.{cluster_id}.drone.{drone_id}.status | Used for reporting drone availability to scheduler. Also appears on dashboard. |
| `DroneConnectRequest` | drone.register | cluster.{cluster_id}.announce_drone | Rename to DroneAnnouncement. Currently includes IP of drone which is used for routing, but once Spawner DNS is implemented this will just be used for logging. Maybe this becomes a regular log message? |
| `SpawnRequest` | drone.{drone_id}.spawn | cluster.{cluster_id}.drone.{drone_id}.spawn | Tells a drone to run a backend which has already been scheduled to that drone. |
| `TerminationRequest` | backend.{backend_id}.terminate | cluster.{cluster_id}.backend.{backend_id}.terminate | Tells a drone to terminate a backend. |
| `BackendStateMessage` | backend.{backend_id}.status | cluster.{cluster_id}.backend.{backend_id}.status | Emitted when a backend changes state. Rename BackendStatusMessage? |
| **cert.rs** |  |  |  |
| `SetAcmeDnsRecord` | acme.set_dns_record | cluster.{cluster_id}.dns.txt | Rename to SetDnsRecord, support A (and eventually AAAA) records? |
| **logging.rs** |  |  |  |
[...]

PR #275, opened 2022-11-29
Add cluster to messages (prereq of #145 and #274)

paulgb:
This will enable #145, but doesn't actually implement the schema change, because that will be a breaking change.

This is the first stage of a two-stage deploy. Once this version of Plane is rolled out to all drones and controllers, we can make the optional `cluster` fields non-optional.


rolyatmax (approved):
Should we bump the version too?

PR #286, opened 2022-12-13
Initial GPU support

paulgb:
I still need to test this but I think it will work.


pretentious7:
I think it'd be good to note somewhere that this relies on undocumented docker features, but LGTM!!

pretentious7 (approved):

PR #290, opened 2022-12-23
Add passthrough flag for proxy

### LINKED ISSUE #289: Drone Proxy: Allow Custom Routes
We're using plane in the standalone configuration (single drone, without dnsmasq / firefox).
We use a wilcard dns to map *.example.com => 1.2.3.4 

In order to manage plane, we have a custom webservice. Currently we need a separate host or port for that, but we would like to host that on the same host and port as the drone. 
For Example: 

- When we visit example.com we reach our custom webservice (e.g. listening on 127.0.0.1:4567)
- When we visit e485cf9a-fc65-4d7c-935e-656572068226.example.com => we reach the launched backend

It would be nice if we could provide custom subdomain to route mapping in the drone.toml as a map.
Example:
```
[proxy]
# IP to listen for connections on.
bind_ip = "0.0.0.0"

[proxy.custom_route]
host = "status.example.com"
route = "http://127.0.0.1:9874"

[proxy.custom_route]
host = "example.com"
route = "http://127.0.0.1:4567"
```

If I'm correct, I could extend the code here, to make that work:
https://github.com/drifting-in-space/plane/blob/dbd550e9b1db33e611ab00730182a28eaf12e7db/drone/src/proxy/service.rs#L203-L236

paulgb:
This adds a flag to the proxy to allow a "passthrough" `SocketAddr` used when a backend can't be resolved. _All_ unknown hosts will be proxied to this backend, but it will all be TLS terminated with the same certificate, so if this is used to resolve other hostnames a custom certificate will need to be passed in which covers all of them.

Only HTTP is supported for the upstream connection.

This will close #289.

### DIFF
--- a/dev/tests/proxy.rs
+++ b/dev/tests/proxy.rs
@@ -62,7 +62,7 @@ impl Proxy {
         }
     }
 
-    pub async fn new() -> Result<Proxy> {
+    pub async fn new(passthrough: Option<SocketAddr>) -> Result<Proxy> {
         let certs = SelfSignedCert::new("proxy", vec!["*.plane.test".into(), "plane.test".into()])?;
         let bind_ip = random_loopback_ip();
         let db = DroneDatabase::new(&scratch_dir("proxy").join("drone.db")).await?;
@@ -73,6 +73,7 @@ impl Proxy {
             bind_port: 4040,
             key_pair: Some(certs.path_pair.clone()),
             cluster_domain: CLUSTER.into(),
+            passthrough,
         };
         let guard = expect_to_stay_alive(plane_drone::proxy::serve(options));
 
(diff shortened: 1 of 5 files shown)


rolyatmax:
From [this comment](https://github.com/drifting-in-space/plane/issues/289#issuecomment-1361309640)
> One possibility would be to allow a "fallback" address that anything that is not a current backend is proxied to. In a multi-drone context this would provide a way to display a custom page when an invalid domain is accessed.

If I'm understanding correctly, user visits `status.example.com`, Plane DNS sends them to a drone (?), the drone's proxy doesn't find a `status` backend and so proxies to the `passthrough` upstream? How does Plane DNS know to send `status.example.com` to a drone?

paulgb:
The use case where this came up is a situation with just one Drone, where Plane DNS isn’t used at all. Instead, status.blah.blah is caught by a wildcard DNS record and sent to the IP that Plane is running on.

Another use case is custom error pages when DNS is cached (so we don’t have a chance to serve NXDOMAIN) but a backend has been swept. Currently we serve a plain 404 in this case, but this could be used to serve a custom error page.

pretentious7 (approved):

rolyatmax (approved):

PR #292, opened 2022-12-27
Allow specifying docker user in drone config

paulgb:
This allows the docker user (equivalent to `docker --user`) to be specified when starting a drone. This allows us to ensure that containers are not run as `root`/uid 0.


pretentious7 (approved):
does this always work? like is it just the equivalent of a USER command in the dockerfile somewhere? if so couldn't like mkdir's and chattr's and stuff fail?

paulgb:
Good point @pretentious7, now that you bring it up the use cases of this in its proposed state are more limited than I had imagined. I'm going to close this for now.

Issue #294, opened 2023-01-05
Provide a way to pass a volume mount in `DockerExecutableConfig`


paulgb:
Adam would like a way to pass through a volume mount to Docker.

This would be guarded by a flag on the drone, like `--allow-volumes`, until we formalize a way to provide capabilities to executors.

Example config:

```
p["HostConfig"]["Mounts"] = [
	{
		"Type":     "volume",
		"Source":   "myvolume",
		"Target":   "/data",
		"ReadOnly": false,
		"VolumeOptions": {
			"Labels": {},
			"DriverConfig": {
				"Name": "mydriver",
				"Options": {
					"arbitrary": "string key/values"
				},
			},
		},
	}
]
```

ajbouh:
A link to the docker spec: https://pkg.go.dev/github.com/docker/docker@v20.10.22+incompatible/api/types/mount#Mount

paulgb:
Closed by #295

PR #295, opened 2023-01-05
Provide a way to pass a volume mount (#294)

paulgb:
For maximum flexibility, I chose to simply pass the json value through. The advantage is that any possible volume configuration can be passed through; the disadvantage is that we don't check the schema until it arrives at the drone, so it's a bit tricker to debug. However, as a first pass, I think this is a step in the right direction.


paulgb:
This fixes #294

pretentious7 (approved):

pretentious7:
If a nonexistent mount is specified, it errors out with failedToStart right?

pretentious7:
Ye lgtm!

PR #299, opened 2023-01-14
Use Depot to build images in CI, add ARM images

jacobwgillespie:
Hey! 👋 

This PR switches the GitHub Actions workflows to use Depot to build images in CI, for both Intel and Arm containers. I've added the ID of your `Plane` Depot project in `depot.json`, but let me know if you'd prefer a different ID there instead.

You'll need to add the repository to your project's trust relationships on the project settings page. That will use the GitHub Actions OIDC endpoint to authenticate with Depot, without needing to configure any static access token. It will look something like this (you'll want to pick `All workflows`):

image

One other consideration: Depot is not able to be used today inside PRs from third-party open-source forks. GitHub limits all access to secrets and the OIDC endpoint (`id-token: write`) in those `pull_request` workflows, so there's not a way to authenticate with Depot in those workflows. 

Our `depot/build-push-action` supports a `buildx-fallback` option, where if it's unable to acquire a Depot builder, will run the build with local Docker instead, so I've added that here for that use-case. We have two options for how we do it:

1. As implemented currently, OSS PRs will build for both Intel and Arm. When falling back to Docker, they will build the Arm portion using QEMU emulation, which will be slow. The benefit to this approach is OSS PRs are tested against Arm as well.
2. We could make Arm conditional, so that it's only built when using Depot. The downside is that an Arm build isn't attempted, but this skips the slow emulated build.

This PR is implementing option (1), but let me know if you'd prefer the other.


.github/workflows/build-cli-image.yml
```diff
@@ -27,6 +27,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v3
 
+      - name: Set up Depot
+        uses: depot/setup-action@v1
```
jacobwgillespie, line 31: This action installs the Depot CLI

.github/workflows/build-cli-image.yml
```diff
@@ -53,11 +56,12 @@ jobs:
 
       - name: Build and push Docker image
         id: build-and-push
-        uses: docker/build-push-action@ac9327eae2b366085ac7f6a2d02df8aa8ead720a
+        uses: depot/build-push-action@v1
```
jacobwgillespie, line 59: Our `depot/build-push-action` implements the same inputs as Docker's

.github/workflows/build-cli-image.yml
```diff
@@ -53,11 +56,12 @@ jobs:
...
         with:
           context: "."
           file: "cli/Dockerfile"
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
-          platforms: linux/amd64
+          platforms: linux/amd64,linux/arm64
+          buildx-fallback: true
```
jacobwgillespie, line 67: This option means that if `depot build` fails, the action will re-try the build with local `docker buildx build`

depot.json
```diff
@@ -0,0 +1 @@
+{ "id": "58j0517pw2" }
```
jacobwgillespie, line 1: Having this file at the root means that running `depot build` anywhere inside this repo will default to that project ID - it's possible to provide this to the action directly if you'd prefer to skip the file.

paulgb: Good to know, I'll leave it in for now but I occasionally prune the root directory of stuff that isn't strictly necessary so that the `README.md` doesn't fall below the fold on GitHub.

paulgb (approved):
Thanks for this, looking forward to faster builds and arm64 images!

jacobwgillespie:
^ So the emulated Arm build is gonna be rather slow (it's been running now for an hour and 45 minutes). Unless you feel strongly that you need the emulated Arm build for external PRs, I'm also happy to change it so that it skips Arm for external PRs. PRs you open from branches within this repo would still build Arm, etc.

paulgb:
I see, yes, I think it would be good to make the arm build conditional on not being emulated.

paulgb:
Cool, thanks for this!

PR #302, opened 2023-02-02
Support authentication

paulgb:
Context: #118.

This PR introduces this new functionality:
- The scheduler now respects `require_bearer_token: true` in a `ScheduleRequest`. It generates a random token, includes it in the `SpawnRequest`, and returns it in the `ScheduleResponse`.
- The proxy requests the bearer token on a `SpawnRequest` if present. It requires that requests contain the token either as an `Authorization` header or as a cookie called `_plane_auth`.
- The proxy supports a special URL called `_plane_auth`, which takes a `token` as a `GET` parameter. When a request to this endpoint is received, a `Set-Cookie` header is returned to set the `_plane_auth` cookie, and then redirect the browser to `/`.

The expected usage is:
- When you are accessing a plane backend from a server, or from the browser via `fetch` calls that you control, use the `Authorization` header to pass the bearer token.
- For cases where you want the browser to load resources directly from the backend, and can’t control the request headers, ensure that the initial request each user makes is to `[backend]/_plane_auth?token=[token]`. Subsequent requests will just work, because the cookie should be present. Note that it is not currently possible to redirect the user somewhere other than `/` after the initial `_plane_auth` request.

Other changes made by this PR:
- Added a `bearer_token` to the Drone's sqlite `route` table. This is where the proxy finds a bearer token, if one exists.
- Added some tools to the devcontainer.
- Rewrote the `generate-sqlx-data.mjs` script as a regular ol' shell script


paulgb:
After discussing with Adam on Discord, I will add a parameter to pass a URL to redirect to after `_plane_auth` is hit.

ajbouh:
Ideally that parameter is on the _plane_auth endpoint itself (and not part of spawn inputs)

dev/tests/proxy.rs
```diff
@@ -395,19 +409,145 @@ async fn update_certificates() {
...
+        .await
+        .unwrap();
+
+    proxy
+        .db
+        .insert_proxy_route(&sr.backend_id, "foobar", &server.address.to_string(), Some("foobar"))
+        .await
+        .unwrap();
+
+    let result = proxy.http_get("foobar", "/", Some("foobar"), None).await.unwrap();
```
rolyatmax, line 475: Might be good to test the token-in-query-param use case, too.

rolyatmax:
We should drop this too, now: https://github.com/drifting-in-space/plane/pull/302/files#diff-5fbc07702291f14e035feab2dac2ea9f8e4a8860fcb6890059d8bc090a95945eL202-L206

drone/src/database.rs
```diff
@@ -132,27 +138,28 @@ impl DroneDatabase {
...
         subdomain: &str,
         address: &str,
+        bearer_token: Option<&str>,
     ) -> Result<()> {
         let backend_id = backend.id().to_string();
         sqlx::query!(
             r"
             insert or replace into route
-            (backend, subdomain, address, last_active)
+            (backend, subdomain, address, last_active, bearer_token)
```
pretentious7, line 155: should we hash before storing in the db?

controller/src/lib.rs
```diff
@@ -38,8 +38,8 @@ pub async fn run_scheduler(nats: TypedNats) -> NeverResult {
                 }
             },
 
-            spawn_request = spawn_request_sub.next() => {
-                match spawn_request {
+            schedule_request = schedule_request_sub.next() => {
+                match schedule_request {
                     Some(schedule_request) => {
                         tracing::info!(spawn_request=?schedule_request.value, "Got spawn request");
```
pretentious7, line 44: we might not want to log these now that they have the tokens in them, or filter out the token before logging?

pretentious7: my b got confused.

dev/tests/proxy.rs
paulgb: The `simple_cookie_set` test tests this endpoint, unless I'm misunderstanding?

drone/src/database.rs
paulgb: Yeah, no harm in hashing these, good call.

controller/src/lib.rs
```diff
@@ -57,7 +57,7 @@ pub async fn run_scheduler(nats: TypedNats) -> NeverResult {
                                         ScheduleResponse::Scheduled {
                                             drone: drone_id,
                                             backend_id: spawn_request.backend_id,
-                                            bearer_token: None,
+                                            bearer_token: spawn_request.bearer_token,
```
pretentious7, line 60: Maybe we should have a newtype for the bearer_token? then we could write a Display trait for it  that does obfuscation.

dev/tests/proxy.rs
rolyatmax: Oh right - somehow I missed that! Nevermind me!

controller/src/lib.rs
paulgb: It wouldn't help with obfuscation in our case because anything that gets logged is through the `Serialize` code path, not the `Display` one. We want tokens to be serialized, so we can't do obfuscation on that code path.

drone/src/proxy/service.rs
```diff
@@ -271,15 +272,17 @@ impl ProxyService {
...
-                                .header("Set-Cookie", format!("{}={}", PLANE_AUTH_COOKIE, token))
-                                .body(Body::empty())?);
-                        }
-                    }
+                    let params: PlaneAuthParams =
+                        serde_html_form::from_str(req.uri().query().unwrap_or_default())?;
+
+                    return Ok(Response::builder()
+                        .status(StatusCode::FOUND)
+                        .header("Location", params.redirect.as_deref().unwrap_or("/"))
```
rolyatmax, line 280: It might make sense to only redirect to the provided URL if it's on the same domain. Endpoints that blindly redirect based on a query param can be used by bad actors to trick users into visiting sites they didn't want to visit.

paulgb: Good point. Locations can be relative, so maybe rather than match the domain we just ensure that it is a relative path?

ajbouh: I think a relative path is simplest and easiest to reason about

paulgb: Done.

rolyatmax (approved):
LGTM!

PR #305, opened 2023-02-17
Begin deprecating jetstream in drone apis

paulgb:
Rather than giving drones direct access to NATS Jetstream, we are moving to a model where drones send updates to the controller as a request and the controller writes the change to Jetstream. The controller acks the drone's request. If the controller does not ack the request in time, the drone will retry until it does.

The goals here are:
- Decouple the drone<>controller interface from the way that data is accessed by the controller.
- Simplify the drone<>controller interface so that it's easier to reason about from an auth perspective.

The goal of this PR is to gets to a drone<>controller interface that works. There is more work I'd like to do on both the drone and controller side to make this more robust.

This PR makes the following changes:
- The state machine approach is completely ripped out. The controller never used this (#285 was cancelled), so this doesn't affect ongoing operation. It does break a few Plane CLI commands until we re-implement them, but I see getting the refactor moving forward as more important right now and those CLI commands would have needed a rewrite anyway.
- The drone is no longer aware of `BackendStateMessage`. Instead, a new `UpdateBackendStateMessage` is introduced which the drone sends to the controller, causing the controller to write the equivalent `BackendStateMessage`. This is acked, mirroring the semantics of JetStream.
  - For now, drones can still send `BackendStateMessage` directly, so existing drones will still work when a new version of Plane is deployed on the controller. However, new drones **will not** be compatible with an old controller, so we will need to update the controller before rotating drones.
- `DroneStatusMessage` is sent over core NATS instead of Jetstream. We were only using Jetstream here to timestamp it.
- `SetDnsRecord` is sent over core NATS instead of Jetstream. This happens in a loop anyway, so if a message fails we still receive the next one.


controller/src/run.rs
```diff
@@ -14,6 +16,35 @@ use std::future::Future;
...
+
+        let value = BackendStateMessage {
+            backend: value.backend,
+            state: value.state,
+            cluster: Some(value.cluster),
+            time: value.time,
+        };
+
+        if let Err(e) = nc.publish_jetstream(&value).await {
+            tracing::error!(error=%e, "Failed to publish backend state message.");
```
rolyatmax, line 40: Do we want to retry if publishing fails?

drone/src/agent/executor.rs
```diff
@@ -89,6 +86,28 @@ impl<E: Engine> Clone for Executor<E> {
...
+        cluster,
+        backend,
+        time: Utc::now(),
+        drone,
+    };
+    let nc = nc.clone();
+
+    while let Err(error) = nc.request(&message).await {
+        tracing::error!(?error, "Failed to update backend state, retrying.");
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
rolyatmax, line 107: If the controller doesn't `msg.respond(&())`, does NATS make `nc.request()` return an `Err`? I'm trying to understand how the controller's ack figures into this logic - and how often the drone resends the message if no ack is received.

controller/src/run.rs
```diff
@@ -14,6 +16,35 @@ use std::future::Future;
...
+/// BackendStateMessages over JetStream.
+pub async fn update_backend_state_loop(nc: TypedNats) -> NeverResult {
+    let mut sub = nc
+        .subscribe(UpdateBackendStateMessage::subscribe_subject())
+        .await?;
+    loop {
+        let msg = sub
+            .next()
+            .await
+            .ok_or_else(|| anyhow!("UpdateBackendStateMessage subscription ended."))?;
```
pretentious7, line 29: what exactly happens if a subscription ends like this? There's a nats loop somewhere else that restarts all these loops on reconnect right?

drone/src/agent/executor.rs
paulgb: Yes, if the controller doesn't hear back within NATS' [request timeout](https://docs.rs/async-nats/latest/async_nats/struct.ConnectOptions.html#method.request_timeout) (default 10s), it will return a timeout error.

But more likely than timing out, it will error right away if it can't send the message, and then leave a 1s gap.

controller/src/run.rs
paulgb: Currently, no. These subscriptions shouldn't end based on my understanding of NATS, and I want to know if it happens.

controller/src/run.rs
paulgb: I've added a `continue` so that if there is a jetstream error, we do not ack, and the drone's retry loop kicks in.

PR #306, opened 2023-02-17
add memory limit opt

pretentious7:
also ran cargo fmt

### DIFF
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -376,7 +376,8 @@ pub struct ResourceLimits {
     #[serde_as(as = "Option<DurationSeconds>")]
     pub cpu_time_limit: Option<Duration>,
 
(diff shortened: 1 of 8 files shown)


core/src/messages/agent.rs
```diff
@@ -375,6 +375,8 @@ pub struct ResourceLimits {
     /// Total cpu time allocated to container    
     #[serde_as(as = "Option<DurationSeconds>")]
     pub cpu_time_limit: Option<Duration>,
+
+    pub memory_limit: Option<i64>,
```
paulgb, line 379: What’s the unit here? How about something like `memory_limit_mb` or similar depending on what the unit is?

pretentious7: changed to memory_limit_bytes with comment

paulgb (approved):

PR #307, opened 2023-02-21
Add cluster to message schemas

paulgb:
This adds a `cluster` field to all messages that cross the drone<>controller boundary.

Logs are an exception. For now, drones will be allowed to publish (but not subscribe) to `logs.drone.>`.

Planning doc: https://docs.google.com/document/d/1hXeucu2e5zuMD2f0Ft3K9wpRaJ8hp0zniFUK8s6HP9w/edit

### DIFF
--- a/core/src/messages/agent.rs
+++ b/core/src/messages/agent.rs
@@ -107,6 +107,7 @@ impl JetStreamable for DroneLogMessage {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct BackendStatsMessage {
+    cluster: Option<ClusterName>,
     backend_id: BackendId,
     /// Fraction of maximum CPU.
     pub cpu_use_percent: f64,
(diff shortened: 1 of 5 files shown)


core/src/messages/agent.rs
```diff
@@ -107,6 +107,7 @@ impl JetStreamable for DroneLogMessage {
 
 #[derive(Serialize, Deserialize, Debug)]
 pub struct BackendStatsMessage {
+    cluster: Option<ClusterName>,
```
rolyatmax, line 110: We'll change this to a required field once all the drones are updated?

paulgb: Exactly. This just allows us to push a version of the controller that is able to accept messages that don't include it.

rolyatmax (approved):
Two questions, but otherwise LGTM:

1. Is the idea that we get all drones running with an optional `cluster` field (this PR), then in a follow-up, make the cluster field required and use it to generate the subject name?
2. Do we need to update the version so we can tell when all drones have been updated in the dashboard?

paulgb:
> Two questions, but otherwise LGTM:
> 
> 1. Is the idea that we get all drones running with an optional `cluster` field (this PR), then in a follow-up, make the cluster field required and use it to generate the subject name?

I think I'll actually change the subject name first, and then make it a mandatory field. Initially it will send messages to both the old and new subject.

> 2. Do we need to update the version so we can tell when all drones have been updated in the dashboard?

Good idea, just threw in a version update. I'm not sure we will want to deploy this drone version (vs. one with a few more changes), but this gives us some flexibility.

PR #308, opened 2023-02-22
Use cluster-based inbox

paulgb:
This will enable a drone to operate even if it is restricted to messaging on `cluster.{cluster}.>`

Also bumps version.

### DIFF
--- a/controller/src/plan.rs
+++ b/controller/src/plan.rs
@@ -21,7 +21,7 @@ pub struct ControllerPlan {
 
 impl ControllerPlan {
     pub async fn from_controller_config(config: ControllerConfig) -> Result<Self> {
-        let nats = config.nats.connect_with_retry().await?;
+        let nats = config.nats.connect_with_retry("controller.inbox").await?;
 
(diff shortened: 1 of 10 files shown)


controller/src/plan.rs
```diff
@@ -21,7 +21,7 @@ pub struct ControllerPlan {
 
 impl ControllerPlan {
     pub async fn from_controller_config(config: ControllerConfig) -> Result<Self> {
-        let nats = config.nats.connect_with_retry().await?;
+        let nats = config.nats.connect_with_retry("controller.inbox").await?;
```
pretentious7, line 24: so If I understand correctly, responses will be published to "controller.inbox.<UID>" ?

pretentious7 (approved):
one clarifying question, LGTM!

controller/src/plan.rs
paulgb: It will be `controller.inbox.{some random string generated by NATS}`

rolyatmax (approved):
LGTM! Nice that inbox-prefixing was already supported by NATS!

PR #311, opened 2023-02-22
Send duplicate messages to new and old NATS subjects

paulgb:
This is a step in the transition to a new NATS subject schema. Once we switch the controller to listen on the new subjects, the old subjects will be removed.

### DIFF
--- a/core/src/nats.rs
+++ b/core/src/nats.rs
@@ -13,6 +13,7 @@ use async_nats::{Client, Message, Subscriber};
 use bytes::Bytes;
 use dashmap::DashSet;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use tokio::select;
 use std::fmt::Debug;
 use std::marker::PhantomData;
 use std::sync::Arc;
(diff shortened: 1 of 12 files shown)


core/src/nats.rs
```diff
@@ -421,11 +449,22 @@ impl TypedNats {
...
+            .request(value.subject(), bytes.clone());
+        
+        let result = if let Some(tmp_alt_subject) = value.tmp_alt_subject() {
+            let result2 = self.nc
+                .request(tmp_alt_subject, bytes);
+
+            select! {
+                result = result1 => result.to_anyhow()?,
+                result = result2 => result.to_anyhow()?,
+            }
```
rolyatmax, line 464: Not sure I'm following exactly how this `select!` and the `result = ... => ...` parts are working

rolyatmax (approved):
LGTM!

core/src/nats.rs
paulgb: The syntax is specific to the `select!` macro but looks like a `match` statement. Whichever of `result1` or `result2` resolves first becomes the branch that gets executed. In both branches, the result gets bound to `result` (because we have `result =` in both branches) and returned (after calling `.to_anyhow()?` in both cases.

PR #312, opened 2023-02-22
Update base64 version

paulgb:


rolyatmax (approved):
Does this new version just change the library's API or has the functionality changed?

paulgb:
> Does this new version just change the library's API or has the functionality changed?

It just changes the API, as far as I can tell.

PR #314, opened 2023-02-23
Logging and robustness improvements

paulgb:


drone/src/cert/mod.rs
```diff
@@ -89,44 +93,61 @@ pub async fn get_certificate(
...
+        .context("Sending SetDnsRecord")?;
 
         tracing::info!("Validating challenge.");
-        let challenge = challenge.validate().await?;
-        let challenge = challenge.wait_done(Duration::from_secs(5), 3).await?;
+        let challenge = challenge.validate().await.context("Validating challenge")?;
+        let challenge = challenge
+            .wait_done(Duration::from_secs(5), 3)
+            .await
+            .context("Waiting for challenge")?;
```
pretentious7, line 117: so if this errors then it'll return the context in the error?

drone/src/cert/mod.rs
```diff
@@ -62,21 +62,25 @@ pub async fn get_certificate(
...
+    let authorizations = order
+        .authorizations()
+        .await
+        .context("Fetching authorizations")?;
     for auth in authorizations {
         tracing::info!("Requesting challenge.");
         let challenge = auth
             .get_challenge(DNS_01)
-            .ok_or_else(|| anyhow!("Couldn't obtain dns-01 challenge."))?;
+            .context("Obtaining dns-01 challenge")?;
```
pretentious7, line 79: why is the .context better than having an ok_or_else here?

drone/src/cert/mod.rs
```diff
@@ -62,21 +62,25 @@ pub async fn get_certificate(
...
 
         let value = challenge
-            .key_authorization_encoded()?
+            .key_authorization_encoded()
+            .context("Encoding authorization")?
             .ok_or_else(|| anyhow!("No authorization value."))?;
```
pretentious7, line 84: is this ok_or_else necessary?

pretentious7 (approved):
this looks good! a few doubts about the .context() stuff.

core/src/nats.rs
```diff
@@ -450,24 +450,29 @@ impl TypedNats {
...
-                .request(tmp_alt_subject, bytes);
-
-            select! {
-                result = result1 => result.to_anyhow()?,
-                result = result2 => result.to_anyhow()?,
-            }
-        } else {
-            result1.await.to_anyhow()?
+
+        let mut futs = FuturesUnordered::new();
```
rolyatmax, line 454: Pushing a future to `futs` here will immediately call `.await` on it (i.e. "fire" it off)? Or does `futs.next()` fire them all off? Or something else?

core/src/nats.rs
```diff
@@ -450,24 +450,29 @@ impl TypedNats {
...
-        let value: T::Response = serde_json::from_slice(&result.payload)?;
-        Ok(value)
+        while let Some(result) = futs.next().await {
+            match result {
+                Ok(result) => {
+                    let value: T::Response = serde_json::from_slice(&result.payload)?;
+                    return Ok(value);
+                }
+                Err(err) => {
+                    tracing::warn!(error = ?err, "One future failed while waiting for response");
```
rolyatmax, line 469: Will we see this whenever there's no listener on the subject? In other words, we'll basically always see one of these warn messages since we only expect a listener to be on either the old _or_ the new subject?

rolyatmax (approved):
A few questions there, but LGTM!

drone/src/cert/mod.rs
paulgb: Yes, exactly. It's an anyhow thing.

drone/src/cert/mod.rs
paulgb: ok_or_else replaces the original error; context is additive

core/src/nats.rs
paulgb: Awaiting futs.next() will poll both promises, and then will return the result of each in the order it resolves.

core/src/nats.rs
paulgb: Usually, yes. The exception is that if we are running a wiretap, NATS doesn't know that the wiretap won't respond, so it will wait for some timeout before assuming that there are no responders.

drone/src/cert/mod.rs
paulgb: Good catch, probably not.

PR #316, opened 2023-02-27
Consume new subjects (WIP)

paulgb:
This is WIP because I realized that `DroneStatusMessage` will need to be split into two messages like `BackendStateMessage`, so I'm doing that first.

### DIFF
--- a/core/src/messages/cert.rs
+++ b/core/src/messages/cert.rs
@@ -26,6 +26,6 @@ impl TypedMessage for SetAcmeDnsRecord {
 
 impl SetAcmeDnsRecord {
     pub fn subscribe_subject() -> SubscribeSubject<Self> {
-        SubscribeSubject::new("acme.set_dns_record".to_string())
+        SubscribeSubject::new("cluster.*.set_dns_record".to_string())
     }
 }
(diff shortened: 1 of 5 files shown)


core/src/messages/cert.rs
```diff
@@ -20,12 +20,12 @@ impl TypedMessage for SetAcmeDnsRecord {
     }
 
     fn tmp_alt_subject(&self) -> Option<String> {
-        Some(format!("cluster.{}.set_acme_record", self.cluster))
```
paulgb, line 23: this will break backwards compatibility, but we are intentionally breaking backwards compatibility soon anyway and this particular instance is low-risk

pretentious7: Will this cause https downtime when we switch over?

pretentious7 (approved):
Small concern re what happens if the acme messages dont work on new drones, but lgtm otherwise!

core/src/messages/cert.rs
paulgb: No, drones that have already obtained certs will continue to be able to use them. As long as we roll the drones before their certs expire, we'll avoid any downtime.

PR #317, opened 2023-02-27
Add identity token option

paulgb:
When using user/password credentials, Docker makes a round trip to an authentication server to obtain a JWT before pulling the image. The Docker API provides an alternative where we can provide this JWT directly. This PR exposes that option so that the spawner can avoid the round-trip if it already has an identity token.


pretentious7 (approved):

rolyatmax:
Quick work of this! Nice job!

paulgb:
> Quick work of this! Nice job!

Thanks! The p2 version is coming along, but docker auth is kind of annoying.

PR #318, opened 2023-02-28
Stop using 16 core workers

paulgb:
The github action is just kicking off a depot job anyway, so it doesn't need to be heavyweight.

I also removed it from the clippy one because we shouldn't need 16 cores just to run a linter.

### DIFF
--- a/.github/workflows/build-cli-image.yml
+++ b/.github/workflows/build-cli-image.yml
@@ -17,7 +17,6 @@ env:
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


paulgb:
> How much slower do builds run when removing this?

They should actually run faster, because build machines from the general pool are allocated quicker than build machines from the high-compute pool. We're not actually doing anything compute-intensive on the build machines any more; the actual building runs on a machine that Depot provisions.

PR #329, opened 2023-03-28
State machine refactor

paulgb:
This refactors the way that state is retained by the Plane controller. Instead of maintaining state across a number of data structures and persisting into several NATS JetStreams, this change moves all state into a `WorldState` object.

Each controller instance maintains only one instance of a `WorldState`, which is accessed via a `StateHandle` object which contains a shared reference to that global instance.

Here's a simplified diagram:

image

- The implementation of `WorldState` is in `controller/src/state/world_state.rs`
- The "state loop" is started by `start_state_loop` in `controller/src/state/mod.rs`.
  - This will be refactored by #334
- `DroneStateUpdate` is implemented in `core/src/messages/drone_state.rs`. It is effectively the union of the existing `DroneStatusMessage`, `UpdateBackendStateMessage`, `SetAcmeDnsRecord`, and `DroneConnectRequest`.
  - It uses `#[serde(untagged)]` to detect which of those messages a message from the drone is.
  - To avoid changing the API, this has four different subscribe subjects, one for each message type.
- In `monitor_drone_state` (`controller/src/drone_state.rs`), we listen on all four subscribe subjects but handle messages from each in the same way, by passing them to `convert_to_state_message`. This converts the message into the type(s) needed to store it in the world state machine.
  - The resulting messages are then sent to the state machine, and if a reply inbox is provided, it receives a reply. Once the drone receives this reply, it can assume that the state change has been persisted.

TODO:

- [x] Describe the changes in more detail
- [x] Write a test for A-record creation
  - [x] Something needs to publish `BackendMessageType::Assignment`, probably the scheduler?
- [x] Store the drone version somewhere
- [x] Make it pick up existing live drones

Saving for follow-up PRs:

- Simplify drone<>controller communication by using `UpdateBackendStateMessage` in the drone.
- Detect and clean up “lost” drones.
- Remove unused fields from messages.
- Keep only one TXT record per drone (requires adding drone ID to TXT messages)
- Clean up drones after they are lost.


core/src/messages/agent.rs
```diff
@@ -537,54 +556,6 @@ impl BackendState {
...
-            "cluster.{}.backend.{}.status",
-            cluster.subject_name(),
-            backend.id()
-        ))
-    }
-}
-
 /// An message representing a change in the state of a backend.
 /// **DEPRECATED** for drone-side use by UpdateBackendStateMessage.
 /// Drones will send an UpdateBackendStateMessage and the controller
```
rolyatmax, line 561: Should we update this comment now that `UpdateBackendStateMessage` is going away?

controller/src/drone_state.rs
```diff
@@ -0,0 +1,297 @@
...
+
+    let mut drone_status_sub = nats
+        .subscribe(DroneStateUpdate::subscribe_subject_drone_status())
+        .await?;
+    tracing::info!("Subscribed to drone status messages.");
+
+    let mut backend_status_sub = nats
+        .subscribe(DroneStateUpdate::subscribe_subject_backend_status())
+        .await?;
+    tracing::info!("Subscribed to backend status messages.");
```
rolyatmax, line 117: Proposal: maybe we use "status" and "state" to mean more specific things. E.g. we could use "status" to describe heartbeat-like messages and something like "state change" for messages reporting when a backend's state has changed. Thoughts?

controller/src/drone_state.rs
```diff
@@ -0,0 +1,297 @@
...
+    loop {
+        let message = select! {
+            acme_msg = acme_sub.next() => acme_msg,
+            connect_msg = connect_sub.next() => connect_msg,
+            drone_status_msg = drone_status_sub.next() => drone_status_msg,
+            backend_status_msg = backend_status_sub.next() => backend_status_msg,
+        };
+
+        if let Some(message) = message {
+            tracing::info!(message=?message.value, "Got state message from drone.");
```
rolyatmax, line 128: This will end up printing out a lot of drone status messages (we had them set to debug level before). Maybe this is fine though? Or we could rework this to have drone status messages be debug level while the others are info level? I'm fine with leaving as is, too

controller/src/drone_state.rs
```diff
@@ -0,0 +1,297 @@
...
+            cluster: msg.cluster.clone(),
+            message: ClusterStateMessage::DroneMessage(DroneMessage {
+                drone: msg.drone_id.clone(),
+                message: DroneMessageType::Metadata(DroneMeta {
+                    ip: msg.ip,
+                    git_hash: msg.git_hash.clone(),
+                    version: msg
+                        .version
+                        .clone()
+                        .unwrap_or("missing_drone_id".to_string()), // Existing drones do not report a version in their connect message.
```
rolyatmax, line 38: Should be "missing_drone_version", no?

controller/src/lib.rs
```diff
@@ -1,86 +1,85 @@
...
-                    None => return Err(anyhow!("spawn_request_sub.next() returned None.")),
+                        ScheduleResponse::Scheduled {
+                            drone: drone_id,
+                            backend_id: spawn_request.backend_id,
+                            bearer_token: spawn_request.bearer_token.clone(),
+                        }
+                    }
+                    Ok(false) => {
+                        tracing::warn!("No drone available.");
+                        ScheduleResponse::NoDroneAvailable
```
rolyatmax, line 63: Should this be something more like "Drone did not accept backend spawn"?

controller/src/lib.rs
```diff
@@ -1,86 +1,85 @@
...
+                    }
+                    Err(error) => {
+                        tracing::warn!(?error, "Scheduler returned error.");
+                        ScheduleResponse::NoDroneAvailable
```
rolyatmax, line 67: Should we have a `ScheduleResponse::InternalError` or some other error type here? Or does this error really mean that no drone is available?

controller/src/state/world_state.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn new(world_state: WorldState) -> Self {
+        Self(Arc::new(RwLock::new(world_state)))
+    }
+
+    pub fn get_ready_drones(
+        &self,
+        cluster: &ClusterName,
+        timestamp: DateTime<Utc>,
+    ) -> Result<Vec<DroneId>> {
+        let world_state = self.0.write().expect("Could not acquire world_state lock.");
```
rolyatmax, line 31: Can this just be a read() lock?

controller/src/state/world_state.rs
```diff
@@ -0,0 +1,172 @@
...
+            ClusterStateMessage::DroneMessage(message) => {
+                let drone = self.drones.entry(message.drone).or_default();
+                drone.apply(message.message);
+            }
+            ClusterStateMessage::BackendMessage(message) => {
+                let backend = self.backends.entry(message.backend).or_default();
+                backend.apply(message.message);
+            }
+            ClusterStateMessage::AcmeMessage(message) => {
+                if self.txt_records.len() > 10 {
```
rolyatmax, line 95: Should this be a `>=`? Or do we intend the txt_records to scale up to 11? (Probably doesn't really matter?)

rolyatmax (approved):
A few comments there, but otherwise LGTM!

controller/src/lib.rs
paulgb: Currently `NoDroneAvailable` is a catch-all for instances that don't spawn. I'm trying to minimize the surface area of API changes in this PR, so I've opened #335 to track this as a follow-up.

controller/src/state/world_state.rs
paulgb: yes, good catch!

PR #331, opened 2023-04-01
Log backend exit code

paulgb:
We should expose this through the API as well but for now this is useful for debugging.

### DIFF
--- a/drone/src/agent/engines/docker/mod.rs
+++ b/drone/src/agent/engines/docker/mod.rs
@@ -392,7 +392,8 @@ impl Engine for DockerInterface {
                 Some(_) => EngineBackendStatus::Failed,
             };
 
-            let state_json = serde_json::to_string(&state).ok();
+            let state_json = serde_json::to_string(&state)
+                .unwrap_or_else(|_| "Failed to serialize container state.".into());
 
             tracing::info!(?status, state=?state_json, "Set to terminal state.");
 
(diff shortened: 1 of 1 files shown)


drone/src/agent/engines/docker/mod.rs
```diff
@@ -386,11 +386,15 @@ impl Engine for DockerInterface {
...
-                Some(0) => Ok(EngineBackendStatus::Exited),
-                Some(_) => Ok(EngineBackendStatus::Failed),
-            }
+            let status = match state.exit_code {
+                None => EngineBackendStatus::Terminated,
+                Some(0) => EngineBackendStatus::Exited,
+                Some(_) => EngineBackendStatus::Failed,
+            };
+
+            tracing::info!(?status, ?state.exit_code, "Set to terminal state.");
```
rolyatmax, line 395: I think we should log the entire `state`. There's quite a bit of useful info in there:
```
    pub status: Option<ContainerStateStatusEnum>,
    pub running: Option<bool>,
    pub paused: Option<bool>,
    pub restarting: Option<bool>,
    pub oom_killed: Option<bool>,
    pub dead: Option<bool>,
    pub pid: Option<i64>,
    pub exit_code: Option<i64>,
    pub error: Option<String>,
    pub started_at: Option<String>,
    pub finished_at: Option<String>,
    pub health: Option<Health>,
```

rolyatmax (approved):

PR #338, opened 2023-04-11
Improve local development ergonomics for Plane

paulgb:
This PR makes it easier to develop Plane locally.

It:
- Removes some of the more especially chatty logging. This is recently-introduced logging that is not load-bearing.
- Adds `./dev-*.sh` scripts to simplify common tasks.
- Adds Plane config files for running locally, including a `curl` wrapper.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /target
 test-scratch
 lcov.info
+drone.sqlite3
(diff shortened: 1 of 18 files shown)


pretentious7 (approved):
Would be good to run the scripts through shellcheck, but not blocker.

Lgtm!

rolyatmax (approved):

PR #340, opened 2023-04-12
get list-drones to work

paulgb:
Also:
- fixes `nats` -> `nats-server` in `dev-nats.sh`
- switches some hashmaps to btreemaps so that order is preserved in CLI.


pretentious7 (approved):
cool!

PR #347, opened 2023-04-17
Abhi/return errors from backend start

pretentious7:
Adds a docker DroneLogMessageKind, allows for injecting logs using a tokio mpsc channel. 

also one test


dev/Cargo.toml
```diff
@@ -22,7 +22,7 @@ rcgen = "0.10.0"
 reqwest = { version = "0.11.11", features=["native-tls", "cookies"] }
 rustls-pemfile = "1.0.1"
 serde_json = "1.0.83"
-tokio = { version = "1.20.1", features = ["macros"] }
+tokio = { version = "1.20.1", features = ["full"] }
```
paulgb, line 25: Do you know what additional tokio features we're using? It would probably be better for compile times to list them explicitly.

dev/src/container.rs
```diff
@@ -206,3 +210,32 @@ pub async fn pull_image(docker: &Docker, name: &str) -> Result<()> {
 
     Ok(())
 }
+
+fn random_lowercase_string(len: i32) -> String {
+    let mut rng = rand::thread_rng();
+    let alphabet = Uniform::from('a'..='z');
+    (0..len).map(|_| rng.sample(alphabet) as char).collect()
```
paulgb, line 217: There is already a `random_string` function in `util`, can that be used here? If it needs to be lowercase, you could change the other function to be lowercase without breaking anything.

dev/src/container.rs
```diff
@@ -206,3 +210,32 @@ pub async fn pull_image(docker: &Docker, name: &str) -> Result<()> {
...
+}
+
+pub async fn build_image<P: AsRef<Path>>(path: P) -> Result<String> {
+    // the docker api requires tarballs, which is just an unnecessary headache, so just using the binary here.
```
paulgb, line 221: yeah, that's a whole rabbit hole with implementing `.dockerignore` handling etc., calling the binary is 100% better.

dev/src/util.rs
```diff
@@ -15,6 +15,8 @@ use std::{
 };
 use tokio::net::TcpSocket;
```
paulgb, line 17: delete this line and rerun `cargo fmt`

dev/test-images/invalid-image/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM busybox:latest
+ENTRYPOINT [ "does-not-exist" ]
```
paulgb, line 2: add newline

dev/tests/agent.rs
```diff
@@ -253,6 +256,66 @@ async fn drone_sends_draining_status() {
         .unwrap();
 }
 
+#[allow(dead_code)]
+struct Ctx {
```
paulgb, line 260: Maybe rename to `TestContext` or `TestFixture`?

dev/tests/agent.rs
```diff
@@ -253,6 +256,66 @@ async fn drone_sends_draining_status() {
...
+    nats_server: Nats,
+    nats_connection: TypedNats,
+    controller: MockController,
+    agent: Agent,
+}
+
+async fn do_spawn_request(mut request: SpawnRequest) -> (Ctx, BackendStateSubscription) {
```
paulgb, line 267: Since `Ctx` exists to be returned by this function, how about putting the state subscription as a field on `Ctx` rather than returning a tuple?

drone/src/agent/executor.rs
```diff
@@ -220,7 +220,16 @@ impl<E: Engine> Executor<E> {
         let (send, mut recv) = channel(1);
         self.backend_to_listener
             .insert(spawn_request.backend_id.clone(), send);
-
+        self.backend_to_monitor.insert(
```
paulgb, line 223: Since `resume_backends` calls `run_backend`, we no longer need to insert this in `resume_backends` (lines 201-212 can be deleted)

drone/src/agent/executor.rs
```diff
@@ -288,6 +279,16 @@ impl<E: Engine> Executor<E> {
                     match state {
                         BackendState::Loading => {
                             state = BackendState::ErrorLoading;
+                            if let Err(_inject_err) = self
```
paulgb, line 282: we should probably throw this in the log line, e.g.

```rust
tracing::error!(err=?inject_err, "Failed to...");
```

core/src/messages/agent.rs
```diff
@@ -46,10 +46,11 @@ impl From<&DockerCredentials> for bollard::auth::DockerCredentials {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub enum DroneLogMessageKind {
     Stdout,
     Stderr,
+    Executor,
```
paulgb, line 53: Maybe `Meta` or `Event`?

paulgb (commented):
Looks good, left a few comments

drone/src/agent/backend.rs
```diff
@@ -99,19 +99,16 @@ impl BackendMonitor {
...
         let backend_id = backend_id.clone();
 
         tokio::spawn(async move {
             tracing::info!(%backend_id, "Log recording loop started.");
 
-            while let Some(v) = tokio::select! {
-                v = stream.next() => { v }
-                v = meta_log_rx.recv() => { v }
-            } {
+            while let Some(v) = stream.next().await {
```
paulgb: oh, this is much nicer!

Although, since this is an `mpsc` sender, have you considered cloning the sender? That way instead of merging streams all the messages (from both clones of the sender) could go to the same stream.

paulgb (commented):
Let's work through the tests issue today, and I've noted a few follow-ups we can do, but otherwise this lgtm, good tidy job!

dev/tests/agent.rs
pretentious7: I was thinking there'd be other things that also eventually returned TestFixture, and also their individual things in a tuple.

core/src/nats.rs
```diff
@@ -200,6 +204,20 @@ where
...
+where
+    T: TypedMessage + std::marker::Unpin,
+{
+    type Item = MessageWithResponseHandle<T>;
+
+    fn poll_next(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> Poll<Option<MessageWithResponseHandle<T>>> {
+        let mut fut1 = Box::pin(Self::next(&mut self));
```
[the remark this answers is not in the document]
pretentious7: yep, that was my plan for the future, but I only understand about 70% of what this is doing, and I didn't want to tempt fate.

drone/src/agent/backend.rs
pretentious7: uhhhh, I don't think I follow, I'm merging the receiver stream with the nats stream so that it only terminates when they're both done.

paulgb (approved):

PR #355, opened 2023-04-20
Implement Supervisor

paulgb:


core/src/supervisor.rs
```diff
@@ -0,0 +1,38 @@
...
+/// will log an error and restart the future after a short delay.
+pub struct Supervisor {
+    name: &'static str,
+    handle: JoinHandle<()>,
+}
+
+impl Supervisor {
+    pub fn new<F, Fut>(name: &'static str, factory: F) -> Self
+    where
+        //F: Fn() -> Pin<Box<dyn Future<Output = NeverResult> + Send>> + Send + 'static,
```
pretentious7, line 16: remove comment

drone/src/run.rs
```diff
@@ -66,35 +57,35 @@ async fn drone_main() -> NeverResult {
...
+        }))
+    } else {
+        None
+    };
 
-    if let Some(agent_options) = agent_options {
-        futs.push(Box::pin(run_agent(agent_options)))
-    }
+    #[allow(clippy::manual_map)]
+    let _agent_supervisor = if let Some(agent_options) = agent_options {
```
pretentious7, line 77: why not use map here?

core/src/cli.rs
```diff
@@ -10,6 +11,21 @@ struct CliArgs {
     config_file: Option<String>,
 }
```
pretentious7, line 13: hmm why is this in cli.rs?

controller/src/run.rs
```diff
@@ -62,38 +56,45 @@ async fn controller_main() -> NeverResult {
...
+        None
+    };
 
-    try_join_all(futs.into_iter()).await?;
-    // try_join_all either returns an Err, or Ok() with a list of Never values.
-    // Since Never values are not constructable, if we get here, we can assume that
-    // try_join_all returned an empty list. This means that no event loops ever
-    // actually ran, i.e. futs was empty.
-    Err(anyhow!("No event loops selected."))
+    wait_for_interrupt().await
```
pretentious7, line 94: cool! so when this goes off it'll actually do the drop for the event loops right? does this mean containers get shut down?

drone/src/run.rs
paulgb: I can't use map for the ones that require `.clone()`, so I prefer to keep the form consistent for the ones that can.

core/src/cli.rs
paulgb: Hmm, fair point. I see it as the counterpart to `init_cli` -- `init_cli` is run at the beginning of the `async main` and `wait_for_interrupt` is run at the end. But maybe `cli` isn't the best name for this module any more? Maybe `process` or something?

controller/src/run.rs
paulgb: it will drop the event loops, but it won't terminate the containers.

core/src/supervisor.rs
paulgb: good catch.

pretentious7 (approved):

PR #358, opened 2023-04-21
Plane CLI command for cleaning up swept backends

paulgb:


cli/src/main.rs
```diff
@@ -74,6 +79,33 @@ async fn main() -> Result<()> {
...
+                    if state == BackendState::Swept {
+                        println!(
+                            "Removing backend {} from cluster {}, swept at {}",
+                            backend_id, cluster_name, timestamp
+                        );
+
+                        stream
+                            .purge()
+                            .filter(format!(
+                                "state.cluster.{}.backend.{}.>",
```
pretentious7, line 99: hmmmm so this goes through the entire stream for each swept backend? wouldn't it be faster to construct a filter for all the relevant backends then purge().fliter() once? Not sure if that's possible.

paulgb: Unfortunately NATS doesn't allow this. I'm not too worried about the performance of this particular task, though; it's something that will eventually be run in a loop.

pretentious7 (approved):

PR #362, opened 2023-04-27
Abhi/build hello world

pretentious7:
builds hello world image instead of using uploaded image, also makes a couple changes to prevent creating redundant docker tags.

the new hello world uses busybox's httpd.


pretentious7:
should I delete all the material for the old hello world in this pr?

paulgb:
> should I delete all the material for the old hello world in this pr?

from the past commits? I think this is fine, they will be flattened when committed anyway

pretentious7:
nono I mean like the files themselves, I haven't removed them yet, dev/test-server.

but now that I think about it really no reason to keep them around. I'll delete

dev/src/container.rs
```diff
@@ -213,8 +213,7 @@ pub async fn build_image<P: AsRef<Path>>(path: P) -> Result<String> {
     // the docker api requires tarballs, which is just an unnecessary headache, so just using the binary here.
     let path = path.as_ref();
     let mut docker_cmd = Command::new("docker");
-    let image_name = random_string(10);
```
paulgb, line 216: You can remove the `random_string` import.

dev/src/container.rs
```diff
@@ -225,6 +224,6 @@ pub async fn build_image<P: AsRef<Path>>(path: P) -> Result<String> {
         );
         return Err(anyhow!("docker build failed in dir {}", path.display()));
     }
-
-    Ok(image_name)
+	let docker_quiet_out: String = String::from_utf8(output.stdout).unwrap();
```
paulgb: Tabs; Run `cargo fmt`

dev/test-images/buildable-test-server/resources/cgi-bin/index.cgi
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+cat <<EOF
```
paulgb, line 2: does this need to be CGI? Does httpd not serve static files?

dev/test-images/buildable-test-server/resources/server.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+export EXIT_TIMEOUT
+export EXIT_CODE
+set -x
+
+[ -n "${EXIT_TIMEOUT}" ] && { sleep "${EXIT_TIMEOUT}" && kill -USR1 "$$" ; } &
```
paulgb, line 6: Add comment `# kill the server after $EXIT_TIMEOUT if it is set`

paulgb (commented):
Left a few comments, otherwise LGTM once tests pass

dev/test-images/buildable-test-server/resources/cgi-bin/index.cgi
pretentious7: yes, but it feels the need to add a newline and I can't set the header requiring Content-Length:12 (which removes the newline)

I could also just add a \n in the test comparison...

paulgb (commented):
Thanks for the changes, this is almost there, replied to your comment.

dev/test-images/buildable-test-server/resources/cgi-bin/index.cgi
pretentious7: that sounds good to me, I've put in a hello.txt file in this PR. I'm following this up with another pr with an env.cgi that returns some environment variable.

paulgb (approved):
LGTM, thanks!

PR #363, opened 2023-04-27
Abhi/schedule-images

pretentious7:
this is a skeleton for how scheduling image resources might work.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -177,7 +177,10 @@ async fn main() -> Result<()> {
                                     continue;
                                 }
                                 format!("is alive at {}", timestamp.to_string().bright_blue())
-                            }
+                            },
+							DroneMessageType::Image(ref image) => {
+								format!("image {} on drone {}", image, drone)
+							}
                             DroneMessageType::Metadata(metadata) => {
                                 format!(
                                     "has IP: {}, version: {}, git hash: {}",
@@ -333,7 +336,7 @@ async fn main() -> Result<()> {
                 .await?;
 
             match result {
-                ScheduleResponse::Scheduled {
+                ScheduleResponse::ScheduledBackend {
                     drone,
                     backend_id,
                     bearer_token,
@@ -348,6 +351,9 @@ async fn main() -> Result<()> {
                         println!("Bearer token: {}", bearer_token.bright_blue());
                     }
                 }
+				ScheduleResponse::ScheduledImage { drone, image } => {
+					println!("image {} on drone {}", image, drone)
+				}
                 ScheduleResponse::NoDroneAvailable => tracing::error!(
                     %cluster,
                     "Could not schedule backend because no drone was available for cluster."
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -33,6 +33,10 @@ struct Opts {
 enum Command {
     ListDrones,
     ListBackends,
+	DownloadImage {
+		cluster: String,
+		url: String
+	},
     Spawn {
         cluster: String,
         image: String,
@@ -93,6 +97,9 @@ async fn main() -> Result<()> {
         .await?;
 
     match opts.command {
+		Command::DownloadImage { cluster, url } => {
+			todo!()
+		},
         Command::Cleanup {
             dry_run,
             include_missing_state,
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -7,7 +7,7 @@ use plane_core::{
             BackendState, BackendStateMessage, DockerExecutableConfig, ResourceLimits,
             TerminationRequest,
         },
-        scheduler::{DrainDrone, ScheduleRequest, ScheduleResponse},
+        scheduler::{BackendResource, DrainDrone, Resource, ScheduleRequest, ScheduleResponse},
         state::{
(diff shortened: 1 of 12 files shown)


controller/src/lib.rs
```diff
@@ -25,12 +25,22 @@ pub async fn run_scheduler(nats: TypedNats, state: StateHandle) -> NeverResult {
     let mut schedule_request_sub = nats.subscribe(ScheduleRequest::subscribe_subject()).await?;
     tracing::info!("Subscribed to spawn requests.");
 
-    while let Some(schedule_request) = schedule_request_sub.next().await {
-        tracing::info!(spawn_request=?schedule_request.value, "Got spawn request");
-        let result = match scheduler.schedule(&schedule_request.value.cluster, Utc::now()) {
+    while let Some(
+        ref msg @ MessageWithResponseHandle {
```
paulgb, line 29: This will end the scheduler loop if we get a schedule request for another type of resource, I don't think that's what you intended?

core/src/messages/agent.rs
```diff
@@ -380,6 +380,21 @@ pub struct SpawnRequest {
...
+pub struct ImageDownloadRequest {
+    pub cluster: ClusterName,
+
+    pub drone_id: DroneId,
+
+    pub image_url: String,
+}
+
+pub enum ResourceRequest {
+    Spawn(SpawnRequest),
```
paulgb, line 394: Since this is a request for a resource (noun), let's use nouns instead of verbs for the variants, e.g.

`ResourceRequest::Backend`, `ResourceRequest::Image`

core/src/messages/scheduler.rs
```diff
@@ -11,29 +11,24 @@ use std::{collections::HashMap, time::Duration};
 
 #[serde_as]
 #[derive(Serialize, Deserialize, Debug, Clone)]
-pub struct ScheduleRequest {
-    pub cluster: ClusterName,
-
-    /// The name of the backend. This forms part of the hostname used to
-    /// connect to the drone.
+pub struct BackendResource {
     pub backend_id: Option<BackendId>,
```
paulgb, line 15: The docstring went missing.

core/src/messages/scheduler.rs
```diff
@@ -11,29 +11,24 @@ use std::{collections::HashMap, time::Duration};
...
-    /// Metadata for the spawn. Typically added to log messages for debugging and observability.
-    pub metadata: HashMap<String, String>,
-
     /// Configuration for docker run (image, creds, env vars etc.)
     pub executable: DockerExecutableConfig,
 
     #[serde(default)]
     pub require_bearer_token: bool,
+
+    pub metadata: HashMap<String, String>,
```
paulgb, line 27: Another docstring went missing.

paulgb (commented):
I like the approach of re-using the existing infrastructure, I think the general idea of allocating arbitrary resources is a good direction for Plane.

A couple of high-level comments/questions:
- What does the full flow look like to go from running a `plane-cli image add jamcr.io/foo/bar` to having that image available on all drones?
- In particular, do we re-use the "spawn request" flow for downloading images as well?

If it's easier, maybe we could whiteboard some of this out.

controller/src/lib.rs
```diff
@@ -20,62 +24,120 @@ pub mod plan;
...
-                let spawn_request = schedule_request.value.schedule(&drone_id);
-                match nats.request(&spawn_request).await {
-                    Ok(true) => {
-                        tracing::info!(
-                            duration=?timer.duration(),
-                            backend_id=%spawn_request.backend_id,
-                            %drone_id,
-                            "Drone accepted backend."
-                        );
+                    nats.publish(&WorldStateMessage {
```
pretentious7, line 47: it'd be nice to enforce that ResourceRequests being accepted corresponded with a StateUpdate in the typesystem somehow.

cli/src/main.rs
```diff
@@ -93,6 +98,19 @@ async fn main() -> Result<()> {
         .await?;
 
     match opts.command {
+        Command::DownloadImage {
+            cluster,
+            url,
+            drone,
+        } => {
+            let result = nats
```
paulgb, line 106: unused var

dev/src/util.rs
```diff
@@ -155,18 +155,20 @@ pub async fn invalid_image_spawn_request() -> SpawnRequest {
...
-            env: vec![("PORT".into(), "8080".into())].into_iter().collect(),
-            image: TEST_IMAGE.into(),
-            credentials: None,
-            resource_limits: Default::default(),
-            pull_policy: Default::default(),
-            port: None,
-            volume_mounts: vec![],
-        },
-        require_bearer_token: false,
+        resource: Resource::Backend(BackendResource {
```
paulgb, line 158: This breaks the existing drone API, right? i.e. a running drone would not work if we updated the controller version?

paulgb (commented):
a couple more comments, but I forgot I had some from before here still as well.

dev/src/util.rs
pretentious7: Yep, I'm going to figure out how to get serde to let me keep the old way of doing things or have ScheduleResourceRequest be separate from ScheduleRequest.

But I want to know if this pr after fixes should be merged, or if I should try and coalesce things a bit such that instead of  having three different types there's traits on one type or something.

PR #364, opened 2023-04-28
Only create jetstreams from controller

paulgb:
The drone was failing because it would try to create jetstreams on load. This changes the behavior so that only the controller does this. This also means that the CLI (and other users of the `TypedNats` interface, like p2) will not create these streams, which I think is correct: the controller essentially "owns" them.

This PR also includes a version bump, dependencies update, and some other tidying.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.20"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


rolyatmax (approved):
Looks good to me. Just curious: why was this causing the drones to fail? Was it because they didn't have permissions to read `STREAM.INFO.*` which `get_or_create_stream()` relies on? Or something else?

paulgb:
Yes, they are not permitted to talk to any jetstream subjects. This wasn't a problem before because we were lazily creating jetstreams when we first used them, but since we are moving state into one jetstream, creating the jetstream at initialization reduces complexity. The Plane tests don't catch this because they do not use our production NATS subject permissions, but our platform integration tests caught it.

PR #365, opened 2023-05-01
Abhi/env test

pretentious7:


dev/test-images/buildable-test-server/resources/cgi-bin/env.cgi
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+#take one env var from query string of forme http://host/cgi-bin/env.cgi/<<varname>> 
+ENV_VAR_NAME=$(echo "${PATH_INFO}" | tail -c +2)
```
paulgb, line 3: We're always passing the same variable name, right? Seems like it would be a lot easier to follow the code if we hard-coded the variable name instead of passing it through `PATH_INFO`.

pretentious7: I didn't want it to be possible for the variable name in the test and in this file to desync.

But you're right it'll be simpler, I'll hardcode it to TEST_VAR

paulgb: Gotcha, yeah that's a good instinct but I think in this case the chances we'll touch it arbitrarily are low, and the additional burden in reading the code is high enough that I'd prefer to hard code it.

paulgb (approved):

PR #366, opened 2023-05-01
write exit code to logs if failed backend

pretentious7:


dev/tests/agent.rs
```diff
@@ -268,7 +268,11 @@ struct TestFixture {
     agent: Agent,
 }
 
-async fn do_spawn_request(mut request: SpawnRequest) -> (TestFixture, BackendStateSubscription) {
+async fn do_spawn_request(
```
paulgb, line 271: I know this pre-existed this PR, but maybe we should rename this `spawn_test_backend` to be more accurate now that it has arguments that relate to the specific backend it spawns.

drone/src/agent/executor.rs
```diff
@@ -317,7 +317,25 @@ impl<E: Engine> Executor<E> {
...
             }
         }
-
+        if let Ok(EngineBackendStatus::Failed { code }) =
+            self.engine.backend_status(spawn_request).await
+        {
+            if let Err(inject_err) = self
+                .backend_to_monitor
+                .try_get_mut(&spawn_request.backend_id)
+                .unwrap()
```
paulgb, line 326: I know this `unwrap()` should always succeed but given that this function is somewhat complex I'm a bit uneasy about relying on manual analysis to convince myself of that. Let's at least use `expect()` here rather than unwrap with a message like `Expected backend monitor missing when logging exit code` so that we can grep it if the impossible ever does happen.

dev/tests/agent.rs
pretentious7: was a mistake to put this stuff here, I removed it.

drone/src/agent/executor.rs
pretentious7: addressed.

paulgb (approved):

PR #368, opened 2023-05-04
termination request now works if fired right after spawning

### LINKED ISSUE #336: Termination request issued in the window between a backend starting and getting ready is not applied

pretentious7:
so the problem was that awaiting the database insertion was taking a bit, and within that time the backend didn't exist in the hashtable.

I'd put this in the response to the spawnrequest, but then resumption would break so I just have creating the sigchannel as a preamble to run_backend. alternatively I could put it an Option I suppose

*mostly* fixes #336 , I say mostly because ofcourse it's possible for the thread to be preempted in such a way that the backend_listener is unset when the termination request's handler fires. to solve this, I want to somehow require that a backendContext is *required* to exist prior to run_backend, but I'll have to think about that.

In the meantime I don't think it's worthwhile to do some janky barrier stuff.

### DIFF
--- a/dev/tests/agent.rs
+++ b/dev/tests/agent.rs
@@ -701,25 +701,18 @@ async fn handle_termination_request() {
         .unwrap();
 
     let mut request = base_spawn_request().await;
-    // Ensure spawnee lives long enough to be terminated.
     request.drone_id = drone_id.clone();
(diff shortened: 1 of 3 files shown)


rolyatmax (approved):
LGTM!

PR #373, opened 2023-05-10
Abhi/better soa

pretentious7:


controller/src/dns/mod.rs
```diff
@@ -58,16 +58,18 @@ impl ClusterDnsServer {
     async fn do_lookup(&self, request: &Request) -> Result<Vec<Record>> {
         let name = request.query().name().to_string();
         let (backend_id, cluster_name) = name
-            .split_once('.')
+            .strip_suffix('.')
+            .and_then(|name| name.split_once('.'))
+            .and_then(|(backend_id, potential_cluster_name)| {
+                match potential_cluster_name.matches('.').count() {
```
paulgb, line 64: I think this whole match expression could be replaced with 

```rust
potential_cluster_name.contains('.')
    .then_some((backend_id, ClusterName::new(potential_cluster_name)))
```

Not for the sake of code golf, but I think it's a bit easier to read.

paulgb (commented):
one comment, otherwise looks good

paulgb (approved):

PR #374, opened 2023-05-11
adds ability to configure authoritative domain name for DNS server

pretentious7:


controller/src/config.rs
```diff
@@ -19,6 +19,9 @@ pub struct DnsOptions {
     /// however, the email provided here should be a normal "username@domain.tld"-format
     /// email.
     pub soa_email: Option<String>,
+
+    /// The domain name at which the A record for this DNS server is
```
paulgb, line 23: I find the wording a bit awkward, how about:

"A domain name whose A record is the IP of this DNS server"?

controller/src/plan.rs
```diff
@@ -45,9 +46,25 @@ impl ControllerPlan {
                 None
             };
 
+            let domain_name = options
+                .domain_name
+                .and_then(|ref name| {
+                    Name::from_utf8(name)
+                        .or_else(|e| {
```
paulgb, line 53: I find that doing chaining inside of another chain gets pretty hard to read, maybe this could be expressed better in another way? At least for now, I'd take clippy's suggestion and use `map_err`.

paulgb (commented):
Couple small comments, otherwise LG.

controller/src/plan.rs
pretentious7: yea, I tried using anyhow and a context, but that was even trickier.

forgot about map_err though

paulgb:
Can you give this PR a more meaningful title?

paulgb (approved):
lgtm, but please rename the PR to something more meaningful before merging

PR #375, opened 2023-05-15
move drone-stats into plane as metrics

pretentious7:


paulgb:
Nice work, but looks like it's missing some deps, can you get the build to pass and then I'll have a deeper look?

pretentious7:
Wait does cargo add not add things to the cargo toml? Huh

metrics/run-stat-reports.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
paulgb, line 1: Does this script need to be in the plane codebase?

metrics/src/utils.rs
```diff
@@ -0,0 +1,41 @@
...
+use async_nats as nats;
+use serde::Serialize;
+use std::boxed::Box;
+use std::str::FromStr;
+
+#[derive(Debug, Clone)]
+pub struct NatsSubjectComponent(pub String);
+
+#[derive(Serialize)]
+pub struct DroneStatsMessage {
```
paulgb, line 12: I'm fine with this living here for now, but just flagging that we will eventually want to move it to `/core` so that we can use it in the controller for scheduling.

metrics/test-report-stats.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+drone-stats -t
```
paulgb, line 2: should this be `drone-stats test`?

Is it necessary?

.github/workflows/build-metrics-image.yml
```diff
@@ -0,0 +1,64 @@
...
+    paths-ignore: ["/docs"]
+    
+  pull_request:
+    branches: ["main"]
+    paths:
+    - '.github/workflows/build-metrics-image.yml'
+    
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: drifting-in-space/plane-metrics
```
paulgb, line 16: maybe we should use this opportunity to move it to `plane/metrics-collector` on dockerhub instead of ghcr.io? Up to you, I can do a follow-up PR if it's easier.

paulgb (commented):
Looks good, left a few comments.

metrics/run-stat-reports.sh
pretentious7: hmm I use it in the dockerfile

paulgb (approved):

metrics/run-stat-reports.sh
pretentious7: I think I could add the nats username password stuff into the runargs, then this script can go. But I think that can go in a follow up pr.

PR #376, opened 2023-05-22
TypedMessage derive

pretentious7:
adds a derive for TypedMessage

doesn't fully work yet, need to add a param for ResponseType

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,4 +7,5 @@ members = [
     "dev",
     "dev/integration-test",
     "drone",
+	"core/nats-macros"
 ]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,7 @@ members = [
     "cli",
     "controller",
     "core",
-	"core/nats-macros",
+    "core/nats-macros",
     "dev",
     "dev/integration-test",
     "drone",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,8 +4,8 @@ members = [
     "cli",
     "controller",
     "core",
+	"core/nats-macros",
     "dev",
     "dev/integration-test",
     "drone",
-	"core/nats-macros"
 ]
--- a/core/nats-macros/Cargo.toml
+++ b/core/nats-macros/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "nats-macros"
+name = "plane-core-nats-macros"
 version = "0.3.19"
 edition = "2021"
 authors = ["Abhishek Cherath <abhi@driftingin.space>", "Paul Butler <paul@driftingin.space>"]
--- a/core/nats-macros/Cargo.toml
+++ b/core/nats-macros/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "nats-macros"
+version = "0.3.19"
+edition = "2021"
+authors = ["Paul Butler <paul@driftingin.space>"]
+homepage = "https://plane.dev"
+description = "Session backend orchestrator for ambitious browser-based apps."
+repository = "https://github.com/drifting-in-space/plane"
+license = "MIT"
+readme = "README.md"
+
+[dependencies]
+async_nats = "0.29.0"
+
+[lib]
+proc-macro = true
--- a/core/nats-macros/Cargo.toml
+++ b/core/nats-macros/Cargo.toml
@@ -2,7 +2,7 @@
 name = "nats-macros"
 version = "0.3.19"
 edition = "2021"
-authors = ["Paul Butler <paul@driftingin.space>"]
+authors = ["Abhishek Cherath <abhi@driftingin.space>", "Paul Butler <paul@driftingin.space>"]
 homepage = "https://plane.dev"
 description = "Session backend orchestrator for ambitious browser-based apps."
 repository = "https://github.com/drifting-in-space/plane"
--- a/core/nats-macros/Cargo.toml
+++ b/core/nats-macros/Cargo.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 [dependencies]
 async-nats = "0.29.0"
 darling = "0.20.1"
+proc-macro2 = "1.0.59"
 quote = "1.0.27"
 syn = "2.0.16"
 
--- a/core/nats-macros/Cargo.toml
+++ b/core/nats-macros/Cargo.toml
@@ -10,7 +10,10 @@ license = "MIT"
 readme = "README.md"
 
 [dependencies]
-async_nats = "0.29.0"
+async-nats = "0.29.0"
+darling = "0.20.1"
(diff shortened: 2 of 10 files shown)


core/src/messages/agent.rs
```diff
@@ -120,7 +115,8 @@ impl JetStreamable for DroneLogMessage {
     }
 }
 
-#[derive(Serialize, Deserialize, Debug, Clone)]
+#[derive(Serialize, Deserialize, Debug, Clone, TypedMessage)]
+#[typed_message("cluster.#cluster.backend.#backend_id.stats")]
```
paulgb, line 119: `as_str()` on a `ClusterName` will result in `plane.test` rather than `plane_test` (`.subject_name()`)

Maybe we have another trait, `AsSubjectComponent`, and implement `.as_subject_component()` on our ID types? This would nullify the code savings, but it would help us avoid bugs that came up from using the wrong method (as have occurred before).

paulgb (commented):
Neat, a couple things to address though:

- How do you imagine the response type working? `#[typed_message("blah", MyReturnType)]`? And maybe it's an optional param?
- We don't always want `as_str()` of a component, sometimes we want `.subject_name()`

pretentious7:
>     * How do you imagine the response type working? `#[typed_message("blah", MyReturnType)]`? And maybe it's an optional param?


yea I was going to have it be an optional thing on the the typed_message attribute. This actually is quite simple to do using darling and syn. I was just a little confused so wrote a more basic one to start with.

pretentious7:
ready now, used darling to parse response.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "cli",
     "controller",
     "core",
+	"core/nats-macros",
```
paulgb, line 7: stray tab

core/nats-macros/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+
+#[derive(Debug)]
+struct NatsSubjectMacroInvocation(syn::Macro);
+
+impl FromMeta for NatsSubjectMacroInvocation {
+    fn from_string(value: &str) -> darling::Result<Self> {
+        let props: Vec<String> = value
+            .split('.')
+            .map(|ea| ea.strip_prefix('#'))
+            .filter(|ea| ea.is_some())
```
paulgb, line 31: I think lines 31, 32, 33 could all be replaced with one `.filter_map(|ea| ea.strip_prefix('#'))`

paulgb (commented):
I have to catch the boat but I have a couple comments so I'll submit those for now. I like what I see so far!

paulgb (commented):
I have to catch the boat but I have a couple comments so I'll submit those for now. I like what I see so far!

core/nats-macros/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "nats-macros"
```
paulgb, line 2: We should also add this to the workspace `Cargo.toml` at the root of this repo so that `cargo ws` (workspace management) is aware of it. This helps us keep versions of related crates in lockstep, among other things.

core/nats-macros/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+            .filter_map(|ea| ea.strip_prefix('#'))
+            .collect();
+        let mut fstring = value.to_string();
+        for prop in props.clone() {
+            fstring = fstring.replace(&("#".to_owned() + prop), "{}");
+        }
+        //note: this can likely be cleaned up with some judicious use of quote! and tokens!
+        let propstr = props
+            .iter()
+            .map(|prop| "self.".to_owned() + prop + ".as_subject_component()")
```
paulgb, line 39: This will only work if `AsSubjectComponent` is available in the namespace where the macro is used. Instead, I'd use a fully-qualified path, like:

    plane_core::types::AsSubjectComponent::as_subject_component(&self.prop)

E.g.:

    .map(|prop| "plane_core::types::AsSubjectComponent::as_subject_component(&self.".to_owned() + prop + ")")

core/nats-macros/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+            .reduce(|a, b| a + "," + &b)
```
paulgb, line 40: Use [join](https://doc.rust-lang.org/std/primitive.slice.html#method.join) for this.

paulgb (commented):
I like this with `AsSubjectComponent` a lot. I've written bugs that this macro would have caught.

Remaining comments are mostly macro hygiene nits, should be small fixes so I think we can merge this today.

core/nats-macros/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+
+#[proc_macro_derive(TypedMessage, attributes(typed_message))]
+pub fn typed_message_impl(input: TokenStream) -> TokenStream {
+    let ast = parse_macro_input!(input as DeriveInput);
+    let opts = Opts::from_derive_input(&ast).expect("Wrong options!");
+    let typ = ast.ident;
+    let resp_typ: syn::Type = opts.response.0;
+    let fmacro = opts.subject.0;
+    quote! {
+        impl TypedMessage for #typ {
```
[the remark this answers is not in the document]
paulgb: (This one is less important because if the user is using a derive macro they _should_ already have this in scope, but in general I like to fully-qualify everything in macros so that I don't have to think through that sort of thing when I'm reading the code)

core/nats-macros/Cargo.toml
pretentious7: uhh didnt i do that?

I added core/nats-macros to the Cargo.toml in the root.

do i have to do it somewhere else also for workspaces to work?

core/nats-macros/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+        }
+		let as_subject_component_path : syn::Path = match std::env::var("CARGO_PKG_NAME").unwrap().as_str() {
+			"plane-core" => syn::parse_quote!{ crate::types::AsSubjectComponent::as_subject_component },
+			_ => syn::parse_quote!{ plane_core::types::AsSubjectComponent::as_subject_component }
+		};
+
+        let propexpr = props
+            .iter()
+            .map(|prop| {
+				let pr : proc_macro2::TokenStream = syn::parse_str(prop).unwrap();
```
paulgb, line 46: run cargo fmt

core/nats-macros/Cargo.toml
paulgb: I think just that should be enough for cargo ws.

Let's also rename it to `plane-core-nats-macro` so that it is under the `plane-` prefix when we push it to crates.io.

paulgb (approved):
LGTM!

PR #380, opened 2023-05-30
fix double termination

pretentious7:
So the issue here is that for whatever reason after the terminate signal is received, another interrupt signal arrives as as consequence of the fact that the thing just got stopped. This then tries to terminate it again, which leads to an error.


paulgb (approved):
Good catch! Thanks for tracking this down.

paulgb (approved):
Good catch! Thanks for tracking this down.

PR #386, opened 2023-06-01
version bump 0.3.21

rolyatmax:
Once #385 is landed, we can rebase this PR


paulgb:
Let's get #387 in first too.

PR #390, opened 2023-06-06
Abhi/fix no sweeps

pretentious7:


pretentious7:
I misunderstood how the destructor code gets called. It gets called in the transition for
a terminal state, not TO a terminal state.

drone/src/agent/executor.rs
```diff
@@ -265,30 +265,36 @@ impl<E: Engine> Executor<E> {
...
-                        },
-                        None => {
-                            if state.terminal() {
-                                tracing::info!(%state, "Channel closed, backend in terminal state");
-                            } else {
-                                tracing::error!(
-                                    %state, "Channel closed but backend not in terminal state");
+                if state.terminal() {
+                    //we stop the signal handler to ensure that there are no double terminations
+                    recv.close();
```
paulgb, line 270: Can you remind me why this is necessary? It seems like if the state is terminal, `self.step` will return `Ok(None)` and then we get to the end and `recv` will be dropped when it is closed, no?

paulgb (commented):
Thanks for the fast fix, one question inline.

drone/src/agent/executor.rs
pretentious7: yea... this is the right solution to the double termination issue, not closing the signal handler

this was only necessary when the if condition only checked for Swept, since in that case the self.step(req, BackendState::Terminated) would cause a Signal::Interrupt before returning but after killing the backend, which got it back to the select! and ran the termination twice.

paulgb (approved):
looks good

rolyatmax:
299b looks good now

rolyatmax (approved):
Termination and Sweeping both work

PR #392, opened 2023-06-07
Use durable consumers for everything for now

paulgb:


paulgb:
Going to cancel this and wait for the duration solution to be merged into `async_nats`.

PR #395, opened 2023-06-09
Create tutorial.html

pretentious7:
not really sure if this will work, but let's see


paulgb:
I'm glad to have this written out so that we can start simplifying the approach. Two things that would remove steps from this tutorial are:

- Supporting HTTP and not requiring a certificate
- Supporting URL-based routing instead of host-based

I have a feeling the latter is going to be a better complexity/power trade-off for single-node users than mdns, what do you think?

Aside: is there a way to get org-mode to output markdown instead of HTML? GitHub is surprisingly bad at previewing HTML.

pretentious7:
So @rolyatmax and i discussed url based routing a while back, and came to the conclusion that it would be too disruptive to the codebase.

The advantage with mdns is that it lowers the tryout threshold while not requiring any core code changes, just that the dns loop talk to avahi/bonjour instead of doing its own dns server.

PR #396, opened 2023-06-12
Set inactive threshold for NATS consumers

paulgb:
Updates the async-nats version and uses the new `inactive_threshold` parameter to keep the consumer alive for longer.

### DIFF
--- a/core/src/nats.rs
+++ b/core/src/nats.rs
@@ -17,11 +17,18 @@ use std::fmt::Debug;
 use std::future::Future;
 use std::marker::PhantomData;
 use std::task::Poll;
+use std::time::Duration;
 use time::OffsetDateTime;
 use tokio_stream::{Stream, StreamExt};
 
 /// This code is returned by NATS when an expected last sequence number is violated.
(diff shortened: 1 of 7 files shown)


core/src/nats.rs
```diff
@@ -17,12 +17,19 @@ use std::fmt::Debug;
...
 use tokio_stream::{Stream, StreamExt};
 
 /// This code is returned by NATS when an expected last sequence number is violated.
 const NATS_WRONG_LAST_SEQUENCE_CODE: &str = "10071";
 
+/// How long NATS should keep consumer state before forgetting it.
+/// This is effectively how long a subscriber can sustain a NATS outage. Even though
+/// we expect NATS outages to be short, there's little harm in making this too long;
+/// consumer state is minimal and we make efficient use of consumers.
+const INACTIVE_THRESHOLD_SECONDS: u64 = 60 * 60 * 12; // 12 hours
```
rolyatmax, line 31: What does consumer state consist of?

rolyatmax (approved):

PR #397, opened 2023-06-12
Make locks work with bearer tokens

paulgb:
If a backend has a bearer token and a lock, and we send a spawn request with the same lock, the returned result now includes the bearer token.

This works by attaching the bearer token to the drone assignment message.

### DIFF
--- a/core/src/messages/state.rs
+++ b/core/src/messages/state.rs
@@ -86,6 +86,7 @@ pub enum BackendMessageType {
     Assignment {
         drone: DroneId,
         lock: Option<String>,
+        bearer_token: Option<String>,
     },
     State {
         state: BackendState,
--- a/dev/tests/scheduler.rs
+++ b/dev/tests/scheduler.rs
@@ -434,7 +434,7 @@ async fn schedule_request_lock() {
 
     sleep(Duration::from_millis(100)).await;
     let r1 = mock_agent
-        .schedule_drone(&drone_id, true, Some("foobar".to_string()))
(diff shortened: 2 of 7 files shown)


dev/tests/scheduler.rs
```diff
@@ -452,3 +452,49 @@ async fn schedule_request_lock() {
...
+            drone_id: drone_id.clone(),
+            drone_version: PLANE_VERSION.to_string(),
+            ready: true,
+            state: DroneState::Ready,
+            running_backends: None,
+        })
+        .await
+        .unwrap();
+
+    sleep(Duration::from_millis(100)).await;
```
pretentious7, line 479: what condition is this sleep to ensure?

pretentious7 (approved):
LGTM!

core/src/messages/state.rs
```diff
@@ -86,6 +86,7 @@ pub enum BackendMessageType {
     Assignment {
         drone: DroneId,
         lock: Option<String>,
+        bearer_token: Option<String>,
```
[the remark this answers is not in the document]
paulgb: My thinking is that any UTF-8 string is valid as far as Plane is concerned. It is only used for comparisons.

dev/tests/scheduler.rs
paulgb: It was in the test I copied, I've forgotten at this point. I agree that we should go back and remove sleeps, but I'd rather merge with some sleeps in test for now and then clean up later.

PR #398, opened 2023-06-12
Redirect HTTP requests to HTTPS

pretentious7:
redirects a configured http_port (80 by default) to https using a 308 in the proxy. Plus a test.

### DIFF
--- a/dev/tests/proxy.rs
+++ b/dev/tests/proxy.rs
@@ -224,6 +224,21 @@ async fn simple_backend_proxy() {
         .await
         .unwrap();
 
+    let hostname = format!("{}.{}", "foobar", CLUSTER);
+    let client = ClientBuilder::new()
+        .resolve(&hostname, proxy.bind_redir_address)
+        .build()
+        .unwrap();
+
+    let req = client.get(format!(
+        "http://{}:{}{}",
+        hostname,
+        proxy.bind_redir_address.port(),
+        "/"
+    ));
+    let response = req.send().await.unwrap();
+
+    /*
     let response = reqwest::get(format!(
         "http://{}.{}:{}/{}",
         "foobar",
@@ -233,6 +248,7 @@ async fn simple_backend_proxy() {
     ))
     .await
     .expect("Failed to send request");
+    */
 
     assert_eq!(StatusCode::PERMANENT_REDIRECT, response.status());
     let mut https_uri = response.url().clone();
--- a/dev/tests/proxy.rs
+++ b/dev/tests/proxy.rs
@@ -232,10 +232,10 @@ async fn simple_backend_proxy() {
         .unwrap();
 
     let req = client.get(format!(
-        "http://{}:{}{}/TESTING/PATHS",
+        "http://{}:{}{}",
         hostname,
         proxy.bind_redir_address.port(),
-        "/"
+        "/testing/paths"
     ));
     let response = req.send().await.unwrap();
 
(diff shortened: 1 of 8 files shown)


dev/tests/proxy.rs
```diff
@@ -220,6 +224,29 @@ async fn simple_backend_proxy() {
...
+    let response = req.send().await.unwrap();
+
+    assert_eq!(StatusCode::PERMANENT_REDIRECT, response.status());
+    let mut https_uri = response.url().clone();
+    https_uri.set_scheme("https").unwrap();
+    assert_eq!(
+        response.headers().get("LOCATION").unwrap(),
+        https_uri.as_str()
+    );
+
```
rolyatmax, line 249: Should we have an assertion that makes sure the path matches too?

sample-config/plane-config/drone.toml
```diff
@@ -54,7 +54,7 @@ ip = { api = "http://ip-api:8080/" }
 # Optional Docker settings for the agent.
 [agent.docker]
 # The runtime to use (defaults to "runc")
-runtime = "runc"
+# runtime = "runc"
```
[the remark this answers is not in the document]
pretentious7: yea, the problem with specifying the runtime is that docker hardcodes search paths, so if runc is somewhere other than the usual bin areas it won't work, whereas if we just leave it it uses the default system configured runc.

basically it improves the situation a little bit for me and makes no difference to anyone else.

dev/tests/proxy.rs
pretentious7: I'm checking the whole URL, so that includes the path.

pretentious7: but I think I'm misunderstanding you.

paulgb: I see what you mean. You could throw `/foo/bar` into the path of the request so that an implementation that failed to preserve the path would fail.

sample-config/plane-config/drone.toml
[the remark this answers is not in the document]
pretentious7: Yea so distros and such generally ship docker with the runc path configured in daemon.json, whereas if you pass in runc as the executor it goes hunting.

drone/src/proxy/mod.rs
```diff
@@ -66,7 +73,41 @@ async fn run_server(options: ProxyOptions, connection_tracker: ConnectionTracker
...
+                }))
+            });
+            Server::bind(&redirect_address).serve(redirect_service)
+        };
+        tokio::select! {
+            a = server => a.context("Error from TLS proxy.")?,
+            b = redirect_server => b.context("Error from HTTP redirect.")?
+        }
     } else {
         let server = Server::bind(&bind_address).serve(make_proxy);
```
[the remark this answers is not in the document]
paulgb: we only redirect over HTTP right? You still need keys because HTTPS still needs to work?

drone/src/plan.rs
```diff
@@ -48,12 +48,22 @@ impl DronePlan {
         };
 
         let proxy_options = if let Some(proxy_config) = config.proxy {
+            let (bind_port, bind_redir_port, key_pair) = if let Some((key_pair, https_port)) =
```
paulgb, line 51: The zip makes this hard to read, I'd prefer

    if let (Some(key_pair), Some(https_port)) = (config.cert, proxy_config.https_port)

paulgb (commented):
Getting close, one nit.

paulgb (approved):

PR #399, opened 2023-06-13
Cleanup cleans up abandoned backends

paulgb:
Any backend not in the `Ready` state and for which we haven't heard from in 24 hours is removed from controller's world state.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -108,12 +108,14 @@ async fn main() -> Result<()> {
             for (cluster_name, cluster) in &state.clusters {
                 for (backend_id, backend) in &cluster.backends {
                     if let Some((timestamp, state)) = backend.state_timestamp() {
-                        if state == BackendState::Swept || state == BackendState::Terminated {
+                        if state.terminal() {
                             println!(
                                 "Removing backend {} from cluster {}, {} at {}",
                                 backend_id, cluster_name, state, timestamp
                             );
                         } else if timestamp < chrono::Utc::now() - chrono::Duration::hours(24) && (state != BackendState::Ready) {
+                            // This catches backends that remained in a non-ready and non-terminal state (Loading or Starting)
+                            // for more than 24 hours. This either means we lost the drone, or the drone has a bug.
                             println!(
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -108,13 +108,20 @@ async fn main() -> Result<()> {
...
+                        if state == BackendState::Swept || state == BackendState::Terminated {
+                            println!(
+                                "Removing backend {} from cluster {}, {} at {}",
+                                backend_id, cluster_name, state, timestamp
+                            );
+                        } else if timestamp < chrono::Utc::now() - chrono::Duration::hours(24) && (state != BackendState::Ready) {
+                            println!(
+                                "Removing abandoned backend {} from cluster {}, last update was {} at {}",
+                                backend_id, cluster_name, state, timestamp
+                            );
```
rolyatmax, line 120: Should we ensure those backend containers are no longer running by telling the drones to stop them and then set their statuses to `Lost`?

paulgb: That should really be handled by the drone itself. The platform currently has no way of telling the drone that a backend should be considered lost. This is more of a catch to ensure that if the drone is buggy we don't end up accumulating a bunch of dead state.

cli/src/main.rs
```diff
@@ -108,13 +108,20 @@ async fn main() -> Result<()> {
             for (cluster_name, cluster) in &state.clusters {
                 for (backend_id, backend) in &cluster.backends {
                     if let Some((timestamp, state)) = backend.state_timestamp() {
-                        if state != BackendState::Swept {
+                        if state == BackendState::Swept || state == BackendState::Terminated {
```
rolyatmax, line 111: Should we include all terminal states here, not just swept and terminated? (i.e. `Exited`, `Failed`, `ErrorLoading`, `ErrorStarting`, `TimedOutBeforeReady`, and `Lost`)

cli/src/main.rs
paulgb: Done.

rolyatmax (approved):

PR #400, opened 2023-06-13
remove need to specify https_port

pretentious7:
reasoning here is as follows:
if certs provided, with no https port -> serve https on 443
if certs proved, https_port provided -> serve https on https_port
if certs not provided, https_port provided -> panic
if certs not provided, https_port not provided -> serve http

the change from the previous pr is in case 1. while there is some semantic value in requiring the https_port to be specified, it seems reasonable to infer that a user providing keys will want https, and it won't break existing configs.


paulgb (approved):
Good change, thanks.

PR #401, opened 2023-06-15
Abhi/fix lock race condition

pretentious7:


controller/src/lib.rs
```diff
@@ -102,32 +109,65 @@ pub async fn run_scheduler(nats: TypedNats, state: StateHandle) -> NeverResult {
...
-                        ScheduleResponse::NoDroneAvailable
-                    }
-                    Err(error) => {
-                        tracing::warn!(?error, "Scheduler returned error.");
-                        ScheduleResponse::NoDroneAvailable
+                    if let Some(lock) = schedule_request.value.lock.clone() {
+                        let notify = std::sync::Arc::new(Notify::new());
+                        lock_to_ready.insert(lock, notify.clone());
+                        tokio::spawn(async move {
+                            while state.state().get_logical_time() < seq_id {
```
paulgb, line 116: I don't like polling here, I implemented an alternative here: https://github.com/drifting-in-space/plane/pull/403

controller/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashMap;
+
```
paulgb, line 2: rm extra line and run `cargo fmt` again after to sort imports

controller/src/lib.rs
```diff
@@ -20,114 +25,204 @@ pub mod plan;
...
-                    let state = state.state();
-                    let cluster = state
-                        .cluster(&schedule_request.value.cluster)
-                        .ok_or_else(|| anyhow!("Expected cluster to exist."))?;
-
-                    let backend_state = cluster
-                        .backend(&backend)
-                        .ok_or_else(|| anyhow!("Lock held by a backend that doesn't exist."))?;
+async fn spawn_backend(
+    ref nats: TypedNats,
```
paulgb, line 29: I don't think `ref` is necessary here? We are cloning it when we pass it in anyway, I don't see why we need it as a reference.

controller/src/lib.rs
```diff
@@ -20,114 +25,204 @@ pub mod plan;
...
+                        message: BackendMessageType::Assignment {
+                            drone: drone.clone(),
+                            lock: schedule_request.lock.clone(),
+                            bearer_token: spawn_request.bearer_token.clone(),
+                        },
+                    }),
+                })
+                .await?;
+
+            tracing::info!(
```
paulgb, line 58: indentation is off; might need to format this manually since it's in a macro

controller/src/lib.rs
```diff
@@ -20,114 +25,204 @@ pub mod plan;
...
+            tracing::warn!(?error, "Scheduler returned error.");
+            Ok((ScheduleResponse::NoDroneAvailable, None))
+        }
+    }
+}
 
-                    let drone = backend_state.drone.clone().ok_or_else(|| {
-                        anyhow!("Lock held by a backend without a drone assignment.")
-                    })?;
+fn locked_backend(
```
paulgb, line 84: This could be more descriptive, maybe `backend_of_lock` or add a docstring to that effect?

controller/src/lib.rs
```diff
@@ -20,114 +25,204 @@ pub mod plan;
...
+    state
+        .state()
+        .cluster(cluster_name)
+        .ok_or(anyhow!("no cluster"))?
+        .locked(&lock)
+        .ok_or(anyhow!("no backend"))
+}
 
-                    let bearer_token = backend_state.bearer_token.clone();
+fn fetch_backend(
```
paulgb, line 97: I think of "fetch" as implying that a network trip or at least hitting disk is involved. Maybe `schedule_request_for_existing_backend`?

controller/src/lib.rs
```diff
@@ -20,114 +25,204 @@ pub mod plan;
...
+    state: &StateHandle,
+    cluster: ClusterName,
+    backend: BackendId,
+) -> anyhow::Result<ScheduleResponse> {
+    // Anything that fails to find the drone results in an error here, since we just
+    // checked that the lock is held which implies that the drone exists.
+    let state = state.state();
+    let Some(cluster) = state.cluster(&cluster) else {
+		panic!()
```
paulgb, line 106: cargo fmt, but also this should not be a panic

core/src/state/world_state.rs
```diff
@@ -65,32 +65,59 @@ impl StateHandle {
...
-    listeners: HashMap<u64, Arc<Notify>>,
+    listeners: RwLock<HashMap<u64, ClosableNotifier>>,
 }
 
 impl WorldState {
-    pub fn get_listener(&mut self, sequence: u64) -> Arc<Notify> {
-        match self.listeners.entry(sequence) {
-            Entry::Occupied(entry) => entry.get().clone(),
+    pub fn get_listener(&self, sequence: u64) -> Result<ClosableNotify, SequenceNumberInThePast> {
+        if self.logical_time >= sequence {
```
paulgb, line 111: Good catch, but tather than an error in this case, should we just return a notifier which resolves immediately? Maybe `ClosableNotifier`'s `notifier` becomes an `Option`, and if it is `None`, `notified` does not await anything.

dev/tests/scheduler.rs
```diff
@@ -129,6 +129,7 @@ impl MockAgent {
                 "Backend should be scheduled on the expected drone."
             );
         }
+        println!("schedule_drone {:?}; lock {:?}", result, lock);
```
paulgb, line 132: remove `println`s; if it's useful to log these use tracing.

dev/tests/scheduler.rs
```diff
@@ -449,6 +452,19 @@ async fn schedule_request_lock() {
...
+		Ok(ScheduleResponse::Scheduled { spawned: a, .. }),
+		Ok(ScheduleResponse::Scheduled { spawned: b, .. })
+	) = tokio::join!(
+			mock_agent.schedule_drone(&drone_id, false, Some("speedlock".to_string()))
+		,
+		async {
+			tokio::time::sleep(Duration::from_millis(300)).await;
+			mock_agent.schedule_locked(false, Some("speedlock".to_string())).await
+		}
+	) else { panic!("hello") };
```
paulgb, line 466: This will be hard to debug if it catches; instead prefer:
- assign each side of the pair to its own variable
- use a `let _ else {}` for each of them, with a panic that at least tells what the problem is

controller/src/lib.rs
```diff
@@ -20,114 +25,204 @@ pub mod plan;
...
+            tracing::info!(?lock, "scheduling lock");
+            let cluster_name = sr.cluster.clone();
+
+            let Ok(response) = dispatch(
+                state.clone(),
+                cluster_name,
+                sr.clone(),
+                scheduler.clone(),
+                nats.clone(),
+                sr.lock.clone().zip(lock_mut_maybe)
```
paulgb, line 214: I'm not a fan of zips on options. Probably cleanest to create a named variable from this.

paulgb (commented):
Left some comments, mostly surface level, but getting them cleared up will help me grok the rest.

PR #403, opened 2023-06-15
Implement wait_for_seq

paulgb:
This add an async method `wait_for_seq` to `StateHandle` that blocks until the given sequence number is reached.

### DIFF
--- a/core/src/state/world_state.rs
+++ b/core/src/state/world_state.rs
@@ -10,12 +10,12 @@ use crate::{
 use anyhow::{anyhow, Result};
 use chrono::{DateTime, Utc};
 use dashmap::DashMap;
-use tokio::sync::Notify;
 use std::{
     collections::{BTreeMap, BTreeSet, VecDeque},
     net::IpAddr,
     sync::{Arc, RwLock, RwLockReadGuard, RwLockWriteGuard},
 };
+use tokio::sync::Notify;
 
 #[derive(Default, Debug, Clone)]
 pub struct StateHandle {
@@ -34,7 +34,10 @@ impl StateHandle {
         cluster: &ClusterName,
         timestamp: DateTime<Utc>,
     ) -> Result<Vec<DroneId>> {
-        let world_state = self.state.read().expect("Could not acquire world_state lock.");
+        let world_state = self
+            .state
+            .read()
+            .expect("Could not acquire world_state lock.");
 
         let min_keepalive = timestamp - chrono::Duration::seconds(30);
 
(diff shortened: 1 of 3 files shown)


core/src/state/world_state.rs
```diff
@@ -47,25 +56,51 @@ impl StateHandle {
...
 
 #[derive(Default, Debug)]
 pub struct WorldState {
     logical_time: u64,
     pub clusters: BTreeMap<ClusterName, ClusterState>,
+    listeners: Arc<DashMap<u64, Vec<Arc<Notify>>>>,
 }
 
 impl WorldState {
+    pub fn add_listener(&mut self, sequence: u64, notify: Arc<Notify>) {
```
pretentious7, line 89: Beautiful. Beautiful. Gj.

core/src/state/world_state.rs
```diff
@@ -47,25 +56,51 @@ impl StateHandle {
...
+        self.write_state().add_listener(sequence, notify.clone());
+        notify.notified().await;
     }
 }
 
 #[derive(Default, Debug)]
 pub struct WorldState {
     logical_time: u64,
     pub clusters: BTreeMap<ClusterName, ClusterState>,
+    listeners: Arc<DashMap<u64, Vec<Arc<Notify>>>>,
```
pretentious7, line 85: Does this need to be in an arc?

pretentious7: Also if you use notify_all should be able to get rid of the Vec.

pretentious7:
Also, this makes it so that the worldstate can no longer be recreated from messages. Maybe we should put the listener on the StateHandle that holds the WorldState?

core/src/state/world_state.rs
paulgb: I couldn't figure out what you mean by notify_all but I realized I could use a broadcast channel so I did that.

Good catch, it doesn't need to be an arc, or a dashmap for that matter.

paulgb:
> Also, this makes it so that the worldstate can no longer be recreated from messages. Maybe we should put the listener on the StateHandle that holds the WorldState?

I agree in principle, but the worldstate handle doesn't have an easy way to hook into when the state is updated. I am ok with this because a valid worldstate can still be recreated from the messages, it just might be missing some ephemeral state (listeners).

core/src/state/world_state.rs
pretentious7: https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html#method.notify_waiters

my b got the method name wrong, but a broadcast channel is fine too.

paulgb: ooh yeah I like that, it feels lighter-weight

pretentious7 (approved):
LGTM!

core/src/state/world_state.rs
pretentious7: Idk about weight, I'd bet they're basically the same efficiency wise, easier to read with the notify though. your call.

PR #404, opened 2023-06-15
Use persistent account key for cert refresh

paulgb:
Some ACME2 providers (google) don't like the same EAB keys to be used for different accounts. Accounts are identified with the private key passed to the ACME2 server, which must be different from the certificate signing key.

This change adds an option to specify an `account_key_path` which is a path. If it is set, we write an account key there the first time it is accessed, and then read that account key thereafter.

This adds a named dependency to `openssl` so that we can generate a key used by `acme2_eab`. `acme2_eab` already depends on `openssl`, so this does not actually add another dependency to the build.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -113,7 +113,9 @@ async fn main() -> Result<()> {
                                 "Removing backend {} from cluster {}, {} at {}",
                                 backend_id, cluster_name, state, timestamp
                             );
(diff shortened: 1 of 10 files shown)


rolyatmax (approved):

pretentious7:
I'm very confused by what google is doing here tbh, because if the key exists then shouldn't it return the existing account?

pretentious7 (approved):

PR #414, opened 2023-06-17
Refactor agent to use fine-grained supervisors

paulgb:


paulgb:
Previously, all of the agent tasks ran under one supervisor, meaning that if any of them crashed they would all be restarted. This change runs them each as supervisors, so that a failure of one does not bring down the others. It also gives tokio more leeway for scheduling them since they do not have to run on the same thread.

rolyatmax:
> This change runs them each as supervisors, so that a failure of one does not bring down the others.

@paulgb Just curious - is there an advantage to having these loops fail/restart individually rather than having them all restart when one fails?

paulgb:
> > This change runs them each as supervisors, so that a failure of one does not bring down the others.
> 
> @paulgb Just curious - is there an advantage to having these loops fail/restart individually rather than having them all restart when one fails?

For one thing, there will be a period of time where they do not have an active subscription (because the task that held the subscription was dropped, but the new one hasn't requested the subscription yet) in which messages will be lost. It's not a big deal, because the sender will retry if it is important and the message isn't acknowledged.

The bigger fear I have is that if any of these gets stuck in a retry loop for some reason, it will bring the other loops down with it, because they will be constantly being restarted.

rolyatmax (approved):

PR #418, opened 2023-06-26
Abhi/refactor scheduler

pretentious7:
Refactors the scheduler to make it more amenable to locking fixes by handling each schedule req in its own fiber.

### DIFF
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -52,10 +52,7 @@ async fn spawn_backend(
                 })
                 .await?;
 
-            tracing::info!(
-            logical_time=?seq_id,
-            "backend state updated at time"
-                );
+            tracing::info!(logical_time=?seq_id, "backend state updated at time");
 
             Ok((
                 ScheduleResponse::Scheduled {
--- a/controller/src/lib.rs
+++ b/controller/src/lib.rs
@@ -21,12 +21,11 @@ pub mod plan;
 pub mod run;
 mod scheduler;
 
-type LogicalTime = u64;
 async fn spawn_backend(
     nats: &TypedNats,
     drone: &DroneId,
     schedule_request: &ScheduleRequest,
-) -> anyhow::Result<(ScheduleResponse, Option<LogicalTime>)> {
+) -> anyhow::Result<ScheduleResponse> {
     let timer = Timer::new();
     let spawn_request = schedule_request.schedule(drone);
     match nats.request(&spawn_request).await {
@@ -54,23 +53,20 @@ async fn spawn_backend(
 
             tracing::info!(logical_time=?seq_id, "backend state updated at time");
 
-            Ok((
-                ScheduleResponse::Scheduled {
-                    drone: drone.clone(),
-                    backend_id: spawn_request.backend_id,
-                    bearer_token: spawn_request.bearer_token.clone(),
-                    spawned: true,
-                },
-                Some(seq_id),
-            ))
+            Ok(ScheduleResponse::Scheduled {
+                drone: drone.clone(),
+                backend_id: spawn_request.backend_id,
+                bearer_token: spawn_request.bearer_token.clone(),
+                spawned: true,
+            })
         }
         Ok(false) => {
             tracing::warn!("Drone rejected backend.");
-            Ok((ScheduleResponse::NoDroneAvailable, None))
+            Ok(ScheduleResponse::NoDroneAvailable)
         }
         Err(error) => {
             tracing::warn!(?error, "Scheduler returned error.");
-            Ok((ScheduleResponse::NoDroneAvailable, None))
+            Ok(ScheduleResponse::NoDroneAvailable)
         }
     }
 }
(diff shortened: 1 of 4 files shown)


controller/src/lib.rs
```diff
@@ -20,114 +21,184 @@ pub mod plan;
...
+                            drone: drone.clone(),
+                            lock: schedule_request.lock.clone(),
+                            bearer_token: spawn_request.bearer_token.clone(),
+                        },
+                    }),
+                })
+                .await?;
+
+            tracing::info!(
+            logical_time=?seq_id,
```
paulgb, line 56: looks like rustfmt is missing the indentation here

controller/src/lib.rs
```diff
@@ -20,114 +21,184 @@ pub mod plan;
 pub mod run;
 mod scheduler;
 
+type LogicalTime = u64;
+async fn spawn_backend(
+    nats: &TypedNats,
+    drone: &DroneId,
+    schedule_request: &ScheduleRequest,
+) -> anyhow::Result<(ScheduleResponse, Option<LogicalTime>)> {
```
paulgb, line 29: It looks like the `Option<LogicalTime>` is never actually used, am I missing something?

controller/src/lib.rs
```diff
@@ -20,114 +21,184 @@ pub mod plan;
...
+    if let Some(lock_name) = sr.lock.clone() {
+        tracing::info!(?lock_name, "scheduling lock");
+
+        if let Ok(backend) = backend_of_lock(state, &cluster_name, &lock_name) {
+            tracing::info!(?backend, "fetch preexisting backend");
+            return schedule_response_for_existing_backend(state, cluster_name, backend);
+        }
+
+        tracing::info!("spawn with lock");
+        let drone = match scheduler.schedule(&cluster_name, Utc::now()) {
```
paulgb, line 148: I'm struggling to see a difference in behavior between the code path here and the code path below here and the `else`-branch, is there one?

If not, we could probably remove the `else` branch and have the code be in the main function body, and have the execution just drop through here in the else-case.

paulgb (commented):
I like the decomposition of spawning. Left a few comments, may have more later.

controller/src/lib.rs
pretentious7: Yea all the finicky locking stuff ends up here, It's in the big pr, I'll have the specific one for that up tom.

controller/src/lib.rs
pretentious7: Yea, it's used later for get_listener().

controller/src/lib.rs
paulgb: yeah, it's kind of hard to review this without the rest of the context :/

pretentious7: hmm, I think I'll minimize this one to just do the spawn per schedule req (I figured that was mostly orthogonal so I could separate it out)

controller/src/lib.rs
paulgb: Which PR is that in?

paulgb: This PR looks good otherwise but I'd like to see how the `Option<LogicalTime>` part is used in context because I generally think returning tuples is a code smell and suspect there may be a better way.

I suggest this path forward:
- Change this function to return a plain `anyhow::Result<ScheduleResponse>`. If you think this will take more than 5 minutes let me know.
- Request a review again and I'll LGTM it.
- In another PR, add the message type to request an existing lock without spawning (I'd imagine that PR will benefit from this refactoring)
- Once that's done we can have a design discussion about locking. It might be worth going straight to doing a two-phase commit.

controller/src/lib.rs
```diff
@@ -20,114 +21,168 @@ pub mod plan;
...
+		&state,
+		&schedule_request.value.clone(),
+		&scheduler,
+		&nats
+	).await else {
+		tracing::error!(?schedule_request.value, "schedule request failed");
+		panic!("failed to dispatch");
+	};
+
+    let Ok(_) = schedule_request.respond(&response).await else {
```
paulgb, line 165: huh, rustfmt doesn't understand let-else yet so it just... allows tabs?

https://github.com/rust-lang/rustfmt/issues/4914

paulgb (commented):
looks good, just a formatting nit and then I think we're good!

controller/src/lib.rs
```diff
@@ -20,114 +21,168 @@ pub mod plan;
...
+    nats: TypedNats,
+) -> anyhow::Result<()> {
+    let Ok(response) = process_response(
+		&state,
+		&schedule_request.value.clone(),
+		&scheduler,
+		&nats
+	).await else {
+		tracing::error!(?schedule_request.value, "schedule request failed");
+		panic!("failed to dispatch");
```
paulgb, line 162: Rather than panic here, let's just exit. The panic is lost anyway and we're already logging a message that will be more relevant than the panic message, so it's less noisy just to exit.

paulgb (approved):

PR #419, opened 2023-06-26
Abhi/closable notify

pretentious7:


pretentious7:
As I have mentioned in comment:

We want a thing that waits until notified, and importantly waits
for that notify on all of its replicas.
The default notifier from tokio does not behave the way we need
since, if multiple clones are made, notify_one only causes the first
to return immediately when it invokes the notification and awaits.
notify_waiters does not cause even the first one to do that.
Therefore, we implement our own notifier that can be closed and
will return immediately on all its dependent notifies.

This PR adds that ClosableNotify, as well as making the needed changes to get_listener.

core/src/state/world_state.rs
```diff
@@ -65,31 +65,73 @@ impl StateHandle {
...
+    }
+}
+
+impl ClosableNotifier {
+    fn new() -> Self {
+        let (tx, _rx) = tokio::sync::broadcast::channel(128);
+        Self { notifier: tx }
+    }
+
+    fn notify(&self) -> ClosableNotify {
```
paulgb, line 100: nit, but I would just implement `Clone` since semantically that's what's happening here. Unfortunately we can't derive clone because `self.notifier` is not `Clone`, but you can use the same body you used here for an explicit `impl Clone`.

core/src/state/world_state.rs
```diff
@@ -65,31 +65,73 @@ impl StateHandle {
...
-        match self.listeners.entry(sequence) {
-            Entry::Occupied(entry) => entry.get().clone(),
+    pub fn logical_time(&self) -> u64 {
+        self.logical_time
+    }
+
+    pub fn get_listener(&self, sequence: u64) -> ClosableNotify {
+        if self.logical_time >= sequence {
+            let mut notifier = ClosableNotifier::new();
+            let notify = notifier.notify();
```
paulgb, line 124: Is this clone necessary? `notifier` gets dropped anyway.

What if instead of `get_listener`, we combine this with the (former) async `wait_for_seq`, such that it is an async function. That way this whole branch could just be an immediate `return`, like it was in `wait_for_seq`.

core/src/state/world_state.rs
```diff
@@ -65,31 +65,73 @@ impl StateHandle {
...
+        let _ = self.notifier.send(());
     }
 }
 
 #[derive(Default, Debug)]
 pub struct WorldState {
     logical_time: u64,
     pub clusters: BTreeMap<ClusterName, ClusterState>,
-    listeners: HashMap<u64, Arc<Notify>>,
+    listeners: RwLock<HashMap<u64, ClosableNotifier>>,
```
paulgb, line 113: let's just use a `DashMap` instead of an `RwLock<HashMap<..>>`

core/src/state/world_state.rs
```diff
@@ -100,8 +142,10 @@ impl WorldState {
         cluster.apply(message.message);
         self.logical_time = sequence;
 
-        if let Some(sender) = self.listeners.remove(&sequence) {
-            sender.notify_waiters();
+        if self.listeners.read().unwrap().contains_key(&sequence) {
```
paulgb, line 145: Technically there is a race condition here (albeit an improbable one): if another thread adds the value between the check and the next line, we miss it. Using a `DashMap` will fix this anyway, but in general my inclination is to write-lock for a check-and-set operation (like `remove`) even if the write might not happen.

paulgb (commented):
Nice work, I like this approach. Left a few comments.

core/src/state/world_state.rs
pretentious7: I was running into a lot of deadlocks with dashmap, RwLock<HashMap> is much easier to understand.

core/src/state/world_state.rs
pretentious7: I actually did think of that, but since this happens *after* self.logical_time=sequence, the other thread should return with an activated notify.

I guess it's possible for the other thread to clear the check in get_listener, then get preempted, have this one cross the read lock and then have that thread get the write lock. ugh. I really wanted to avoid having to get a write lock here most of the time since it'll contend hard :(

I'd *really* rather not use DashMap, I wasted like 3 hours debugging weird deadlocks in that library. 

I'll see if there's a better solution, but in the meantime I'll just take a write lock and use the entry API. If/when we need more perf I can switch to doing atomic shenanigans on self.logical_time

core/src/state/world_state.rs
pretentious7: I'll fix that.

As for the second thing, making this async is iffy since I can't hold an RwLock across an await. I messed with this a little before rejecting it. Now I could use a tokio::RwLock, but then apply would have to become async, which is boo. (And DashMap, *especially* in this case, deadlocks like the US parliament.)

(I'm being a little mean, this is all in its documentation, but doesn't give any errors... probably PEBCAK tho)

core/src/state/world_state.rs
pretentious7: ooo, I like that!

paulgb:
This is surpassed by #422, right? Mind if I close it?

PR #422, opened 2023-06-27
remove sleeps for scheduler test.

pretentious7:
Removes all sleeps from the scheduler tests

works by running a list of debugconditions (which are just Fn(&WorldState) -> bool) on every state update, and notifying when they return true. This code is conditionally compiled only in debug, so does not impact prod.

I added some utility to debugconditions for drone readiness and unreadiness.

also added lazy_static for Plane_Cluster, but this isn't *necessary* I can remove it if that's better.


dev/tests/scheduler.rs
```diff
@@ -432,40 +453,59 @@ async fn schedule_request_lock() {
...
+            Ok(ScheduleResponse::Scheduled { spawned: a, .. }),
+            Ok(ScheduleResponse::Scheduled { spawned: b, .. }),
+        ) => assert_eq!(!a, b),
+        other => {
+            tracing::error!(
+                ?other,
+                concat!(
+                    "simultaneous schedule requests failed to produce two schedule",
+                    "responses with one spawned and the other retrieved"
+                )
```
pretentious7, line 489: this fails presently, triggers the race condition.

core/src/state/world_state.rs
```diff
@@ -65,44 +66,134 @@ impl StateHandle {
...
+        Self {
+            notifier: self.notifier.clone(),
+            rx: self.notifier.subscribe(),
         }
-        let receiver = { self.write_state().get_listener(sequence) };
-        receiver.notified().await;
     }
 }
 
+pub trait DebugConditionT: Sync + Send + Fn(&WorldState) -> bool {}
```
pretentious7, line 114: please let your eyes glaze over this whole song and dance. it's just a workaround for the fact that Fn doesn't implement debug.

dev/tests/scheduler.rs
```diff
@@ -20,20 +22,24 @@ use plane_dev::{
...
     util::{base_scheduler_request, random_loopback_ip},
 };
-use std::{net::IpAddr, time::Duration};
-use tokio::time::sleep;
+use std::net::IpAddr;
+use tokio::sync::Mutex;
 
 pub const CLUSTER_DOMAIN: &str = "plane.test";
 const PLANE_VERSION: &str = env!("CARGO_PKG_VERSION");
+lazy_static! {
```
pretentious7, line 30: I was trying to move more stuff in here, but was having difficulties. open to removing this entirely.

dev/tests/scheduler.rs
```diff
@@ -1,4 +1,5 @@
 use anyhow::Result;
+//use std::sync::Mutex;
```
paulgb, line 2: just remove if not used

dev/Cargo.toml
```diff
@@ -32,3 +32,4 @@ tokio-tungstenite = "0.19.0"
 url = "2.0.0"
 trust-dns-resolver = "0.22.0"
 trust-dns-server = "0.22.0"
+lazy_static = "1.4.0"
```
paulgb, line 35: Remove if not needed

dev/Cargo.toml
```diff
@@ -8,7 +8,7 @@ publish = false
 anyhow = "1.0.62"
 async-nats = { git = "https://github.com/nats-io/nats.rs.git", rev = "261ead4" }
 bollard = "0.14.0"
-chrono = { version="0.4.22", default_features = false }
+chrono = { version="0.4.22", default-features = false }
```
paulgb, line 11: good catch :)

core/src/state/world_state.rs
```diff
@@ -65,44 +66,94 @@ impl StateHandle {
...
         let cluster = self.clusters.entry(message.cluster.clone()).or_default();
         cluster.apply(message.message);
         self.logical_time = sequence;
 
-        if let Some(sender) = self.listeners.remove(&sequence) {
-            sender.notify_waiters();
-        }
+        let mut listeners = self.listeners.write().unwrap();
+        *listeners = {
+            let b = listeners.split_off(&(sequence + 1));
```
paulgb, line 151: Good find. Too bad there's no `split_off_reverse` or this could be quite elegant! I think you've found the best option, I'd only suggest renaming `b` something like `outstanding_listeners`.

core/src/state/world_state.rs
```diff
@@ -65,44 +66,94 @@ impl StateHandle {
...
-    pub fn get_listener(&mut self, sequence: u64) -> Arc<Notify> {
-        match self.listeners.entry(sequence) {
-            Entry::Occupied(entry) => entry.get().clone(),
-            Entry::Vacant(entry) => {
-                let notify = Arc::new(Notify::new());
+    pub fn logical_time(&self) -> u64 {
+        self.logical_time
+    }
+
+    pub fn get_listener(&self, sequence: u64) -> ClosableNotify {
```
paulgb, line 125: Couldn't we simplify this and eliminate the need for `CloseableNotify` with something like:

```
pub async fn wait_for_seq(&self, sequence: u64) {
    if self.logical_time >= sequence {
        return;
    }
    match self.listeners.write().unwrap().entry(sequence) {
        Entry::Occupied(entry) => entry.get().subscribe().recv().await,
        Entry::Vacant(entry) => {
            let (send, recv) = broadcast::channel(1024);
            entry.insert(send);
            recv.recv().await
        }
    }
}
```

`self.listeners` would be a `BTreeMap<u64, Sender<()>>` and instead of calling `notify_waiters()` we would just call `.send(())` directly.

paulgb (commented):
Looks good, question about simplifying.

core/src/state/world_state.rs
pretentious7: And pass around sequence numbers to wait for instead of the notifies themselves? I guess that works.

The big reason for having closable notify is that it's clone, which makes it useful for the locking use case. Since I can shove it in a map and wait on it when needed.

pretentious7: Hmm, on second thought, we really can't do the wait_for_seq thing the way you suggest since that'll require holding a lock to the statehandle, which won't work at all, unless I'm missing something. 

One thing I could do is to just use the normal notify if we want to keep complexity down in this code, and just have a doc comment telling people it's possible for it to deadlock if you don't await it in time.

dev/Cargo.toml
pretentious7: Not me, cargo told me when I removed lazy_static. (Now it didn't tell me when I added lazy_static, so not sure ehats up there)

core/src/state/world_state.rs
pretentious7: actually found a way to use a thinner wrapper around normal notify

paulgb: I had a go at it, let me know what you think of #424

pretentious7: that looks good! For some reason I just wasn't thinking of future as just another trait. I've cherry-picked that commit onto this pr.

core/src/state/world_state.rs
```diff
@@ -65,44 +68,55 @@ impl StateHandle {
...
-        }
+        };
+
+        Box::pin(async move {
+            recv.recv().await.unwrap();
+        })
     }
 
     pub fn apply(&mut self, message: WorldStateMessage, sequence: u64) {
+        tracing::info!(?message, ?sequence, "World state message!");
```
paulgb, line 107: this will be noisy, rm

core/src/state/world_state.rs
```diff
@@ -65,44 +68,55 @@ impl StateHandle {
...
-        receiver.notified().await;
-    }
 }
 
 #[derive(Default, Debug)]
 pub struct WorldState {
     logical_time: u64,
     pub clusters: BTreeMap<ClusterName, ClusterState>,
-    listeners: HashMap<u64, Arc<Notify>>,
+    listeners: RwLock<BTreeMap<u64, Sender<()>>>,
```
paulgb, line 77: I think we discussed this offline early in the design, but given where things landed, does it still make sense to have double-nested `RwLock`s? The way I see it, the "owned" value is the `StateHandle`, which already has `WorldState` in an `RwLock`. If we expose `StateHandle::write_state`, we could leave this as a `HashMap` and simplify the code a bit.

paulgb (commented):
Looking good! A couple comments but otherwise this is almost ready.

core/src/state/world_state.rs
pretentious7: I agree, but I don't like exposing write_state like that tbh, I've put wait_for_seq on StateHandle, which I think is better.

paulgb (approved):

PR #423, opened 2023-06-30
add FetchBackendForLock message

pretentious7:


controller/src/lib.rs
```diff
@@ -168,19 +170,63 @@ async fn dispatch_schedule_request(
...
+        &lock_request.value.lock,
+    ) {
+        if let Ok(schedule_response) = schedule_response_for_existing_backend(
+            &state,
+            lock_request.value.cluster.clone(),
+            backend,
+        ) {
+            response = schedule_response.try_into().expect(
+                "already checked that backend exists, \\
+						 so schedule_response should be convertible to FetchBackendForLockResponse",
```
paulgb, line 190: we found another place that rustfmt misses tabs

although in this case it might be legit -- I think these tabs are actually part of the string, since they don't match the indentation of the previous line (don't quote me on that, I forget the exact rules)

paulgb (approved):
One issue around whitespace, otherwise LGTM

controller/src/lib.rs
```diff
@@ -168,19 +170,63 @@ async fn dispatch_schedule_request(
...
+            Some(schedule_request) = schedule_request_sub.next() => {
+                tracing::info!(metadata=?schedule_request.value.metadata.clone(), "Got spawn request");
+                tokio::spawn(dispatch_schedule_request(
+                    state.clone(),
+                    schedule_request.clone(),
+                    scheduler.clone(),
+                    nats.clone(),
+                ));
+            },
+            Some(locked_backend_req) = locked_backend_request_sub.next() => {
```
rolyatmax, line 220: Probably worth updating the "locked backend" language in this block, too

PR #424, opened 2023-07-02
Paul's suggestions on top of Abhi's remove sleeps PR

paulgb:
I suggest this as a simplification of the “remove sleeps” PR. The premise is still fundamentally the same, but I've made a few changes that eliminate the need for `CloseableNotify`. We still get a send + sync handle that we can pass around, namely a pinned and boxed future, which is nice because we can pass it to things that expect a future (like `timeout`) without having to convert it to a future first.

The fundamental problem with `Notify`, as I understand it, is that it is not attached to a queue or memory, so whether a listener gets notified depends on the order of events. Broadcast channels are a bit more heavyweight because they have a queue, but they are robust to different orderings.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -99,15 +99,15 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.1"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


pretentious7:
incorporated into https://github.com/drifting-in-space/plane/pull/422

thanks!

PR #426, opened 2023-07-30
Add heartbeat to world state

paulgb:
This PR turns `WorldStateMessage` into an enum and adds a `Heartbeat` variant. It is similar to the approach taken in #425 but different in two ways:

- The schema of existing WorldStateMessages is unchanged, and a test is added to ensure JSON-level schema stability. This is accomplished by using an [untagged](https://serde.rs/enum-representations.html#untagged) enum representation.
- The heartbeat message itself doesn't include the time; instead, the timestamp of the NATS message is used. This will be the timestamp on the NATS RAFT leader at the time the message was acknowledged.

This PR does not implement the heartbeat loop itself or any of the locking stuff -- the goal of this PR is simply to add a heartbeat message type while preserving a stable schema.


core/src/state/world_state.rs
```diff
@@ -100,16 +101,24 @@ pub struct WorldState {
     logical_time: u64,
     pub clusters: BTreeMap<ClusterName, ClusterState>,
     listeners: BTreeMap<u64, Sender<()>>,
+    last_heartbeat: Option<OffsetDateTime>,
```
pretentious7, line 104: hmmm, why do we use both time and chrono?

we should pick one, ok with this for now though.

core/src/messages/state.rs
```diff
@@ -8,26 +8,36 @@ use serde::{Deserialize, Serialize};
 use std::net::IpAddr;
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
-pub struct WorldStateMessage {
-    pub cluster: ClusterName,
-
-    #[serde(flatten)]
-    pub message: ClusterStateMessage,
+#[serde(untagged)]
```
pretentious7, line 11: ooo v nice.

core/src/messages/state.rs
```diff
@@ -8,26 +8,36 @@ use serde::{Deserialize, Serialize};
...
+pub enum WorldStateMessage {
+    ClusterMessage {
+        cluster: ClusterName,
+        #[serde(flatten)]
+        message: ClusterStateMessage,
+    },
+    Heartbeat {
+        heartbeat: (),
```
pretentious7, line 19: maybe have this be a struct along the lines of clustermessage, so
{
message: HeartbeatMessage
}
with 
enum HeartbeatMessage {
None,
}

so that we can change it later if needed

pretentious7, line 19: as per offline convo, make it so this serializes as {}

pretentious7 (changes requested):
just the heartbeat change needed to serialize as {}, then we can merge this, I'll rebase ack-lock on top of it.

paulgb:
Changed to make it future-compatible with a heartbeat message that has a payload. In the spirit of YAGNI, I just made it optional, so it still encodes as `null` but won't complain if a non-`null` is passed.

core/src/state/world_state.rs
paulgb: I agree, but in this case we get an `OffsetDateTime` and don't have a need to convert it yet, so let's defer for now. `time` is a dependency via `async_nats` either way, and the value we get back is an `OffsetDateTime` either way.

core/src/messages/state.rs
paulgb: I kept it as `null`, but made it future-compatible with a non-null payload.

pretentious7:
lgmt

pretentious7 (approved):

PR #429, opened 2023-08-08
Temporarily ignore assignment message subject mismatches

paulgb:
The subject for assignment messages was renamed from `state.cluster.{}.backend.{}` to `state.cluster.{}.backend.{}.assignment` for consistency with `state.cluster.{}.backend.{}.lock`.

This allows both message types to be used for an interim period, so that deploying still captures existing backends on all subjects. After this has been live for long enough for existing drones to be cycled out, we can roll this back.

### DIFF
--- a/core/src/nats.rs
+++ b/core/src/nats.rs
@@ -124,7 +124,9 @@ fn parse_and_verify_message<T: TypedMessage>(message: &Message) -> Result<T> {
     let value: T = serde_json::from_slice(&message.payload)?;
     let expected_subject = value.subject();
     if expected_subject != message.subject {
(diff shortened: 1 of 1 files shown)


pretentious7 (approved):

pretentious7:
good catch

paulgb:
Link to where the subject changed: https://github.com/drifting-in-space/plane/commit/065ff21471a70d485cd8e064bea35af1c964df43#r123954618

rolyatmax:
LGTM - so in a day or so we'll undo this change, right? Is that as simple as creating a revert of this commit? Maybe we can go ahead and create that PR so we don't forget.

rolyatmax:
The PR to revert this in a couple days: https://github.com/drifting-in-space/plane/pull/430

PR #433, opened 2023-08-10
schema for world state messages from serde::Serialize

pretentious7:
the idea here is to have a small canary of sorts that notifies us if the schema's changed and how, I thought about using one of the existing libraries but they all involve adding tags and stuff and I didn't like that. This just abuses the Serialize trait.

the idea is that we'd have a plain text world_state_schema.txt (checked into VC) file of the forme:
```
subject:
{<<schema>>}
next_subject:
{<<schema>>}
```
and a test that fails if the current schema is different.

I've only implemented the serializer thing thus far, would like some thoughts before continuing

Of course, in the future one can imagine all sorts of cool macro stuff too, but I just want to have some warning of failure with preexisting jetstream for now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -902,6 +902,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "gen-schema-map"
+version = "0.1.0"
+dependencies = [
+ "plane-core",
+ "serde",
+]
+
+[[package]]
 name = "generic-array"
 version = "0.14.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,6 +5,7 @@ members = [
     "controller",
     "core",
     "core/nats-macros",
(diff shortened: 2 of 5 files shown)


paulgb:
Clever implementation, but I worry that this is over-engineering things a bit -- I'm trying to figure out what this solves for us in practice that running `plane-cli dump-state` on an existing NATS stream does not. I also worry that it won't capture subtle changes like a value becoming optional, or a non-optional value becoming optional, because those can result in the same set of calls to the serializer (depending on the value being serialized).

It looks like we will have to construct specific examples of each message we want to use this for, so instead of testing that their schema matches, why not just test if the underlying JSON for those specific messages still deserializes to the same message?

I do think it could be useful to generate a descriptive schema from the NATS types so that when we make schema changes, it appears in a .schema file, but I think we would need a separate macro for that rather than piggybacking on serde as a serializer so that we can get the output without having to instantiate the type.

PR #434, opened 2023-08-17
Add HTTP API to Plane controller

paulgb:
This implements an HTTP API that is a subset of the NATS API. It aims for compatibility with jamsocket.com, so that jamsocket users can test locally with Plane.

### DIFF
--- a/controller/src/http_server.rs
+++ b/controller/src/http_server.rs
@@ -1,0 +1,167 @@
+use crate::config::HttpOptions;
+use anyhow::Result;
+use axum::{
+    extract::{Path, State},
+    http::StatusCode,
+    routing::{get, post},
+    Json, Router,
+};
+use plane_core::{
+    messages::{
+        agent::{DockerExecutableConfig, DockerPullPolicy, ResourceLimits},
+        scheduler::{ScheduleRequest, ScheduleResponse},
+    },
+    nats::TypedNats,
+    types::{ClusterName, ResourceLock},
+    Never,
+};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Duration};
+
+trait LogError<T> {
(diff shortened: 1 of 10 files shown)


controller/src/http_server.rs
```diff
@@ -0,0 +1,268 @@
...
+    volume_mounts: Vec<Value>,
+}
+
+impl HttpSpawnRequest {
+    pub fn into_spawn_request(
+        &self,
+        cluster: ClusterName,
+        image: String,
+    ) -> Result<ScheduleRequest> {
+        let max_idle_secs = Duration::from_secs(self.grace_period_seconds.unwrap_or_else(|| 600));
```
pretentious7, line 66: unwrap_or should be fine here, also would be good to put this in a const at the the top of the file.

controller/src/http_server.rs
```diff
@@ -0,0 +1,268 @@
...
+    port: Option<u16>,
+    #[serde(default = "HashMap::default")]
+    env: HashMap<String, String>,
+    #[serde(default = "Vec::default")]
+    volume_mounts: Vec<Value>,
+}
+
+impl HttpSpawnRequest {
+    pub fn into_spawn_request(
+        &self,
```
pretentious7, line 62: either call this as_* or silence the clippy complaint by turning off wrong_self_convention

controller/src/http_server.rs
```diff
@@ -0,0 +1,268 @@
...
+async fn handle_spawn(
+    State(server_state): State<Arc<ServerState>>,
+    TypedHeader(host): TypedHeader<Host>,
+    Path(service): Path<String>,
+    Json(request): Json<HttpSpawnRequest>,
+) -> Result<Json<HttpSpawnResponse>, StatusCode> {
+    let service = server_state
+        .services
+        .get(&service)
+        .ok_or_else(|| StatusCode::NOT_FOUND)?;
```
pretentious7, line 117: should be fine to use ok_or here too

pretentious7 (approved):
Some clippy stuff, looks great otherwise!

PR #437, opened 2023-08-23
lock names were not being hashed in certain message subjects

pretentious7:


paulgb:
Oof good catch, what are the implications of this?

paulgb (approved):

PR #438, opened 2023-08-24
Path-based routing

paulgb:
This changes the routing behavior of the proxy to facilitate local development without `Host` rewrites.

If the URL of a proxied request begins with `/_plane_backend=[...]`, the value between that prefix and the next `/`, or the end of the string, is used as the name of the backend. Any `Host` parameter is then ignored for that request.

This is intended to be a localhost development feature for advanced users. I am not adding tests in this PR because:
- it's an undocumented feature mostly for supporting our own testing, so if it breaks we can fix it as needed
- it would take an overhaul of our test infrastructure, which makes assumptions based on the DNS-based routing approach

I considered putting this behind a flag so that it is only available for local development, but I do not see a need for this. If clients are connecting directly to a drone, they can craft whatever `Host` header they want; this simply allows a client to pass a path that supersedes the `Host` sent by that same client.

### DIFF
--- a/drone/src/proxy/service.rs
+++ b/drone/src/proxy/service.rs
@@ -256,84 +256,93 @@ impl ProxyService {
         }
     }
 
-    async fn handle(self, mut req: Request<Body>) -> anyhow::Result<Response<Body>> {
+    fn backend_from_request(&self, req: &Request<Body>) -> Result<String> {
+        let mut path = req.uri().path();
+
+        if path.starts_with("/_plane_backend=") {
+            let backend = path.strip_prefix("/_plane_backend=").unwrap();
+            let (backend, new_path) = backend.split_once('/').unwrap_or_else(|| (backend, ""));
+            path = new_path;
+            return Ok(backend.to_string());
+        }
+
         if let Some(host) = req.headers().get(http::header::HOST) {
             let host = std::str::from_utf8(host.as_bytes())?;
-            // If the host includes a port, strip it.
-            let host = host.split_once(':').map(|(host, _)| host).unwrap_or(host);
+            let host = host.rsplit_once(':').map(|(host, _)| host).unwrap_or(host);
+            let Some(subdomain) = host.strip_suffix(&format!(".{}", self.cluster)) else {
+                return Err(anyhow!("Host does not end with cluster domain."));
+            };
+            return Ok(subdomain.to_owned());
+        }
 
-            tracing::info!(ip=%self.remote_ip, url=%req.uri(), "Proxy Request");
+        Err(anyhow!("No backend found in host or URL path."))
+    }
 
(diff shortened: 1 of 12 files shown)


rolyatmax:
I feel like this feature should be opt-in as it adds complexity to parts of the system that must be easy to understand in order to make good decisions re: security. This addition breaks some assumptions we could previously make from inside the Jamsocket backend - assumptions around how the request was routed there (this is often used to determine if a request has already passed through an auth proxy) and which origins the response will be readable from. It may be tempting for devs to rely on the path prefix without realizing it opens them up to vulnerabilities - especially in a multi-tenant situation. Or it may be possible for a malicious script loaded from `victim.jamsocket.run/_plane_backend=attacker` to read cookies/responses from `victim.jamsocket.run`. I don't know how likely any of these are, but I think wherever we can make it easier to reason about security, we should do so. Thoughts?

drone/src/proxy/service.rs
```diff
@@ -256,84 +256,102 @@ impl ProxyService {
...
+
+        tracing::info!(ip=%self.remote_ip, url=%req.uri(), "Proxy Request");
+        let route = self.db.get_proxy_route(&subdomain).await?.or_else(|| {
+            self.passthrough.map(|d| ProxyRoute {
+                address: d.to_string(),
+                bearer_token: None,
+            })
+        });
+
+        if req.uri().path() == "/_plane_auth" {
```
rolyatmax, line 324: Will this work with the `/_plane_backend=` prefix because `path` was mutated in `backend_from_request`? (if so, does `req.uri().path()` return that same mutated path?) And would this code break if the `handle_plane_auth()` step was moved to the top of this function?

Maybe `backend_from_request` should return the backend _and_ the stripped path?

paulgb: Good catch, I've made `backend_from_request` modify the URL in place. I don't love it as a side-effect, but since it's just for a dev feature I'm ok with it. I'd like to refactor the whole thing at some point.

paulgb:
> I feel like this feature should be opt-in as it adds complexity to parts of the system that must remain easy to understand in order to make good decisions re: security. This addition breaks some assumptions we could previously make from inside the Jamsocket backend - assumptions around how the request was routed there (this is often used to determine if a request has already passed through an auth proxy) and which origins the response will be readable from. It may be tempting for devs to rely on the path prefix without realizing it opens them up to vulnerabilities - especially in a multi-tenant situation. Or it may be possible for a malicious script loaded from `victim.jamsocket.run/_plane_backend=attacker` to read cookies/responses from `victim.jamsocket.run`. I don't know how likely any of these are, but I think wherever we can make it easier to reason about security, we should do so. Thoughts?

Yeah, I've come around to this. Implemented it as a flag.

rolyatmax (approved):
One last little comment there, otherwise, LGTM

PR #440, opened 2023-08-29
drone: String -> Vec<String> in plane-cli

pretentious7:
Shouldn't break existing behaviour, just allow specifying multiple drones

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -443,18 +443,15 @@ async fn main() -> Result<()> {
         } => {
             let drain = !cancel;
             for ref drone in drone {
-                nats.request(&DrainDrone {
-                    cluster: ClusterName::new(&cluster),
-                    drone: DroneId::new(drone.clone()),
-                    drain,
-                })
-                .await
-                .map_or_else(
-                    |e| {
-                        println!("{} {}", "Draining failed on drone: ".bright_red(), drone);
-                        println!("Error: {}", e.to_string().bright_red())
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -440,12 +440,14 @@ async fn main() -> Result<()> {
...
-                drain,
-            })
-            .await?;
+            for drone in drone {
+                nats.request(&DrainDrone {
+                    cluster: ClusterName::new(&cluster),
+                    drone: DroneId::new(drone),
+                    drain,
+                })
+                .await?;
```
paulgb, line 449: Instead of aborting on the first error, let's just report the error and move on to the next drone. That way, if your connection is dropped mid-way through draining a bunch of drones, you can just run the same command again.

pretentious7:
I do wish the help reflected the fact that there can be multiple drones.

cli/src/main.rs
```diff
@@ -440,19 +442,30 @@ async fn main() -> Result<()> {
...
                 })
-                .await?;
-            }
-
-            if drain {
-                println!("{}", "Draining started on drone.".bright_green());
-            } else {
-                println!("{}", "Draining cancelled on drone.".bright_green());
+                .await
+                .map_or_else(
```
paulgb, line 452: the `map` here makes it harder to follow. I would avoid `map*` when the result is not actually needed.

paulgb (approved):

PR #446, opened 2023-09-06
Set PORT from HTTP spawn

### LINKED ISSUE #445: [Bug] environment variable `PORT` not set automatically by plane
In the [jamsocket docs](https://docs.jamsocket.com/api-docs/#backend-containers) it describes

> All backends will have the following reserved env vars passed to them. Env vars with these names passed by the user when spawning will be overridden by the platform-provided values.
> 
> * PORT
> * JAMSOCKET_NAME
> * JAMSOCKET_URL
> * JAMSOCKET_SERVICE

My team is using plane.dev to emulate jamsocket locally and expected this behavior at least for `PORT` to be identical but it is not. Plane's controller/drone do not automatically apply the env variable PORT to containers.

To have more parity between Jamsocket and local plane.dev deployments for BYOC customers this would be nice to address.

Currently we have modified our internal jamsocket HTTP client to detect when we're running locally and to automatically add `env: { PORT: request.port ?? '8080' }` to our spawn requests but this is a little brittle since it may diverge from jamsocket/plane's behavior.

paulgb:
Fixes #445

### DIFF
--- a/controller/src/http_server.rs
+++ b/controller/src/http_server.rs
@@ -69,7 +69,7 @@ impl HttpSpawnRequest {
         let lock: Option<ResourceLock> = self.lock.clone().map(|l| l.try_into()).transpose()?;
 
         let mut env = self.env.clone();
-        env.insert("PORT".to_string(), "8080".to_string());
+        env.insert("PORT".to_string(), self.port.unwrap_or(8080).to_string());
 
         let executable = DockerExecutableConfig {
             image,
--- a/controller/src/http_server.rs
+++ b/controller/src/http_server.rs
@@ -68,9 +68,12 @@ impl HttpSpawnRequest {
 
         let lock: Option<ResourceLock> = self.lock.clone().map(|l| l.try_into()).transpose()?;
 
+        let mut env = self.env.clone();
+        env.insert("PORT".to_string(), "8080".to_string());
+
         let executable = DockerExecutableConfig {
             image,
-            env: self.env.clone(),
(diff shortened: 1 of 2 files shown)


controller/src/http_server.rs
```diff
@@ -68,9 +68,12 @@ impl HttpSpawnRequest {
 
         let lock: Option<ResourceLock> = self.lock.clone().map(|l| l.try_into()).transpose()?;
 
+        let mut env = self.env.clone();
+        env.insert("PORT".to_string(), "8080".to_string());
```
HunterLarco, line 72: Shouldn't this be `self.port` and if that's not defined, 8080? As written now, if I spawn a backend with port 4000, PORT environment variable will be set to 8080

paulgb: Good catch.

sample-config/dev-config/controller.toml
```diff
@@ -10,4 +10,4 @@ domain_name = "ns1.example.com."
 
 [http]
 cluster = "plane.test"
-services.hello-world = "ghcr.io/drifting-in-space/demo-image-hello-world:sha-3b58532"
+services.hello-world = "ghcr.io/drifting-in-space/demo-image-hello-world:latest"
```
paulgb, line 13: This is only used for local (manual) testing so I'm fine not pinning it. I updated the image to display `PORT` in order to test this change.

pretentious7 (approved):

PR #450, opened 2023-09-26
limit backend_log to 50gb by default and add option to change it

pretentious7:
this is a reasonable initial limit. also adds log_stream_size_limit_bytes option in controller config to adjust it.


paulgb:
Since we've done this change manually on our end, and have a logging rewrite on the roadmap, I think we can close this now.

PR #455, opened 2023-09-29
Version bump to 0.3.24 + add error logging for host header parsing

rolyatmax:


drone/src/proxy/service.rs
```diff
@@ -296,8 +296,9 @@ impl ProxyService {
 
         if let Some(host) = req.headers().get(http::header::HOST) {
             let host = std::str::from_utf8(host.as_bytes())?;
-            let host = host.rsplit_once(':').map(|(host, _)| host).unwrap_or(host);
-            let Some(subdomain) = host.strip_suffix(&format!(".{}", self.cluster)) else {
+            let host_trimmed = host.rsplit_once(':').map(|(host, _)| host).unwrap_or(host);
```
paulgb, line 299: while we're at it, we could add `.to_lowercase()`. I've never seen an uppercase host, but technically the RFC allows it.

paulgb (approved):

PR #461, opened 2023-10-13
Abhi/fix db error

pretentious7:


paulgb:
NB #462 will fix the build failures, if you want to approve that and rebase on top of it

paulgb (approved):
LGTM, thanks

drone/src/agent/executor.rs
```diff
@@ -336,11 +336,18 @@ impl<E: Engine> Executor<E> {
...
+                        _ => {
+                            tracing::error!(
+                                ?error,
+                                ?state,
+                                "Error unhandled (putting backend into failed state"
+                            );
+
+                            // leads to .step() running with failed.
+                            state = BackendState::Failed;
+                            self.update_backend_state(spawn_request, state).await;
```
rolyatmax, line 348: Do we know for sure at this point that the container is no longer running?

rolyatmax: Ahhh, gotcha gotcha - I see that now 👍

PR #467, opened 2023-10-23
Call out that Plane only works on linux

paulgb:


paulgb:
Preview:

image

https://plane-git-paulgb-callout-linux-drifting-corp.vercel.app/docs/getting-started

pretentious7 (approved):

PR #473, opened 2023-10-26
stream changes

paulgb:
- Rename `DroneLogMessage` to more accurate `BackendLogMessage`
- Log an error (but move on) if we fail to create a stream


rolyatmax (approved):
LGTM!

PR #474, opened 2023-10-27
Add mark-backend-lost command to mark backends as lost [DIS-1294]

pretentious7:
This command doesn't do any validation, simply publishes lost messages for the backendids supplied on the command line. I do not think that this command should access the world state to check that the backend exists because it'll slow down the command.

I considered adding an AbandonRequest style thing with something in the controller, but decided against it since it would be strictly more work for no real benefit.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -94,9 +94,6 @@ enum Command {
 
         #[arg(required = true)]
         backends: Vec<String>,
-
-        #[clap(short, long, default_value = "false")]
-        force: bool,
     },
 }
 
@@ -479,11 +476,7 @@ async fn main() -> Result<()> {
                 }
             }
         }
-        Command::MarkBackendLost {
-            cluster,
-            backends,
-            force,
-        } => {
+        Command::MarkBackendLost { cluster, backends } => {
             let stdin = std::io::stdin();
             let state = get_world_state(nats.clone()).await?;
             let cluster = ClusterName::new(&cluster);
@@ -509,14 +502,12 @@ async fn main() -> Result<()> {
                 if let Some(ref lock) = backend_state.lock {
                     println!("NOTE: backend holds the following lock: {lock}!");
                 }
-                if !force {
-                    println!("are you sure you want to mark this backend as lost?");
-                    println!("(type y and end line to continue, any other input will skip marking this backend as lost)");
-                    let mut confirmation = String::new();
-                    stdin.read_line(&mut confirmation)?;
-                    if confirmation != "y" {
-                        continue;
-                    }
+                println!("are you sure you want to mark this backend as lost?");
+                println!("(type y and end line to continue, any other input will skip marking this backend as lost)");
+                let mut confirmation = String::new();
+                stdin.read_line(&mut confirmation)?;
+                if confirmation != "y" {
+                    continue;
                 }
                 let lost_state_message =
                     BackendStateMessage::new(BackendState::Lost, cluster.clone(), backend);
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -89,7 +89,7 @@ enum Command {
         dry_run: bool,
     },
     /// marks backend(s) as lost.
-    Abandon {
+    MarkBackendLost {
         cluster: String,
 
         #[arg(required = true)]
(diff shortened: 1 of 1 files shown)


cli/src/main.rs
```diff
@@ -88,6 +88,13 @@ enum Command {
         #[clap(long, default_value = "false")]
         dry_run: bool,
     },
+    /// marks backend(s) as lost.
+    Abandon {
```
rolyatmax, line 92: Can we name this something like `MarkLost` or `MarkBackendLost`, for clarity?

pretentious7: yessir.

cli/src/main.rs
```diff
@@ -469,6 +476,16 @@ async fn main() -> Result<()> {
                 }
             }
         }
+        Command::Abandon { cluster, backends } => {
+            let cluster = ClusterName::new(&cluster);
+            for backend in backends {
```
rolyatmax, line 481: Given that this command is essentially a "trust me you can release this lock" command, I would prefer to put in a few safeguards here. I think the main one is that we should make sure we're only setting backends to Lost that aren't already in a terminal state. Doing that check first will also ensure that the backend definitely exists. Secondly, having a prompt that shows you some of the backend info and then asks you to confirm has been useful in the past. That's what the [set-backend-lost](https://github.com/drifting-in-space/p2/blob/main/jamsocket/src/admin/cli.rs#L266-L318) command in the admin CLI does right now, and this command should be able to replace that one.

cli/src/main.rs
```diff
@@ -469,6 +476,16 @@ async fn main() -> Result<()> {
...
+                let backend = BackendId::new(backend);
+                let lost_state_message =
+                    BackendStateMessage::new(BackendState::Lost, cluster.clone(), backend);
+
+                nats.publish_jetstream(&lost_state_message).await?;
```
rolyatmax, line 486: Can we confirm that publishing this message will result in the platform updating the postgres DB with the new Lost state?

pretentious7: Hmm I can confirm that it gets rid of locks, but I didn't check what it would do to the platform, I'll run it locally. iirc the platform listens to the drones for statuses right? in that case it wouldn't

rolyatmax: Actually, I think we might be good - looks like p2 listens to the `BackendStateMessage` subject - https://github.com/drifting-in-space/p2/blob/main/jamsocket/src/platform/mod.rs#L102C29-L102C29

cli/src/main.rs
```diff
@@ -476,10 +479,45 @@ async fn main() -> Result<()> {
...
+                println!("{backend_state:?}");
+                if backend_state
+                    .state()
+                    .map_or(false, |state| state.terminal())
+                {
+                    println!(
+                        "backend state {} is a terminal state!",
+                        backend_state.state().unwrap()
+                    );
+                }
```
rolyatmax, line 508: If the backend is already in a terminal state, then I think we should `continue` and not allow pushing a Lost state to it. The constraint that a backend can only have one terminal state in its history of states is a useful one and is trivial to enforce here.

cli/src/main.rs
```diff
@@ -476,10 +479,45 @@ async fn main() -> Result<()> {
...
+                if !force {
+                    println!("are you sure you want to mark this backend as lost?");
+                    println!("(type y and end line to continue, any other input will skip marking this backend as lost)");
+                    let mut confirmation = String::new();
+                    stdin.read_line(&mut confirmation)?;
+                    if confirmation != "y" {
+                        continue;
+                    }
+                }
```
rolyatmax, line 520: Can we add a message at the end of each loop (and in the confirmation block) confirming what was done ("backend marked as lost" or "skipping backend, not marking as lost" - something like that?)

cli/src/main.rs
```diff
@@ -94,6 +94,9 @@ enum Command {
 
         #[arg(required = true)]
         backends: Vec<String>,
+
+        #[clap(short, long, default_value = "false")]
+        force: bool,
```
rolyatmax, line 99: I'm a little iffy on whether we should include this option at all. It makes it easy to bypass a safeguard - and for not much benefit. I've been setting hundreds of backends to Lost for many months now, and there's no `--force` option there and it hasn't been a problem to just press 'y' a bunch of times.

rolyatmax:
@pretentious7 A few final comments there. Let me know what you think!

rolyatmax (approved):
Looks great! Thanks for those changes! Let's make sure tests pass, then I think we're good to merge

PR #475, opened 2023-11-07
turn one_shot off for docker stats

pretentious7:
the docker.stats() call fails with stream: false, one_shot: true for docker 24.0.7 (this is because the response for those options now returns an object that does not have the name or id fields set, which bollard considers to be required)

this should be reverted once
https://github.com/fussybeaver/bollard/pull/347 has landed.


rolyatmax (approved):
LGTM! Thank you!

PR #476, opened 2023-11-15
add unsubscribe and queue group support to TypedNats

rolyatmax:
This adds unsubscribe and queue group support to TypedNats.

The queue group functionality is necessary for situations where you have more than one process handling messages on a subject (whether for robustness or for availability during zero-downtime deploys), but you want to make sure you only send a given message to a single process.

The unsubscribe allows us to gracefully shut down subscriptions before terminating a subscribe loop.

Ideally, we'd have a `drain()` method on TypedNats, but it doesn't come out-of-the-box in `async_nats` the way it does in `nats`. Adding `unsubscribe()` to the `TypedSubscription` is a good start anyway.


pretentious7 (approved):
Lgtm

PR #491, opened 2023-12-15
DIS-1449 add resource limits to executor config and spawn opts

pretentious7:


plane2/src/drone/docker/commands.rs
```diff
@@ -136,6 +136,32 @@ pub async fn run_container(
         host_config: Some(HostConfig {
             port_bindings: Some(create_port_bindings()),
             runtime: runtime.clone(),
+            memory: exec_config.resource_limits.memory_limit_bytes,
+            cpu_period: exec_config
+                .resource_limits
+                .cpu_period
+                .as_ref()
+                .map(|cpu_period| std::time::Duration::from(cpu_period).as_micros() as i64),
```
paulgb, line 144: Do you know what happens if `cpu_quota` is `Some(_)` but `cpu_period` is `None`?

If Docker ignores it, I would rather that we use the default period so that someone who supplies `cpu_quota` gets the behavior they expect.

I am assuming that `cpu_period` has no effect if `cpu_quota` is not provided (I can't find confirmation in the docker docs). If that's _not_ the case, we probably want to use the default only if `cpu_quota` is provided.

paulgb (commented):
Looks good! A couple minor comments.

plane2/src/drone/docker/commands.rs
```diff
@@ -136,6 +136,32 @@ pub async fn run_container(
...
+            cpu_quota: exec_config
+                .resource_limits
+                .cpu_quota()
+                .map(|cpu_quota| cpu_quota.as_micros() as i64),
+            ulimits: exec_config
+                .resource_limits
+                .cpu_time_limit
+                .map(|cpu_time_limit| {
+                    let mins = (cpu_time_limit.as_secs_f64() / 60.0).floor() as i64;
```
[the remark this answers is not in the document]
pretentious7: I pulled that from the earlier code I think, but it was wrong

image


thanks for the catch.

plane2/src/drone/docker/commands.rs
pretentious7: good question

after doing

`docker run --cpu-quota=10000 -it --rm bash`

I get:

image


so yea it just uses the default, but in anticipation of surprising behaviour I agree we should just pass in the default anyhow.

pretentious7: actually, probably less surprising if we just leave it

paulgb: I agree with leaving the default behaviour to Docker. Since it uses the Docker default if it is `None`, can we get rid of `DockerCpuPeriod::default` and use `None` instead?

pretentious7: That is what it's doing now. Or do you mean get rid of the default impl? I use that to calculate quota from the relative value if the period isn't specified.

paulgb: ahh right, nevermind

paulgb (approved):
LGTM!

PR #492, opened 2023-12-15
clusterId -> clusterName, validate cluster name, make https based on port

rolyatmax:
See commit messages for descriptions

### DIFF
--- a/plane2/src/types.rs
+++ b/plane2/src/types.rs
@@ -5,7 +5,7 @@ use crate::{
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_json::{Map, Value};
-use std::{collections::HashMap, fmt::Display, net::SocketAddr};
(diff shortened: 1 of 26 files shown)


plane2/src/types.rs
```diff
@@ -53,23 +53,36 @@ impl BackendKeyId {
...
 }
 
-impl From<String> for ClusterId {
-    fn from(s: String) -> Self {
-        Self(s)
+impl FromStr for ClusterName {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s = s.trim();
```
paulgb, line 75: I am tempted to say that we should not trim here, because whitespace ending up in here seems like it would be a symptom of other problems. Thoughts?

paulgb (approved):
Looks good, one comment.

plane2/src/types.rs
rolyatmax: Yeah, I think you're right. Updated!

PR #498, opened 2023-12-17
Heartbeat refactor to make way for keys

paulgb:
The initial heartbeat implementation in plane2 was mostly carried over from plane1. plane1 needed heartbeats because we didn't know when a drone connected/disconnected to NATS. In plane2, nodes talk directly to a controller, so we can piggyback on the TCP connection status to determine whether a node is connected or not.

For the purpose of keys, we do still periodically want the drone to send an opaque timestamp to the controller. This will be used by the controller to initially allocate the key lease when starting a drone (to avoid needing to do a round-trip to the drone). The key lease part isn't implemented yet, but rather than getting rid of the existing heartbeat code only to add similar code later, this PR repurposes the existing heartbeat code to keep the timestamp up to date on the controller.

This PR also gets rid of the `NodeStatus` type. This type was supposed to be a generalization of drone status to other nodes, but it ended up not being useful to have. Instead, a `ready` field is added to the `drone` table.

A drone is considered available for spawning if:
- It has `ready = true`
- Its last heartbeat is within 45 seconds
- It has a non-null controller (i.e. it is currently connected to a controller)
- The controller is not marked as offline
- The controller it is connected to has had a heartbeat in the last 45 seconds
- It is not being drained

Note that `ready` is a separate concept from `draining` -- `ready` is a property of the drone itself (e.g. whether it has a connection to Docker), while `draining` is a scheduler concept that the drone is not aware of at all. Currently, a drone is automatically marked ready when it connects to a controller.


plane2/schema/derived_schema.sql
```diff
@@ -253,7 +253,10 @@ ALTER TABLE public.controller OWNER TO postgres;
 
 CREATE TABLE public.drone (
     id integer NOT NULL,
-    draining boolean DEFAULT false NOT NULL
+    ready boolean NOT NULL,
+    draining boolean DEFAULT false NOT NULL,
+    last_heartbeat timestamp with time zone,
+    last_local_epoch_millis bigint
```
paulgb, line 259: Even though this represents a time, to the controller it is just an opaque number of milliseconds -- treating it as a clock time in the controller would be a bug.

plane2/src/database/node.rs
```diff
@@ -193,26 +194,10 @@ pub struct NodeRow {
...
-        if self.controller.is_none() {
-            return false;
-        }
-
-        if self.status_age().num_seconds() > HEARTBEAT_INTERVAL_SECONDS {
-            return false;
-        }
-
-        self.last_status.is_active()
+        !self.controller.is_none()
```
pretentious7, line 201: clippy suggestion here is ok (! is_none -> is_some)

pretentious7:
>     * Its last heartbeat is within 45 seconds
>     * The controller it is connected to has had a heartbeat in the last 45 seconds

Just to restate for my understanding, the controller will spawn only if it has received a heartbeat from the drone in the last 45 seconds and if that heartbeat has a timestamp within the last 45 seconds?

paulgb:
> > ```
> > * Its last heartbeat is within 45 seconds
> > * The controller it is connected to has had a heartbeat in the last 45 seconds
> > ```
> 
> Just to restate for my understanding, the controller will spawn only if it has received a heartbeat from the drone in the last 45 seconds and if that heartbeat has a timestamp within the last 45 seconds?

The second one refers to the heartbeat of the controller itself. Controllers bump a timestamp in the `controller` table every 30 seconds to tell other controllers that they are still alive. The situation we are trying to avoid is that a controller dies, leaving the connected drones online. This is probably unnecessary because the lack of heartbeat would mark the drone offline as well, but it keeps the behaviour consistent with how we will do it elsewhere in the codebase.

rolyatmax (approved):

PR #502, opened 2023-12-19
Simplify typed connection implementation

paulgb:
Currently, there is a different type used for client and server connections. This leads to some duplication of code between client and server side code, as well as functionality gaps between the two (for example, it is currently possible to get a cloneable "sender" type on the client side but not the server side.)

This refactors the connection code so that both the client and server side code return the same type, a `TypedSocket`. The `TypedSocket` holds a sender/receiver pair that sends and receives messages from the websocket connection. A background task is started, which runs in the background until the connection is closed.


plane2/src/typed_socket/mod.rs
```diff
@@ -2,19 +2,61 @@ use crate::PlaneVersionInfo;
...
+}
+
 pub trait ChannelMessage: Send + Sync + 'static + DeserializeOwned + Serialize + Debug {
     type Reply: ChannelMessage<Reply = Self>;
 }
 
-#[async_trait::async_trait]
-pub trait FullDuplexChannel<T: ChannelMessage> {
-    async fn send(&mut self, message: &T) -> anyhow::Result<()>;
+pub struct TypedSocket<T: ChannelMessage> {
```
paulgb, line 19: `TypedSocket` now represents the handle to a connection whether it is a client or server connection. It replaces both the old `TypedSocket` and `TypedSocketServer`, and obviates the need for the `FullDuplexChannel` trait which they both inhereted.

pretentious7 (approved):
so the idea here is that the ws is hooked up to a mpsc to make it easier to send to from a bunch of places right? lgtm

paulgb:
> so the idea here is that the ws is hooked up to a mpsc to make it easier to send to from a bunch of places right? lgtm

Yeah, and it simplifies some other things because it allows the client and server code to be the same.

PR #503, opened 2023-12-19
DIS-1448 Add credentials for custom registry use

pretentious7:


plane2/src/types.rs
```diff
@@ -222,7 +223,7 @@ pub enum PullPolicy {
 pub struct ExecutorConfig {
     pub image: String,
     pub pull_policy: PullPolicy,
-
+    pub credentials: Option<DockerCredentials>,
```
paulgb, line 226: Let's use our own enum type for credentials rather than the bollard type, for two reasons:
- It gives us a more Rusty way of expressing required fields.
- I don't like using structs from dependencies in our public API, because it means that our API can break without us knowing if those deps change under us.

For now I think username/pass is good enough:

```rust
enum DockerRegistryAuth {
    UsernamePassword {
        username: String,
        password: String
    }
}
```

paulgb (approved):
Looks great, thanks!

PR #505, opened 2023-12-20
Implement keys

paulgb:
Implementation notes: https://plane-t2lp-9w6wrngdv-drifting-corp.vercel.app/developing/keys

- Schema changes:
  - Changes `backend_key` table so that instead of having an ID, we just re-use the backend ID of the backend it is associated with. This enforces the fact that backends have at most one key associated with them.
  - Changes `last_renewed` to `expires_at`. This conforms to a general principle that we should push computation (in this case calculating the expiration time) into writers rather than readers, since the less computation readers do the less likely they are to behave differently if multiple different versions are running concurrently.
  - Adds `fencing_token` to `backend_key`
  - Change `last_local_epoch_millis` field to `last_local_time` on `drones`, which simplifies time arithmetic because we don't have to think about units.

- CLI changes:
  - adds `prevent-renew` command to `dbcli` which prevents a specific backend from renewing its key.

- Drone changes:
  - adds a `KeyManager`, which is started for the drone process and manages a mapping of backend ID to a handle that is responsible for ensuring that that backend has an active key, and terminating it if it doesn't.


plane/schema/migrations/20231027092522_initial-schema.sql
```diff
@@ -91,17 +91,20 @@ create index idx_backend_action_pending on backend_action(drone_id) where acked_
...
-    last_renewed timestamptz not null,
-    backend_id varchar(255) references backend(id) not null
+    expires_at timestamptz not null,
+    fencing_token bigint not null,
+    allow_renew bool not null default true
 );
 
 create unique index idx_cluster_namespace_name on backend_key(cluster, namespace, key_name);
 
+comment on column backend_key.allow_renew is 'If false, the key cannot be renewed with the same fencing token.';
```
rolyatmax, line 106: nit: should we add comments for `fencing_token` and `expires_at`? And also update the comment for the `id` column?

plane/schema/migrations/20231027092522_initial-schema.sql
```diff
@@ -91,17 +91,20 @@ create index idx_backend_action_pending on backend_action(drone_id) where acked_
 create index idx_backend_action_backend on backend_action(backend_id);
 
 create table backend_key (
-    id serial primary key,
+    id varchar(255) references backend(id) not null,
```
rolyatmax, line 94: Do we need to designate this as a primary key? Is there any uniqueness constraint right now that makes sure a backend only has one lock?

plane/schema/migrations/20231027092522_initial-schema.sql
paulgb: Yes, I'll make that another PR though.

plane/src/database/backend_key.rs
```diff
@@ -16,40 +21,63 @@ impl<'a> KeysDatabase<'a> {
...
     /// Remove a key, ensuring that it is still expired.
     /// Returns Ok(true) if the key was successfully removed.
-    pub async fn remove_key(&self, id: BackendKeyId) -> Result<bool, sqlx::Error> {
+    pub async fn remove_key(&self, backend: BackendName) -> Result<bool, sqlx::Error> {
         let result = sqlx::query!(
             r#"
             delete from backend_key
             where id = $1
-            and last_renewed - now() > $2
+            and expires_at > now()
```
rolyatmax, line 48: Maybe a dumb question: why do we want to prevent deleting keys that have expired?

plane/src/database/backend_key.rs
```diff
@@ -105,22 +135,23 @@ pub enum BackendKeyHealth {
...
+    pub token: i64,
+    pub key: String,
+    expires_at: DateTime<Utc>,
     as_of: DateTime<Utc>,
 }
 
 impl BackendKeyResult {
     pub fn key_health(&self) -> BackendKeyHealth {
-        let key_age = (self.as_of - self.last_renewed).num_seconds();
+        let key_age = (self.as_of - self.expires_at).num_seconds();
```
rolyatmax, line 148: Is this logic correct? Doesn't `as_of - expires_at` give you the amount of time since the key expired?

plane/src/drone/mod.rs
```diff
@@ -82,6 +110,20 @@ pub async fn drone_loop(
...
+                    let RenewKeyResponse { backend, deadlines } = renew_key_response;
+                    tracing::info!(?deadlines, "Received key renewal response.");
+
+                    if let Some(deadlines) = deadlines {
+                        key_manager
+                            .lock()
+                            .expect("Key manager lock poisoned.")
+                            .update_deadlines(&backend, deadlines);
+                    } else {
+                        // TODO: we could begin the graceful termiation here.
```
rolyatmax, line 123: nit: misspelling

rolyatmax: Also, we need to implement this before the implementation is actually correct, right?

plane/src/database/backend_key.rs
paulgb: not a dumb question; it's a bug! Should be the other way around.

plane/src/drone/mod.rs
rolyatmax: Oh, nevermind, I see that this happens in the renew key loop

rolyatmax:
Few comments there, but otherwise looks good! Nice work! Are there any tests for the key renewal logic?

plane/src/database/backend_key.rs
```diff
@@ -1,12 +1,17 @@
...
+    types::{ClusterName, KeyConfig},
 };
 use chrono::{DateTime, Utc};
 use sqlx::{postgres::types::PgInterval, PgPool};
 use std::time::Duration;
 
+pub const KEY_LEASE_RENEW_AFTER: Duration = Duration::from_secs(30);
+pub const KEY_LEASE_SOFT_TERMINATE_AFTER: Duration = Duration::from_secs(40);
+pub const KEY_LEASE_HARD_TERMINATE_AFTER: Duration = Duration::from_secs(50);
+pub const KEY_LEASE_EXPIRATION: Duration = Duration::from_secs(60);
```
rolyatmax, line 13: Elsewhere we have [tests that enforce invariants](https://github.com/drifting-in-space/Plane/blob/c61e5708fa925430bb15b056d0142fc7f4112f9c/plane/src/heartbeat_consts.rs#L1-L13) about durations like these. I wonder if it would be useful to have something like that here since the renewal loop logic breaks if any of these values isn't greater than the previous one.

plane/src/database/backend_key.rs
paulgb: fixed.

plane/src/database/backend_key.rs
paulgb: Good idea, done. Also it turns out `cargo nextest` does not run doctests, so I've added a command to the workflow to run these.

paulgb:
> Few comments there, but otherwise looks good! Nice work! Also, are there any tests for the key renewal logic?

There are no tests yet -- I'm deferring them until I figure out a good way to do them since they run in realtime and testing a full key lifecycle would be slow. I'm tempted to make all the timeout parameters configurable to make it work, but that would be a bit of a refactor.

rolyatmax (approved):

PR #507, opened 2023-12-20
Abhi/collect metrics

pretentious7:


plane2/src/drone/backend_manager.rs
```diff
@@ -222,6 +251,31 @@ impl BackendManager {
...
+        handle.take();
+
+        let docker = self.docker.clone();
+        let id = self.container_id.clone();
+        let self_clone = self.clone();
+
+        *handle = Some(GuardHandle::new(async move {
+            while {
+                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+                true
```
paulgb, line 270: using [interval](https://docs.rs/tokio/latest/tokio/time/fn.interval.html) should give slightly more uniform time-steps

plane2/src/drone/backend_manager.rs
```diff
@@ -38,17 +41,36 @@ impl StepStatusResult {
     }
 }
 
-type StateCallback = Box<dyn Fn(&BackendState) -> Result<(), Box<dyn Error>> + Send + Sync>;
+struct Callback<T>(Box<dyn Fn(T) -> Result<(), Box<dyn Error>> + Send + Sync>);
```
paulgb, line 44: Is there an advantage to using a newtype here over a type alias? Is it that you wanted Debug implemented?

plane2/src/drone/backend_manager.rs
```diff
@@ -59,7 +81,10 @@ pub struct BackendManager {
     executor_config: ExecutorConfig,
 
     /// Function to call when the state changes.
-    state_callback: StateCallback,
+    state_callback: Callback<BackendState>,
+
+    /// Function to call when metrics produced.
+    metrics_callback: Callback<Stats>,
```
paulgb, line 87: Given that this callback will likely just pass the data over the socket, should we just use a `TypedSender` for this? This PR gave me an idea for improving the ergonomics of `TypedSender` for this purpose, which I opened here: https://github.com/drifting-in-space/plane/pull/508

plane2/src/drone/backend_manager.rs
pretentious7: So i can stick tracing::instrument on things to debug.

pretentious7:
superseded by https://github.com/drifting-in-space/plane/pull/539

PR #508, opened 2023-12-21
TypedSender alteration

paulgb:
From a `TypedSocket` connection, we have a way to create a `TypedSocketSender` which lets us send messages from elsewhere in the program. The problem is, the `TypedSocketSender` still takes messages of the top-level protocol type, so e.g. the heartbeat sender needs to construct a `MessageToDrone` even though it only wants to send messages of the `MessageToDrone::Heartbeat` variant.

This PR changes `TypedSocketSender` so that it holds a boxed callback for converting an input into the type that needs to be sent on the wire. For example, instead of the old way:

```rust
let sender = conn.sender();
sender.send(MessageFromDrone::Heartbeat(Heartbeat {...}));
```

We now have

```rust
let sender = conn.sender(MessageFromDrone::Heartbeat);
sender.send(Heartbeat {...});
```

### DIFF
--- a/plane2/src/typed_socket/mod.rs
+++ b/plane2/src/typed_socket/mod.rs
@@ -22,18 +22,18 @@ pub struct TypedSocket<T: ChannelMessage> {
     pub remote_handshake: Handshake,
 }
 
-pub struct TypedSocketSender<T: ChannelMessage> {
-    send: Sender<SocketAction<T>>,
+pub struct TypedSocketSender<A> {
+    inner_send: Box<dyn Fn(SocketAction<A>) -> anyhow::Result<()> + 'static + Send + Sync>,
 }
 
-impl<T: ChannelMessage> TypedSocketSender<T> {
-    pub fn send(&self, message: T) -> anyhow::Result<()> {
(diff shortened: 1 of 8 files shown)


plane2/src/typed_socket/mod.rs
```diff
@@ -22,18 +22,18 @@ pub struct TypedSocket<T: ChannelMessage> {
     pub remote_handshake: Handshake,
 }
 
-pub struct TypedSocketSender<T: ChannelMessage> {
-    send: Sender<SocketAction<T>>,
+pub struct TypedSocketSender<A> {
+    inner_send: Box<dyn Fn(SocketAction<A>) -> anyhow::Result<()> + 'static + Send + Sync>,
```
pretentious7, line 26: maybe have the error here as a parameter (like A, T: std::error::Error)

or a Box<dyn Error>

but this is ok for now.

pretentious7 (approved):
LGTM!

I think we should have better error types, but that can wait.

plane2/src/typed_socket/mod.rs
paulgb: Done! I added a concrete error type.

PR #509, opened 2023-12-21
Add debug to BackendManager

pretentious7:


pretentious7:
basically the only things to review is [[this][https://github.com/drifting-in-space/plane/pull/509/commits/87d64857eb5938d8f835e1fc77b068ccf2715787]] commit

paulgb:
I don't love adding another layer of indirection just to be able to derive `Debug` (I know it's a zero-cost abstraction, but this is more about a layer of indirection when reading the code).

I made an alternate suggestion in #510, what do you think of that approach? I assume that he fields other than backend ID and container ID do not debug-print useful info.

paulgb:
#510 is merged, mind if I close this?

PR #512, opened 2023-12-22
Misc plane2 cleanup

paulgb:
Misc cleanup, focusing on the Drone in particular:

- More consistency with state vs. status
- Adds some comments
- Remove some unwraps
- Remove some redundant duration conversions
- Simplify `CertManager` data flow

### DIFF
--- a/plane2/src/typed_socket/server.rs
+++ b/plane2/src/typed_socket/server.rs
@@ -14,13 +14,13 @@ pub async fn handle_messages<T: ChannelMessage>(
             Some(msg) = messages_to_send.recv() => {
                 match msg {
                     SocketAction::Send(msg) => {
-                        let msg = Message::Text(serde_json::to_string(&msg).unwrap());
+                        let msg = Message::Text(serde_json::to_string(&msg).expect("Always serializable."));
                         if let Err(err) = socket.send(msg).await {
                             tracing::error!(?err, "Failed to send message on websocket.");
                         }
                     }
                     SocketAction::Close => {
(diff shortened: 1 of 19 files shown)


plane2/src/typed_socket/server.rs
```diff
@@ -14,13 +14,13 @@ pub async fn handle_messages<T: ChannelMessage>(
...
                     SocketAction::Send(msg) => {
-                        let msg = Message::Text(serde_json::to_string(&msg).unwrap());
+                        let msg = Message::Text(serde_json::to_string(&msg).expect("Always serializable."));
                         if let Err(err) = socket.send(msg).await {
                             tracing::error!(?err, "Failed to send message on websocket.");
                         }
                     }
                     SocketAction::Close => {
-                        socket.close().await.unwrap();
+                        let _ = socket.close().await;
```
rolyatmax, line 23: Do we expect this to fail during normal operation, or would a failure here be a useful indicator of something else having gone wrong?

rolyatmax (approved):
LGTM - one question there

plane2/src/typed_socket/server.rs
paulgb: It shouldn't cause anything to break if it fails (since the connection will get closed either way), but I suppose it can't hurt to log since we don't expect to encounter it.

paulgb:
> LGTM - one question there
> 
> Also, looks like there's a failing test - is that expected?

I'm not sure why they failed in that run, but they pass now!

PR #513, opened 2023-12-28
Scaffold Plane2 docs

paulgb:
Preview link: https://plane-t2lp-hn7nn0fdu-drifting-corp.vercel.app/

PR #514, opened 2024-01-02
rename to plane

paulgb:


rolyatmax (approved):
LGTM - do we need to update Vercel to make deployments work with this change?

paulgb:
> LGTM - do we need to update Vercel to make deployments work with this change?

This change doesn't impact the vercel deployments, but vercel deployments are broken on all branches except #513 until it is merged because it is configured for the /docs directory, which does not exist on other branches.

PR #517, opened 2024-01-02
Move cli into main plane binary

paulgb:
This moves the cli from a separate binary to the `plane admin` subcommand. The main advantage of this is that it allows us to ship one docker image that can be used both to run controllers/nodes/proxies and also to run admin commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ The `dev/cli.sh` script runs the Plane CLI, configured to connect to the local P
 
 ```bash
 dev/cli.sh connect \
+    --wait \
     --cluster 'localhost:9090' \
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -66,6 +66,7 @@ The `dev/cli.sh` script runs the Plane CLI, configured to connect to the local P
 
 ```bash
 dev/cli.sh connect \
+    --wait \
```
pretentious7, line 69: What's this do?

pretentious7 (approved):

README.md
paulgb: After issuing the initial request, it opens an event stream, and prints events. The command only exits when the backend is ready.

PR #519, opened 2024-01-02
Abhi/dylib speedup tests

pretentious7:
This creates a sub crate for tests, and has that crate use a plane 2 dylib. The only thing that really changes for users is now you've to run `cargo test --all` to run the integration tests.

recompiling tests after adding a println or something now takes 10 seconds! which is still a bit long, but it's a lot better than crashing my computer.

(as for why the linking was so slow, the debug libplane rlib is 134 mb, the linkers operations on this  in both space and time are apparently ~O(N), for the tests we did that for each test, and since it had to be relinked with every tiny change...)


Cargo.toml
```diff
@@ -2,5 +2,7 @@
 resolver = "2"
 members = [
     "plane",
-    "plane/tests/plane-test-macro",
+	"plane/plane-dynamic",
```
paulgb, line 5: nit: spaces not tabs

plane/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "plane"
 version = "0.2.0"
 edition = "2021"
+autotests = false
```
paulgb, line 5: why was this necessary?

paulgb (approved):
this is great! Left a couple comments, otherwise should be good to go.

plane/Cargo.toml
pretentious7: otherwise it looks for tests in /tests, but actually since we aren't providing another target...

pretentious7: ah that's right it's cause I was calling plane-tests tests earlier and it would complain

pretentious7: thanks

paulgb: but we no longer have a `/tests` directory at that level, right?

paulgb (approved):

PR #524, opened 2024-01-03
use docker for cli commands

paulgb:
Previously, following the tutorial required an install of rust/cargo and would build the CLI. This uses Docker for that part instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ The Plane Controller will be available at http://localhost:8080, and the Plane P
 
 ### Connecting to a process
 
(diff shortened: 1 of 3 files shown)


pretentious7:
might not be the worst idea to invest some time in packaging too. Would be quite low effort now, with vendoring, since the only non rust deps are docker, postgres, and openssl.

pretentious7 (approved):

paulgb:
> might not be the worst idea to invest some time in packaging too. Would be quite low effort now, with vendoring, since the only non rust deps are docker, postgres, and openssl.

let's talk about this today

Issue #291, opened 2022-12-26
Plane should return `410 Gone` rather than `404 Not Found` when a backend is shut down.


paulgb:
This is a request from the discord.

PR #525, opened 2024-01-03
Return FORBIDDEN on unknown connection token

paulgb:
Returns `FORBIDDEN` rather than `NOT FOUND` when a connection token is not found. This is more accurate now that we use connection tokens as a sort of ephemeral bearer token. This does not differentiate between a token that has expired and one that never existed, but both cases are reasonable uses of `FORBIDDEN`.

This addresses #291, where the underlying issue was that it was hard to diagnose a 404 since it could either come from Plane or the backend itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,10 +63,10 @@ The Plane Controller will be available at http://localhost:8080, and the Plane P
 
 ### Connecting to a process
 
-The `dev/cli.sh` script runs the Plane CLI, configured to connect to the local Plane Controller.
+The `docker/cli.sh` script runs the Plane CLI, configured to connect to the local Plane Controller.
 
 ```bash
-dev/cli.sh connect \
+docker/cli.sh connect \
     --wait \
     --cluster 'localhost:9090' \
     --image ghcr.io/drifting-in-space/demo-image-drop-four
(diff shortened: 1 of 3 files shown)


pretentious7 (approved):

pretentious7:
> Returns `FORBIDDEN` rather than `NOT FOUND` when a connection token is not found. This is more accurate now that we use connection tokens as a sort of ephemeral bearer token. This does not differentiate between a token that has expired and one that never existed, but both cases are reasonable uses of `FORBIDDEN`.

Why not use GONE in the former case?

paulgb:
> > Returns `FORBIDDEN` rather than `NOT FOUND` when a connection token is not found. This is more accurate now that we use connection tokens as a sort of ephemeral bearer token. This does not differentiate between a token that has expired and one that never existed, but both cases are reasonable uses of `FORBIDDEN`.
> 
> Why not use GONE in the former case?

A few reasons:
- we may not always be able to tell the difference -- connection tokens will eventually be purged from the database
- the code path a client is expected to take (request another token) is the same in both cases, so it simplifies things to have them return the same output
- if a connection token has expired it should always be FORBIDDEN even if the backend is gone, because we don't want to leak the fact that the backend is gone

PR #530, opened 2024-01-04
Add quickstart dockerfile

paulgb:
This adds a Dockerfile for a "quickstart image" which runs Postgres and a Plane drone, controller, and proxy all in one container.

The advantage is that someone can get Plane up and running without installing anything (except Docker) by running this command:

```bash
docker run plane/quickstart \
    -p 8080:8080 \
    -p 9090:9090 \
    -v /var/run/docker.sock:/var/run/docker.sock
```

However, they still need to have the CLI installed to use it (or use a clunky docker command for each cli call).

This is something of an experiment, I'm tbd on whether it's a good idea, but it does work. This PR does not set up an automatic build/publish or add any docs.


docker/quickstart/Dockerfile
```diff
@@ -0,0 +1,54 @@
+# =======
+# Builder
+# =======
+FROM debian:bookworm as builder
+
+# We can't just use a rust image because we need a version of libssl that is compatible
```
pretentious7, line 6: using rust:bookworm with postgres:bookworm doesn't work?

pretentious7 (approved):

docker/quickstart/Dockerfile
paulgb: ooh good call I didn't realize rust had a bookworm image

PR #533, opened 2024-01-04
Final 'initial migration' changes

paulgb:
Once we deploy this in staging, we will only be able to do DB changes by creating new migrations, so I wanted to take this last chance to clean up the migrations.

This PR:
- Adds a bunch of comments.
- Removes `cluster` from `backend_key`. I couldn't decide what the "right" behavior is for Plane, but every behavior I could think of is accomplishable through some combination of namespace/tag.


plane/schema/migrations/20231027092522_initial-schema.sql
```diff
@@ -8,6 +8,14 @@ create table controller (
     ip inet not null
 );
 
+comment on table controller is 'Self-reported information about controllers.';
+comment on column controller.first_seen is 'The first time the controller came online.';
+comment on column controller.last_heartbeat is 'The last time the controller sent a heartbeat.';
+comment on column controller.is_online is 'Whether the controller is online (self-reported; if a controller dies suddenly this will not be updated).';
+comment on column controller.plane_version is 'The version of the plane running on the controller.';
```
rolyatmax, line 15: nit: "the version of plane..."

rolyatmax (approved):
Small wording nit, otherwise LGTM!

PR #534, opened 2024-01-05
Abhi/add resource limit test

pretentious7:


plane/src/drone/docker/commands.rs
```diff
@@ -187,3 +195,139 @@ pub async fn run_container(
...
+    }
+
+    #[tokio::test]
+    async fn test_resource_limits() -> anyhow::Result<()> {
+        let docker = bollard::Docker::connect_with_local_defaults()?;
+        let backend_name = BackendName::new_random();
+        let _container_id = ContainerId::from(format!("plane-test-{}", backend_name));
+        let mut executor_config = ExecutorConfig::from_image_with_defaults("debian:bookworm");
+
+        let resource_limits: ResourceLimits = serde_json::from_str(
```
paulgb, line 265: nit: I prefer `serde_json::from_value(json! {` because it allows the compiler to enforce that the syntax is well-formed JSON. Not a huge deal here since it's a test, but it's nice to use it everywhere for consistency.

plane/src/drone/docker/commands.rs
```diff
@@ -187,3 +195,139 @@ pub async fn run_container(
...
+            Some(t) if matches!(t.as_str(), "btrfs" | "zfs") => {}
+            Some(t)
+                if t == "overlay2" && {
+                    let fs_ok;
+                    if let Some(status) = info.driver_status {
+                        fs_ok = !status.iter().flatten().any(|s| s.contains("xfs"));
+                    } else {
+                        fs_ok = false
+                    };
+                    !fs_ok
```
paulgb, line 289: I find the double-negation hard to file; consider omitting `fs_ok` entirely and implicitly returning the whole block:

```rust
if info.driver_status {
    status.iter().flatten().any(|s| s.contains("xfs"))
} else {
    true
}
```

plane/src/drone/docker/commands.rs
```diff
@@ -187,3 +195,139 @@ pub async fn run_container(
...
+        config.cmd = Some(vec!["cat".into(), "/sys/fs/cgroup/memory.max".into()]);
+        let out = run_container_with_config(config.clone()).await?;
+        assert_eq!(out.trim(), mem_limit_corr.to_string());
+
+        //unfortunately, the docker disk limit is dependent on the storage backend
+        //hence just validating here that config is as expected.
+        config.host_config.unwrap().storage_opt.map(|hm| {
+            let size = hm.get("size").cloned();
+            assert_eq!(size, Some("10000000000".to_string()));
+        });
```
paulgb, line 329: If I'm reading this right, it doesn't catch the case where `get_container_config_from_executor_config` should set `storage_opt` to a `Some(_)` but doesn't.

Instead, consider:

```rust
if let Some(expected_size) = executor_config.resource_limits.disk_limit_bytes {
    assert_eq!(
        expected_size.to_string(),
        config.host_config
            .unwrap()
            .storage_opt
            .unwrap()
            .get("size")
            .unwrap()
    );
}
```

paulgb (commented):
looking good, left some comments. Is the intention for #535 to be merged before this?

plane/src/drone/docker/commands.rs
pretentious7: oh cool i didn't realise that macro tokenized like that. sweet.

paulgb (approved):

PR #535, opened 2024-01-05
better de/serialization for resourcelimits

pretentious7:


plane/src/types.rs
```diff
@@ -206,15 +210,17 @@ impl From<&DockerCpuPeriod> for std::time::Duration {
...
 pub struct ResourceLimits {
-    /// Period of cpu time
+    /// Period of cpu time (de/serializes as microseconds)
     pub cpu_period: Option<DockerCpuPeriod>,
 
     /// Proportion of period used by container
     pub cpu_period_percent: Option<u8>,
 
     /// Total cpu time allocated to container
+    #[serde_as(as = "Option<serde_with::DurationSeconds<u64>>")]
```
paulgb, line 223: it feels weird having two time units on the JSON representation side. Thoughts on using a floating-point seconds representation for both?

pretentious7: hmmm, this is how docker takes it, that's how we'd justified it last time.

Not fp, too much stuff close to 0. but we could use nanos or micros consistently, wdyt?

paulgb: What's a typical value for `DockerCpuPeriod`? I thought it would be millis at least, but maybe I misunderstand it.

I am ok straying from docker's way of doing things here, the docker API has a lot of vestigial stuff that I'm happy to paper over where we can.

pretentious7: 100 milliseconds, but it reports in microseconds (https://docs.kernel.org/admin-guide/cgroup-v2.html#cgroup-v2-cpu)

you generally only go under that when you've got like super short lived loops and spinlocks and such.

paulgb (approved):
thanks for the clarifications, lgtm

PR #536, opened 2024-01-05
use git root as scratch dir base for tests

pretentious7:


paulgb (approved):
I like this! Just to confirm, this is just an ergonomics thing, it's not to fix anything, right?

We should also move test-scratch into the .gitignore in root (I think it is already in the .gitignore in plane/)

pretentious7:
> I like this! Just to confirm, this is just an ergonomics thing, it's not to fix anything, right?
> 
> We should also move test-scratch into the .gitignore in root (I think it is already in the .gitignore in plane/)

It is in the root, so since I move the plane-tests in one more level, the artifact uploader has been broken, this fixes that.

PR #538, opened 2024-01-09
Allow bearer tokens in client, not just typed socket

paulgb:
Plane can be configured to use a bearer token by sending a connection string like `https://my-bearer-token@my-plane-controller.dev`. Previously, this was implemented in `TypedSocket` and so it only worked for WebSocket connections, not regular HTTP GET/POST requests.

This moves the logic into the client, allowing us to use the same code path for both regular and WebSocket requests.

This also removes base64-encoding from the conversion step. base64-encoding is typically used in `Basic` auth, but not `Bearer` auth.

### DIFF
--- a/plane/src/client/mod.rs
+++ b/plane/src/client/mod.rs
@@ -1,5 +1,6 @@
+use self::controller_address::AuthorizedAddress;
 use crate::{
-    controller::error::ApiError,
+    controller::{error::ApiError, StatusResponse},
     names::{BackendName, DroneName},
     protocol::{MessageFromDns, MessageFromDrone, MessageFromProxy},
     typed_socket::client::TypedSocketConnector,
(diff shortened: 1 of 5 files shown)


pretentious7:
and this bearer token is used for access to the controller's control endpoints, correct?

plane/src/client/mod.rs
```diff
@@ -208,3 +179,45 @@ impl PlaneClient {
...
+}
+
+async fn authed_post<T: DeserializeOwned>(
+    client: &reqwest::Client,
+    addr: &AuthorizedAddress,
+    body: &impl serde::Serialize,
+) -> Result<T, PlaneClientError> {
+    let mut req = client.post(addr.url.clone());
+    if let Some(header) = addr.bearer_header() {
+        req = req.header("Authorization", header);
```
pretentious7, line 218: if the token is in the url, why also set the header?

plane/src/typed_socket/client.rs
```diff
@@ -88,15 +88,14 @@ impl<T: ChannelMessage> TypedSocketConnector<T> {
 
 /// Turns a URL (which may have a token) into a request. If the URL contains a token, it is
```
pretentious7, line 89: this comment is no longer accurate right? since the token is removed at the AuthorizedUrl::From stage?

paulgb: good catch, thanks!

PR #540, opened 2024-01-09
adds tracing spans to tracing_subscriber capture

pretentious7:
This is useful because it means the #[tracing::instrument] decoration on some thing immediately shows up in logs.

Otherwise it only shows up alongside a tracing event (info!, debug! etc)


paulgb (approved):
cool, sgtm

PR #543, opened 2024-01-10
Abhi/container stats to backend metrics

pretentious7:


paulgb (approved):
Nice, I appreciate the thorough tests.

plane/src/drone/docker/mod.rs
```diff
@@ -166,6 +168,112 @@ impl PlaneDocker {
...
+
+    Ok(BackendMetricsMessage {
+        backend_id,
+        mem_total,
+        mem_active,
+        mem_inactive,
+        mem_unevictable,
+        mem_available,
+        cpu_used: container_cpu_used_delta,
+        sys_cpu: system_cpu_used_delta,
```
rolyatmax, line 273: These messages previously had a `mem_used` value along with drone and cluster info. Is there a reason why we left those out? https://github.com/drifting-in-space/plane/blob/plane-0.3.x/core/src/messages/agent.rs#L168

pretentious7: mem_total is intended to be used in the place of mem_used (i could rename it), i thought the new categories were clearer

pretentious7: hmmm, that's true, I changed it because I wanted correct figures for scheduling (the docker figure for mem_used makes no sense, considering that if one is going to not consider page cache, why would one not also remove inactive pages?

but might be good to retain the mem_used field separately to match docker stats?

rolyatmax: I think we should match the docker stats output as much as possible - inconsistencies between what docker outputs and what plane streams will lead to distrust, I'm afraid. Furthermore, this metric is very important for folks who are trying to set memory limits - so if Docker uses `mem_used` when killing containers, we should surface that number. (Not sure if it does, though?) Lastly, do we really need to use backend metrics for scheduling? Don't we want to use the _drone's_ metrics for that?

pretentious7: 1. docker does not use mem_used for killing containers, that's up to the OOMKiller, which uses figures based on the page allocators free lists (the active/ inactive stuff). 
2. Yes, you're correct, to first order the only thing that would matter to the scheduler here would be the mem_limit (of course, if we want to set priorities then it will matter)

As for distrust, the formulae in the docs don't even match what docker stats comes up with some of the time (indeed they can't since a lot of them are cgv1 specific and still wrong)  , but I guess it's nice to have someone else to blame.

pretentious7: so, I think it depends, if we want to to surface data to help people with debugging and monitoring,  I'd say (total, active, inactive, available) is good (probably should also have kernel, but then cgv1 issues, alongside a best effort mem_used to match docker.

otherwise just mem_used is fine too of course, noone's complaining about docker's stats output (I'd guess for the most part perf sensitivity is orthogonal to docker use), wdyt?

pretentious7: as for cluster, the backend state messages used to have that, and it doesn't now, so I figured I'd keep it consistent unless it came to be important while implementing the controller pub/sub side somehow.

paulgb: I think there may ultimately be two sets of numbers with different goals: a number displayed to users, and a number used for scheduling.

The number displayed to users should match the number reported by the Docker CLI to the extent we can -- I realize that their calculations can differ from what's documented. The number should ideally also correlate with the total memory available reported by the system: if I kill a backend with 1GB of memory reported, I should ideally see approximately 1GB of memory returned to the system.

Re. the cluster information, not including it is consistent with what we do for other drone<>controller messages. If we include it here, it's one more thing to check to confirm that a drone is not forging the cluster, so I am on board with leaving it out unless you foresee other issues.

PR #544, opened 2024-01-10
Abhi/metrics manager

pretentious7:


plane/src/drone/backend_manager.rs
```diff
@@ -84,12 +152,19 @@ impl BackendManager {
         state: BackendState,
         docker: PlaneDocker,
         state_callback: impl Fn(&BackendState) -> Result<(), Box<dyn Error>> + Send + Sync + 'static,
+        metrics_sender: TypedSocketSender<BackendMetricsMessage>,
```
paulgb, line 155: the problem with this is that the `TypedSocketSender` is not durable across reconnects.

One alternative here is to pass the `Arc<StateStore>`, and then call `state_store.get_metrics_sender()` to get the sender for each message at send-time.

paulgb (commented):
I like the approach. One correctness issue but I think with the way you've structured things it should be an easy fix (see comment).

plane/src/drone/docker/mod.rs
```diff
@@ -166,6 +168,112 @@ impl PlaneDocker {
...
+
+    Ok(BackendMetricsMessage {
+        backend_id,
+        mem_total,
+        mem_active,
+        mem_inactive,
+        mem_unevictable,
+        mem_available,
+        cpu_used: container_cpu_used_delta,
+        sys_cpu: system_cpu_used_delta,
```
rolyatmax, line 273: These messages previously had a `mem_used` value along with drone and cluster info. Is there a reason why we left those out? https://github.com/drifting-in-space/plane/blob/plane-0.3.x/core/src/messages/agent.rs#L168

pretentious7: sorry, based this pr wrong, discussion here: https://github.com/drifting-in-space/plane/pull/543

plane/src/drone/backend_manager.rs
pretentious7: I've used the same metrics_sender as the one in state_store with an arc, does this work?

pretentious7: oh wait a sec, it does not.

plane/src/drone/backend_manager.rs
```diff
@@ -40,6 +42,73 @@ impl StepStatusResult {
...
+                    ) else {
+                        continue;
+                    };
+                    let Ok(_) = sender
+                        .read()
+                        .expect("failed to get read lock on metrics sender!")
+                        .send(metrics_message)
+                    else {
+                        continue;
+                    };
```
paulgb, line 103: this has no effect, I think?

pretentious7: ayup!

paulgb (approved):

PR #545, opened 2024-01-10
Some debug derivations and implementations

pretentious7:
This is just to be able to stick tracing::instrument in places, it's especially useful to be able to do this on TypedSocketSender::send


plane/src/typed_socket/mod.rs
```diff
@@ -47,7 +53,7 @@ impl<A> From<TrySendError<A>> for TypedSocketError {
     }
 }
 
-impl<A> TypedSocketSender<A> {
+impl<A: Debug> TypedSocketSender<A> {
```
paulgb, line 56: I think this isn't necessary because you've derived debug manually?

pretentious7: It is to instrument send.

paulgb (approved):

PR #547, opened 2024-01-11
Support suffix for DNS server

paulgb:
This supports an optional DNS zone that can be provided to the DNS server. If a zone is provided, it will be stripped off of an incoming request before it is processed.

For example, if the server is started with `--zone example.com`, a request for `TXT` records on `_acme-challenge.foo.bar.baz.example.com` would be treated as a request for TXT records of the cluster `foo.bar.baz`.

If a zone is provided, requests that are not under that zone will fail.

This is useful because it allows us to use CNAME delegation for `_acme-challenge` subdomains.

For every cluster, the `_acme-challenge.[cluster name]` domain should have a CNAME record of `[cluster name].my.dns.server`. The A record for each of these records should point to the IP of the DNS server. The DNS server would be configured with `--zone my.dns.server`.

Background on CNAME delegation for DNS-01 challenges: https://www.eff.org/deeplinks/2018/02/technical-deep-dive-securing-automation-acme-dns-challenge-validation

TODO: do DNS providers typically support recursive wildcards on A requests? If not we may have to add them manually.

### DIFF
--- a/plane/src/dns/mod.rs
+++ b/plane/src/dns/mod.rs
@@ -35,10 +35,15 @@ struct AcmeDnsServer {
     loop_handle: Option<JoinHandle<()>>,
     send: Sender<MessageFromDns>,
     request_map: Arc<DashMap<ClusterName, Sender<Option<String>>>>,
+    zone: Option<String>,
 }
 
 impl AcmeDnsServer {
-    fn new(name: AcmeDnsServerName, mut client: TypedSocketConnector<MessageFromDns>) -> Self {
+    fn new(
+        name: AcmeDnsServerName,
+        mut client: TypedSocketConnector<MessageFromDns>,
+        zone: Option<String>,
+    ) -> Self {
         let (send, mut recv) = broadcast::channel::<MessageFromDns>(1);
         let request_map: Arc<DashMap<ClusterName, Sender<Option<String>>>> = Arc::default();
 
@@ -90,6 +95,7 @@ impl AcmeDnsServer {
             loop_handle: Some(loop_handle),
             send,
             request_map,
+            zone,
         }
     }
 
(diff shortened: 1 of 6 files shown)


pretentious7:
> For every cluster, the `_acme-challenge.[cluster name]` domain should have a CNAME record of `[cluster name].my.dns.server`. The A record for each of these records should point to the IP of the DNS server. The DNS server would be configured with `--zone my.dns.server`.

So the records are:
_acme-challenge.<<cluster_name>> CNAME <<cluster_name>>.my.dns.server
<<cluster_name>>.my.dns.server A <<dns_server_ip>>

and then what does the dns server do?

plane/src/main.rs
```diff
@@ -98,6 +98,16 @@ enum Command {
         #[clap(long)]
         controller_url: Url,
 
+        /// Suffix to strip from requests before looking up TXT records.
+        /// E.g. if the zone is "example.com", a TXT record lookup
+        /// for _acme-challenge.foo.bar.baz.example.com
+        /// will return the TXT records for the cluster "foo.bar.baz".
+        ///
+        /// The DNS record for foo.bar.baz in this case should have a CNAME
+        /// record pointing to this DNS server.
```
pretentious7, line 107: ahhh I understand, :+1:

plane/src/dns/mod.rs
```diff
@@ -124,6 +130,24 @@ impl AcmeDnsServer {
                     });
                 };
 
+                let name = if let Some(zone) = &self.zone {
+                    if let Some(name) = name.strip_suffix(zone) {
+                        name.strip_suffix('.').unwrap_or(name)
```
pretentious7, line 135: maybe put this strip_suffix bit in the parse implementation for cluste name or something? it's a little error prone to have to remember to do this in every code path.

pretentious7 (approved):
the clippy lints need to be addressed, also would be nice to have an example of the records that need be set on the domain registrar's dns confi in a comment.

lgtm!

PR #552, opened 2024-01-13
Add non-wildcard domain to certificate request

paulgb:
Currently, we obtain the certificate for wildcard domains only, which was ok before because we required subdomain routing. Now that we also allow path-only routing, we also want the cert to apply to the base domain itself.


rolyatmax:
I thought we were still going to require subdomain routing by default, and only turn it off for specific customers who request it.

rolyatmax (approved):

PR #558, opened 2024-01-17
Add quickstart builder action

paulgb:


.github/workflows/build-quickstart-image.yml
```diff
@@ -0,0 +1,66 @@
...
+      - name: Build and push Docker image
+        id: build-and-push
+        uses: depot/build-push-action@v1
+        with:
+          context: "."
+          file: "docker/quickstart/Dockerfile"
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+          platforms: ${{ github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != 'drifting-in-space/plane' && 'linux/amd64' || 'linux/amd64,linux/arm64' }}
```
rolyatmax, line 64: Do you know if this builds something that's compatible with macos?

rolyatmax (approved):

.github/workflows/build-quickstart-image.yml
paulgb: Yep! Docker emulates Linux on Mac, so the platform is always `linux` even when running on Mac. `linux/arm64` tells it to build the arm64 version, which is used automatically on Apple Silicon.

PR #559, opened 2024-01-18
Make pull policy and env optional

paulgb:


plane/src/types.rs
```diff
@@ -263,8 +263,9 @@ impl From<DockerRegistryAuth> for DockerCredentials {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct ExecutorConfig {
     pub image: String,
-    pub pull_policy: PullPolicy,
+    pub pull_policy: Option<PullPolicy>,
     pub credentials: Option<DockerRegistryAuth>,
+    #[serde(default)]
     pub env: HashMap<String, String>,
     pub resource_limits: ResourceLimits,
```
rolyatmax, line 270: I think we should make resource limits optional too

plane/src/types.rs
```diff
@@ -263,8 +263,9 @@ impl From<DockerRegistryAuth> for DockerCredentials {
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct ExecutorConfig {
     pub image: String,
-    pub pull_policy: PullPolicy,
+    pub pull_policy: Option<PullPolicy>,
     pub credentials: Option<DockerRegistryAuth>,
+    #[serde(default)]
```
rolyatmax, line 268: Did we say we wanted to make these all `Option`s so that if using the `ExecutorConfig` in, say, a future client library that we might offer, the developer wouldn't need to create an empty `HashMap`?

paulgb: Developers of libraries that use the JSON API can simply omit this, and serde will treat it as an empty `HashMap`. I like to avoid `Option<HashMap<...>>` because it means you have two ways of representing an empty/null hashmap: `None` and `Some(HashMap::new())`.

I also use `#[serde(default)]` for ResourceLimits, for the same reason: all of the fields are optional at the JSON level, so a default ResourceLimits is the same as an empty dict at the JSON level.

rolyatmax: From the conversation we had in person a few days ago, I thought there were two concerns with this approach:

1. Rust client code that's using the ExecutorConfig would have to slot in an empty HashMap and default ResourceLimits instead of just using `None`
2. The defaults won't always be slotted in by the server anymore if Rust clients have to slot them in when creating the struct. This can cause issues if the defaults were to change. (All Rust clients would be slotting in old defaults while non-Rust clients would have their defaults slotted in by the server.)

Maybe these issues are acceptable given other benefits of not using `Option` here? Just didn't want our previous reasoning to be forgotten.

rolyatmax (approved):

plane/src/types.rs
paulgb: My reasoning here is that `ResourceLimits` is basically just used as a way to group some fields together -- we could just as easily have something like:

```rust
#[derive(Clone, Serialize, Deserialize, Debug)]
pub struct ExecutorConfig {
    pub image: String,
    pub pull_policy: Option<PullPolicy>,
    pub credentials: Option<DockerRegistryAuth>,
    #[serde(default)]
    pub env: HashMap<String, String>,

    pub resource_limits_cpu_period: Option<DockerCpuPeriod>,
    pub resource_limits_cpu_period_percent: Option<u8>,
    // etc
}
```

The important thing to me is that the “leaf” fields are `Option<_>`, not `ResourceLimits` itself. It does mean that we can't make the leaf fields have a default other than `None`, but I'm fine with this -- I think if we do ever have defaults here, they won't be hard-coded, but will be configurable at the drone level. This means that the defaults will not live in the type itself, but that we will access them like `resource_limits.cpu_period.unwrap_or(DEFAULT_CPU_PERIOD)`.

1. is not a big issue for me, I don't think it's any more ergonomic to use `HashMap::default()` than `None` and I like to be explicit; I don't want to leave users wondering if they should use a `Some(HashMap::default())` vs. a `None` when they are effectively the same in this case.

PR #560, opened 2024-01-19
Improve cluster ergonomics

paulgb:
This fixes a few things I don't like about clusters right now:

- Since keys are (by default) shared across all clusters managed by a container, there is a weird thing where you can request a connection to a certain cluster's endpoint and get back a result from another cluster. Logically, it makes sense to consider the cluster as part of the (optional) spawn configuration. In fact, a connect request that only wants to connect to an existing backend should not need to specify a cluster at all.

- I expect the majority of self-hosting users will want only a single cluster, but they are forced to specify a cluster in every CLI command and connect request. This PR allows a `default_cluster` to be set in the controller, which is then applied if no cluster is provided in a connect request.

- Internally, even though the `backend_id` is unique across all clusters managed by the same controller, we effectively use the entire (cluster, backend) pair as a key for backends. This creates some unnecessary complexity that we can eliminate by using backend alone as the ID.

This also bumps the version to 0.4.1.


plane/src/controller/mod.rs
```diff
@@ -150,14 +154,14 @@ impl ControllerServer {
...
             .route("/c/:cluster/d/:drone/drain", post(handle_drain))
             .route(
-                "/c/:cluster/b/:backend/soft-terminate",
+                "/b/:backend/soft-terminate",
                 post(terminate::handle_soft_terminate),
             )
             .route(
-                "/c/:cluster/b/:backend/hard-terminate",
+                "/b/:backend/hard-terminate",
                 post(terminate::handle_hard_terminate),
```
paulgb, line 165: This is one of the more impactful changes of this PR; a bunch of paths no longer refer to clusters.

plane/plane-tests/tests/backend_actions.rs
```diff
@@ -6,17 +6,18 @@ use plane::{
...
 use std::time::Duration;
 
 mod common;
 
 /// Return a dummy connect request, which does not use a key.
-fn connect_request() -> ConnectRequest {
+fn connect_request(cluster: &ClusterName) -> ConnectRequest {
     ConnectRequest {
         spawn_config: Some(SpawnConfig {
+            cluster: Some(cluster.clone()),
```
paulgb, line 20: Spawn config now takes an optional cluster name. If it is not provided, we fall back on the default cluster name set on the controller.

plane/src/types.rs
```diff
@@ -285,6 +285,9 @@ impl ExecutorConfig {
 
 #[derive(Clone, Serialize, Deserialize, Debug)]
 pub struct SpawnConfig {
+    /// Cluster to spawn to. Uses the controller default if not provided.
+    pub cluster: Option<ClusterName>,
```
paulgb, line 289: Spawn config now takes an optional cluster name. If it is not provided, we fall back on the default cluster name set on the controller.

plane/src/controller/backend_status.rs
```diff
@@ -37,15 +36,15 @@ async fn backend_status(
...
     State(controller): State<Controller>,
 ) -> Result<Json<TimestampedBackendStatus>, Response> {
-    let status = backend_status(&controller, &cluster, &backend_id).await?;
+    let status = backend_status(&controller, &backend_id).await?;
     Ok(Json(status))
 }
 
 pub async fn handle_backend_status_stream(
-    Path((_cluster, backend_id)): Path<(ClusterName, BackendName)>, // TODO: check cluster id
+    Path(backend_id): Path<BackendName>, // TODO: check cluster id
```
rolyatmax, line 47: Is this comment out of date now?

paulgb: good catch.

rolyatmax (approved):
LGTM! I like stripping out the cluster for endpoints where it's not really required

PR #564, opened 2024-01-20
Initial structured logging support

paulgb:
This adds support for structured logging to Plane.

- In [init_tracing](https://github.com/drifting-in-space/plane/pull/564/files#diff-c5c2b69bfb1808e36a195b34bdbd4b0a0388d5aca2d26872e63d1495f4a1b630), we check if `PLANE_LOG_JSON` is set to a value other than `false` (or empty). If so, we output logs in JSON instead of colored log lines.
- Adds a dependency on `valuable`, which allows us to define types that serialize into JSON when JSON output is used.
- Switches times to use `chrono` for better formatting in logs and more configurable serialization.
- Introduces a few wrapper types that are used to implement `valuable::Valuable` on existing types (`LoggableTime` for `DateTime<Utc>`; `BackendAddr` for `SocketAddr`)

As a follow-up, we should add some spans to capture contextual data, and tweak log levels for dependencies, but this is a big improvement on what we have now.

### DIFF
--- a/plane/src/database/backend.rs
+++ b/plane/src/database/backend.rs
@@ -355,7 +355,6 @@ impl<'a> BackendDatabase<'a> {
 #[derive(Debug, Clone)]
 pub struct TerminationCandidate {
     pub backend_id: BackendName,
-
     pub expiration_time: Option<DateTime<Utc>>,
     pub last_keepalive: DateTime<Utc>,
     pub allowed_idle_seconds: Option<i32>,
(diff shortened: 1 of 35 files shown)


plane/src/database/backend.rs
```diff
@@ -351,9 +352,10 @@ impl<'a> BackendDatabase<'a> {
     }
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
 pub struct TerminationCandidate {
     pub backend_id: BackendName,
+
```
rolyatmax, line 358: Nit: is this newline intentional?

paulgb: removed

rolyatmax (approved):

PR #567, opened 2024-01-22
Re-enable plane-0.3.x workflows

paulgb:


feliciachang (approved):
assuming this isn't a breaking change so just approving

PR #575, opened 2024-01-27
Expose backend termination information on status endpoints

paulgb:
- renames the `backend_status` table to `backend_state`, and stores entire state there instead of status
- renames `TimestampedBackendStatus` to `BackendStatusStreamEntry` and adds additional metadata about termination
- Fixes a bug in the backend manager that caused metadata to be lost when going from terminating to terminated.
- Turns off JSON logging for local CLI commands (unrelated nit)


plane/src/types/backend_state.rs
```diff
@@ -173,9 +178,62 @@ impl Display for BackendStatus {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub termination_reason: Option<TerminationReason>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub termination_kind: Option<TerminationKind>,
+
+    /// Whether the process exited with an error. None if the process
+    /// is still running.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub exit_error: Option<bool>,
```
rolyatmax, line 198: Do we want this to include the actual exit code, not just whether an error occurred?

rolyatmax: I guess the idea is we don't want to expose exit codes for backends on an unauthed endpoint?

If we do decide to leave the exit code out, then, do we add it to the logs like we did in plane1?

paulgb: (Discussed offline)

PR #576, opened 2024-01-27
Use lowercase enum names in json representation

paulgb:
This uses lowercase values for termination kinds, termination reasons, and backend statuses/states. This results in a more idiomatic JSON API.

Note that this is a breaking change; the dev CLI will probably need to be updated. But I figure it's easier to do this now than later on.


rolyatmax (approved):
👍 We should add a linear task to update the dev CLI once this lands

PR #590, opened 2024-01-31
Add error kind

paulgb:
Adds an error kind type.

There's probably some refactoring to do here, e.g. the `StatusCode` could be determined by adding a `status()` method to the `ApiErrorKind` itself. But I wanted to focus on getting the change to the message interface in for now; that way we can refactor the code later without worrying about breaking the interface.


rolyatmax (approved):
LGTM!

PR #591, opened 2024-01-31
Allow backend id to be passed in spawn request

paulgb:


plane/src/admin.rs
```diff
@@ -102,6 +102,11 @@ enum AdminCommand {
         /// The number of seconds without any connected clients to wait before terminating the backend.
         #[clap(long)]
         max_idle_seconds: Option<i32>,
+
+        /// An optional backend to assign the string (not recommended and may be removed. Omitto allow Plane to auto-assign
```
rolyatmax, line 106: typo

rolyatmax (approved):
Looks like some `SpawnConfig`s need to be updated with `id: None`, otherwise I just found a typo. Also, worth writing a test? It's up to you. LGTM!

PR #605, opened 2024-02-07
Implement static token support

paulgb:
Implements a way to spawn a backend with a "static token". This token is then passed in to the backend as an env var, and is returned on subsequent connect requests.

Secret tokens do not make sense in the context of static tokens, so they are made optional.

TODO: confirm this is sufficient to support Jupyter.


rolyatmax (approved):
LGTM - two things to consider as follow-ups:

1. Updating docs to mention the static token option and the altered routing behavior when using a static token (i.e. not stripping the path prefix)
2. Maybe we should fail connect requests that include user/auth when the backend that's returned uses a static token.

Thoughts?

PR #607, opened 2024-02-08
address issue when calling to_terminating() on an already-terminating state

rolyatmax:
I'm seeing some `Terminating { last_status: Terminating, .. }` BackendStates. I think this is only possible if calling `to_terminating()` on an already-terminating status. This alters the `to_terminating()` and `to_terminated()` methods to return `self.clone()` if the state is already terminating or terminated since these methods should return states where the `last_status` is the last _alive_ status.

Also pushed a commit that fixes an issue with the Quickstart image on Linux where the drone can't reach the session backend containers. Maybe there's a more elegant solution, though? More info here: https://stackoverflow.com/questions/48546124/what-is-the-linux-equivalent-of-host-docker-internal

### DIFF
--- a/docker/quickstart/supervisord.conf
+++ b/docker/quickstart/supervisord.conf
@@ -20,7 +20,7 @@ stderr_logfile=/var/log/plane-controller-stderr.log
 stdout_logfile=/var/log/plane-controller-stdout.log
 
 [program:plane-drone]
-command=/bin/plane drone --controller-url ws://localhost:8080 --cluster 'localhost:9090'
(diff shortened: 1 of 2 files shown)


paulgb (approved):
LGTM. I like the idea of warning in these cases, too, but I'll let you decide whether to tack that on to this PR or leave it for a follow-up.

rolyatmax:
> LGTM. I like the idea of warning in these cases, too, but I'll let you decide whether to tack that on to this PR or leave it for a follow-up.

@paulgb updated!

PR #613, opened 2024-02-09
remove stale lock announce messages (older than 1 hour)

rolyatmax:
The World State stream grows unbounded since we aren't removing any lock announce messages. This adds functionality to the CLI's cleanup script to clean out lock announce messages that are older than 1 hour. Lock announce messages are only valid for 30 seconds, so 1 hour should be plenty of buffer.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -233,12 +233,16 @@ async fn main() -> Result<()> {
                         succeeded += 1;
                     }
                     if (failed + succeeded) % 10 == 0 {
-                        tokio::time::sleep(Duration::from_millis(50)).await;
+                        println!(
+                            "Removed {:?} stale lock announce messages so far. ({:?} failed)",
+                            succeeded, failed
+                        );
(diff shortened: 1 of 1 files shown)


rolyatmax:
After spending the day researching this, it appears that we really can't delete messages by sequence as the sequence can change. (Reading messages from a consumer's subscription then using their sequence numbers to query the underlying stream would produce a different message.) Instead, you're supposed to purge by subject. The problem here is that subjects aren't single-message - instead they are scoped to a lock. So, purging a subject can drop more recent lock announcement messages that the world state still needs. Furthermore, these lock announcement messages seem to be necessary when building the world state from scratch - not just for preventing lock assignment race conditions - since a backend will not get a lock assigned to it when building the WorldState if the announcement message was not previously observed in the stream. See this code: https://github.com/drifting-in-space/plane/blob/plane-0.3.x/core/src/state/world_state.rs#L287-L291

PR #614, opened 2024-02-09
Support drone pools 

michaelsilver:
When starting a drone, the user can specify a drone pool (string) that it belongs to. When a connect request is issued, it can contain a drone pool. If we are spawning a new backend, the scheduler must pick a drone whose drone pool matches the request.

The default drone pool (both when starting drones and connecting) is an empty string.

Fixes https://github.com/drifting-in-space/plane/issues/588
Fixes https://github.com/drifting-in-space/plane/issues/637

Usage
=====

CLI
---
```bash
$ ./dev/drone.sh --help
[...]

Options:
[...]
      --pool <POOL>
          Optional pool identifier. If present, will only schedule workloads with a matching `pool` tag on this drone

$ ./dev/cli.sh connect --help
[...]

Options:
[...]
      --pool <POOL>
          An optional drone pool (string), used when selecting where to run the backend
```

API
---
Updated the docs:

> The connect request is a `POST` request that takes a JSON body. An example payload might look like this:
```json
{
    "key": {
        "name": "room-abcde",
        "namespace": "games",
        "tag": "drop-four"
    },
    "spawn_config": {
        "executable": {
            "image": "ghcr.io/drifting-in-space/demo-image-drop-four",
        },
        "lifetime_limit_seconds": 3600,
        "max_idle_seconds": 60,
    },
    "pool": "example-pool",
    "user": "user-123",
    "auth": {
        "any arbitrary JSON object": "can go here",
        "Plane does not read it": "it is just passed through to your backend",
        "even lists and numbers": ["like", 3.14, "are", "fine"]
    }
}
```
> - `pool`: Optional string specifying the resource pool where the backend should be spawned. If not provided, the default pool is used.

Test plan
=========

Automated tests 
----------------
Added automated tests that cover what is demoed below using the CLI. 

CLI
---
In different terminals, run:

```bash
```

```bash
$ ./dev/drone.sh
```

```bash
$ ./dev/drone.sh --pool test
```

```bash
$ ./dev/proxy.sh
```

Then try scheduling backends. When `pool` isn't specified, Plane
chooses the drone without a pool parameter (default pool, empty
string). Connection requests specifying a `pool` will spawn backends
only on a matching drone (in this case `pool = "test"`). Reuse of keys
works as expected -- backends with matching keys get reused, even if
the `pool` parameter isn't passed explicitly.
```bash
$ ./dev/cli.sh connect --cluster 'localhost:9090' --key 'my-first-backend' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Created backend: vvbscjt8fkf0ry
URL: http://localhost:9090/1Nxri0x8ErET_7IE0Qt_sHKM6yV0NIRgpLYcubAsqoI/
Status URL: http://127.0.0.1:8080/pub/b/vvbscjt8fkf0ry/status
Drone: dr-bu5zvksqa25e5z
Status: scheduled at 2024-02-09 17:26:29.050 UTC
Status: loading at 2024-02-09 17:26:29.054 UTC
Status: starting at 2024-02-09 17:26:29.059 UTC
Status: waiting at 2024-02-09 17:26:29.364 UTC
Status: ready at 2024-02-09 17:26:29.368 UTC

$ ./dev/cli.sh connect --cluster 'localhost:9090' --key 'another-backend' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Created backend: kxgpc9kzmtkvpa
URL: http://localhost:9090/DgPcn3t1GQRhNMbt8Cs0AZSEOhrJY6PxCHQs_NGENg4/
Status URL: http://127.0.0.1:8080/pub/b/kxgpc9kzmtkvpa/status
Drone: dr-bu5zvksqa25e5z
Status: scheduled at 2024-02-09 17:27:03.274 UTC
Status: loading at 2024-02-09 17:27:03.278 UTC
Status: starting at 2024-02-09 17:27:03.283 UTC
Status: waiting at 2024-02-09 17:27:03.583 UTC
Status: ready at 2024-02-09 17:27:03.586 UTC

$ ./dev/cli.sh connect --cluster 'localhost:9090' --key 'third-backend' --image 'ghcr.io/drifting-in-space/demo-image-drop-four' --pool test
Created backend: fqwvz84okjucud
URL: http://localhost:9090/fLKDOEtkwH1EvktyssviT6Hqh6qH0sK8U9AOr-fl1z4/
Status URL: http://127.0.0.1:8080/pub/b/fqwvz84okjucud/status
Drone: dr-4b77o1ufsptw6p
Status: scheduled at 2024-02-09 17:28:03.958 UTC
Status: loading at 2024-02-09 17:28:03.963 UTC
Status: starting at 2024-02-09 17:28:03.969 UTC
Status: waiting at 2024-02-09 17:28:04.181 UTC
Status: ready at 2024-02-09 17:28:04.288 UTC

$ ./dev/cli.sh connect --cluster 'localhost:9090' --key 'third-backend' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Reusing backend: fqwvz84okjucud
URL: http://localhost:9090/ZBjavLGCdEW4_Po3nhPZhmVQhzTX-RNlvqGgH39cqWI/
Status URL: http://127.0.0.1:8080/pub/b/fqwvz84okjucud/status
Status: scheduled at 2024-02-09 17:28:03.958 UTC
Status: loading at 2024-02-09 17:28:03.963 UTC
Status: starting at 2024-02-09 17:28:03.969 UTC
Status: waiting at 2024-02-09 17:28:04.181 UTC
Status: ready at 2024-02-09 17:28:04.288 UTC

$ ./dev/cli.sh connect --cluster 'localhost:9090' --key 'another-backend' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Reusing backend: kxgpc9kzmtkvpa
URL: http://localhost:9090/wUXJUysHJ8BqasWHluYWeny-lIuj4q7pwB4kdAit7lQ/
Status URL: http://127.0.0.1:8080/pub/b/kxgpc9kzmtkvpa/status
Status: scheduled at 2024-02-09 17:27:03.274 UTC
Status: loading at 2024-02-09 17:27:03.278 UTC
Status: starting at 2024-02-09 17:27:03.283 UTC
Status: waiting at 2024-02-09 17:27:03.583 UTC
Status: ready at 2024-02-09 17:27:03.586 UTC
```

Drone table of the database:
```
$ ./dev/psql.sh

postgres=# select * from drone;
 id | ready | draining |        last_heartbeat         |      last_local_time       | pool
----+-------+----------+-------------------------------+----------------------------+------
  2 | t     | f        | 2024-02-09 17:23:00.242535+00 | 2024-02-09 17:23:00.241+00 |
  1 | t     | f        | 2024-02-09 17:23:04.542228+00 | 2024-02-09 17:23:04.541+00 | test
(2 rows)
```


michaelsilver:
Note for reviewer: 
I had a bit of trouble with the refactor in [6c549d1](https://github.com/drifting-in-space/plane/pull/614/commits/6c549d119084904406ef9dfc4880ea72904555f9) (see commit for details) -- if it's important, would appreciate help on getting around the pub/private/export stuff as I'm a bit stuck.

I can also squash down everything into one clean commit post-review.

plane/schema/migrations/20240209001513_add-drone-pool-column.sql
```diff
@@ -0,0 +1,6 @@
+-- Add pool column to drone table
+
+alter table drone add column pool text not null default '';
```
paulgb, line 3: we should probably enforce a limit on this string, something like a `varchar(32)` would be reasonable.

plane/src/client/mod.rs
```diff
@@ -70,11 +70,17 @@ impl PlaneClient {
...
     ) -> TypedSocketConnector<MessageFromDrone> {
-        let addr = self
-            .controller_address
-            .join(&format!("/ctrl/c/{}/drone-socket", cluster))
-            .to_websocket_address();
+        let base_path = format!("/ctrl/c/{}/drone-socket", cluster);
+        let addr = match pool {
+            Some(pool_name) => self
+                .controller_address
+                .join(&format!("{}?pool={}", base_path, pool_name)),
```
paulgb, line 79: We should encode `pool_name` here (no need to decode it on the other end; Axum will already do that since it's a query param).

I'm also wondering if we should make it part of the path, though -- maybe `/ctrl/c/{}/drone-socket` is the default (empty string) pool, and then `/ctrl/c/{}/p/{}/drone-socket` is the one for a pool.

The main factor for me in preferring it in the path is that when run behind a reverse proxy, it is generally easier to create path-based rules, so a user could create a rule like "if a drone tries to connect as part of node pool 'abc', require the bearer token 'xyz'". When we've had requests for pools, it was often from a desire to create trust barriers between drones within a cluster, so this would provide a good way to ensure that drones from one pool couldn't "impersonate" drones from another even if entirely compromised.

plane/plane-tests/tests/drone_pools.rs
```diff
@@ -0,0 +1,93 @@
...
+        name: "pool-specific-key".to_string(),
+        ..connect_request.key.unwrap()
+    });
+
+    let response_from_pool = client.connect(&connect_request_with_pool).await.unwrap();
+    tracing::info!("Got response from pool.");
+
+    assert!(response_from_pool.spawned);
+    let response_from_pool_drone = response_from_pool.drone.clone().unwrap();
+    assert_ne!(response_from_pool_drone, response.drone.unwrap());
```
paulgb, line 67: Rather than just testing that these are different, we could affirmatively check that `response.drone` is the ID of `_drone` and `response_from_pool_drone` is the ID of `_drone_in_pool`, however we would need to make the ID of the drone available as a field on `Drone` ([in this file](https://github.com/drifting-in-space/plane/blob/main/plane/src/drone/mod.rs#L142)). I wouldn't be opposed to that, but it could also be a follow-up PR.

paulgb (commented):
Looking good, I think this approach is right and the code looks good. Left a few comments.

plane/src/client/mod.rs
michaelsilver: I had a fun deep dive on how Axum's extractors work to make this happen in one handler, curious what you think of the result.

The less fancy and more explicit way to do it is just having two separate handlers. With the explicit error handling I've done, I believe the difference in the two approaches is simply style preference.

plane/plane-tests/tests/drone_pools.rs
michaelsilver: That's a good callout, I can get working on this in a follow-up PR.

plane/schema/migrations/20240209001513_add-drone-pool-column.sql
michaelsilver: Fixed, went with `varchar(255)` since that was what the rest of the schema used, including for user-defined inputs.

michaelsilver:
@paulgb ready for another look!

plane/src/client/mod.rs
```diff
@@ -70,11 +69,17 @@ impl PlaneClient {
...
-            .to_websocket_address();
+        let path = match pool {
+            Some(pool) => {
+                let encoded_pool: String =
+                    form_urlencoded::byte_serialize(pool.as_bytes()).collect();
+                format!("/ctrl/c/{}/p/{}/drone-socket", cluster, encoded_pool)
+            }
+            None => format!("/ctrl/c/{}/drone-socket", cluster),
+        };
+        let addr = self.controller_address.join(&path).to_websocket_address();
```
paulgb, line 82: If `pool` is `Some("")`, we want the second URL rather than the first.

I see a few options here:
1. call the default pool `"default"` instead of `""` so that we can use it in the path, and only construct URLs of the first form that includes the pool. (We should still accept both types of URLs for now, so that updates go smoothly when we have drones that are already running).
2. Manually handle the case of `Some("")` here
3. Stop using options for pool altogether and just use a string, where the empty string is equivalent to what `None` means now. Although it's nice to be able to distinguish between `Some/None` values, the problem here is that there are really two ways of representing the default "pool": Some("") and None.

I think I'm inclined towards 1, what do you think?

paulgb: Per offline discussion:

- Take path #3 
- Only generate URLs of the form `.../p/{}/drone-socket`, even if the drone pool is the empty string (default)
  - the resulting URL `../p//drone-socket` is a bit ugly but that doesn't matter, it's only used programmatically anyway
- Keep `/ctrl/c/{}/drone-socket` as a working endpoint for now, so that running drones/proxies do not break when the new controller is deployed.

michaelsilver: Per offline discussion (part 2): @paulgb suggests we go back to the _original query param approach_ to passing pool (rather than in the path as we have here). It's not as hard as previously thought to handle query params in reverse proxies. See also https://github.com/drifting-in-space/plane/pull/614#discussion_r1485369710

plane/src/client/mod.rs
michaelsilver: After some offline discussion (see https://github.com/drifting-in-space/plane/pull/614#discussion_r1502970060) we are moving back to using query params (instead of the path approach)

michaelsilver:
@paulgb reverted back to using the query param approach (good news, I think you already reviewed that code?) and no other changes. Lmk if this looks good to go!

plane/plane-tests/tests/drone_pools.rs
michaelsilver: Created an issue so this doesn't get lost: https://github.com/drifting-in-space/plane/issues/637

michaelsilver:
@paulgb very quick fix to remove unnecessary option wrapping pool string (sorry I missed this initially!)

plane/plane-tests/tests/drone_pools.rs
michaelsilver: This is now fixed in this PR.

michaelsilver:
Now addresses #637 as well

michaelsilver:
@paulgb I combed through to sort out all remaining uses of `Optional<String>` to instead use `pool` as a string -- the only exceptions I made were in places where there's user input, e.g. CLI (optional args), the http query param (optional), and the `ConnectRequest` (optional in JSON).

Sorry for missing these initially, I'll make sure to be more thorough going forward with changes like these.

paulgb (approved):
LGTM! Thanks for the fixes.

PR #615, opened 2024-02-12
Add plane cleanup loop

paulgb:
This adds a loop to the controller that runs ever 15 minutes and cleans up:
- old backends (only if configured with a maximum age)
- expired tokens

It also adds a dbcli command for doing this cleanup manually.

### DIFF
--- a/plane/src/cleanup.rs
+++ b/plane/src/cleanup.rs
@@ -1,0 +1,36 @@
+use crate::database::PlaneDatabase;
+use anyhow::Result;
+
+const CLEANUP_LOOP_INTERVAL_SECONDS: u64 = 60 * 15;
+
+pub async fn run_cleanup(db: &PlaneDatabase, min_age_days: Option<i32>) -> Result<()> {
+    tracing::info!("Running cleanup");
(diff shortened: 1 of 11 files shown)


plane/src/cleanup.rs
```diff
@@ -0,0 +1,36 @@
...
+    db.clean_up_tokens().await?;
+
+    tracing::info!("Done running cleanup");
+
+    Ok(())
+}
+
+pub async fn run_cleanup_loop(db: PlaneDatabase, min_age_days: Option<i32>) {
+    // Each controller runs a cleanup loop. To avoid having them all run at the same time, we
+    // introduce a random offset to the start time.
```
rolyatmax, line 22: So if we have 3 controllers running, does that mean we'll be cleaning up the DB every 5 minutes on average?

paulgb: Yep. There is absolutely nothing that tries to make the time gap uniform, but I think that's fine. These should be pretty fast queries.

rolyatmax (approved):

PR #616, opened 2024-02-12
publish metrics for each backend over postgres pubsub

rolyatmax:


plane/src/database/backend.rs
```diff
@@ -31,6 +31,33 @@ impl super::subscribe::NotificationPayload for BackendActionMessage {
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct BackendMetricsMessage {
+    pub backend_id: BackendName,
+    /// Memory used by backend excluding inactive file cache, same as use shown by docker stats
+    /// ref: https://github.com/docker/cli/blob/master/cli/command/container/stats_helpers.go#L227C45-L227C45
+    pub mem_used: u64,
+    /// Memory used by backend in bytes
+    /// (calculated using kernel memory used by cgroup + page cache memory used by cgroup)
+    pub mem_total: u64,
+    /// Active memory ( non reclaimable )
```
paulgb, line 43: nit: I know this code pre-existed, but since we're touching it, can we normalize the spacing on these comments?

plane/src/database/backend.rs
```diff
@@ -355,6 +382,15 @@ impl<'a> BackendDatabase<'a> {
         Ok(())
     }
 
+    pub async fn publish_metrics(&self, metrics: BackendMetricsMessage) -> sqlx::Result<()> {
+        emit_with_key(
```
paulgb, line 386: Currently, everything we emit is stored in the DB. We should have a way of emitting "ephemeral" messages that are not stored in the DB. Since that doesn't yet exist, I'm ok with merging this and then doing that after.

paulgb (approved):
Left a few comments. Looks good.

plane/src/database/backend.rs
rolyatmax: Good call. Pushed a commit to address this. Let me know what you think

paulgb (approved):
Nice, I like this!

PR #621, opened 2024-02-13
Move docker command test

paulgb:
This moves the docker command test into `plane-tests/`. It should be an integration test rather than a unit test because it talks to an external resource (docker).


paulgb:
Closes #611

rolyatmax (approved):

PR #624, opened 2024-02-16
add get_events_since() fn

rolyatmax:


plane/src/database/subscribe.rs
```diff
@@ -191,16 +191,10 @@ impl EventSubscriptionManager {
...
                     if let Some(prev_last_message) = last_message {
-                        let messages = match sqlx::query!(
-                            r#"
-                            select id, kind, key, data, created_at
-                            from event
-                            where id > $1
-                            order by id asc
-                            "#,
-                            prev_last_message as _,
+                        let messages = match EventSubscriptionManager::get_events_since(
```
paulgb, line 194: we should loop here until there are no new events since this introduces a limit

paulgb (approved):

PR #627, opened 2024-02-21
dont 500 when creating backends with already-held keys

rolyatmax:
It's possible to check the db and see no lock is held and then fail to insert the backend a few microseconds later because the lock has been taken. In those cases we need to recheck the lock once more and see if we can return the backend with the held lock.

This code change is pretty ugly IMHO, so I'm definitely open to alternatives.

### DIFF
--- a/plane/src/database/connect.rs
+++ b/plane/src/database/connect.rs
@@ -24,6 +24,20 @@ use std::time::Duration;
 
 const TOKEN_LIFETIME_SECONDS: u64 = 3600;
 
+/// Unique violation error code in Postgres.
+/// NOTE: typically we should use "on conflict do nothing", but that only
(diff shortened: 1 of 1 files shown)


paulgb:
I opened #629 as a suggested alteration from this, thoughts on that approach?

rolyatmax:
> I opened #629 as a suggested alteration from this, thoughts on that approach?

@paulgb Ah, didn't realize the `FailedToAcquireKey` would already retry. Removed the retry loop I had and followed your approach of just returning the `FailedToAcquireKey` connect error directly from the `create_backend_with_key`. Went further here and cleaned up a few other lines, as a result of that change.

paulgb (approved):
Good change, I like it!

PR #629, opened 2024-02-21
Fix 500s on spawn race condition

paulgb:
I think this will do what #627 does, but takes advantage of the fact that there is already a retry loop on spawns, and `ConnectError::FailedToAcquireKey` is considered retryable (`ConnectError::FailedToAcquireKey::retryable() == true`).


paulgb:
Closed in favor of #627

PR #631, opened 2024-02-21
Connection monitor refactor

paulgb:
I realized the atomics were unnecessary here and possibly leading to correctness problems. This is a move towards simpler, more readable code. I'm still inclined to rewrite this module at some point to make the logic even cleaner.

### DIFF
--- a/plane/src/proxy/connection_monitor.rs
+++ b/plane/src/proxy/connection_monitor.rs
@@ -1,20 +1,26 @@
 use crate::{heartbeat_consts::HEARTBEAT_INTERVAL, names::BackendName};
 use std::{
     collections::{hash_map::Entry, HashMap, VecDeque},
-    sync::{
-        atomic::{AtomicBool, AtomicU32, Ordering},
-        Arc, Mutex,
-    },
+    sync::{Arc, Mutex},
     time::SystemTime,
 };
 use tokio::task::JoinHandle;
 
 type BackendNameListener = Box<dyn Fn(&BackendName) + Send + Sync + 'static>;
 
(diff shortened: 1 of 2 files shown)


plane/src/proxy/connection_monitor.rs
```diff
@@ -28,26 +34,29 @@ impl ConnectionMonitor {
...
     pub fn inc_connection(&mut self, backend_id: &BackendName) {
         match self.backends.entry(backend_id.clone()) {
-            Entry::Occupied(entry) => {
-                let (active_connections, had_recent_connection) = entry.get();
-                active_connections.fetch_add(1, Ordering::SeqCst);
-                had_recent_connection.store(true, Ordering::SeqCst);
+            Entry::Occupied(mut entry) => {
+                let backend_entry = entry.get_mut();
+                backend_entry.active_connections += 1;
+                backend_entry.had_recent_connection = true;
```
rolyatmax, line 59: Why aren't the atomics required? Is the proxy not multithreaded? Can't different threads overwrite each other here?

paulgb: The `ConnectionMonitor` is [already in a `Mutex`](https://github.com/drifting-in-space/plane/blob/c1e771c87af44cbb52b3d9357b6a500bd810f527/plane/src/proxy/connection_monitor.rs#L80), so the atomics are redundant.

rolyatmax: Ah gotcha!

rolyatmax (approved):

PR #642, opened 2024-03-04
Clean up containers and images

### LINKED ISSUE #300: Docker images/container cleanup
I've noticed that after a few days of running plane, there are a few containers listed when I run `docker ps -a`. Containers that are no longer running, but were not removed. Similarly, when I run `docker images -a` I get a huge list of images.

Can you elaborate what has been implemented regarding cleanup so far? And what is planned?

Thank you so much for your continuous effort on creating such an awesome product!

paulgb:
- Add a `prune` function to `PlaneDocker` that:
  - Removes docker containers labeled as Plane backends over a certain age (based on creation time)
  - Optionally removes images that are not associated with a container
- Refactor some of the drone configuration that is passed as a growing list of arguments into a `DroneConfig` struct
- Adds cli flags `--auto-prune` and `--cleanup-min-age-seconds` to control pruning behavior.

This also removes the behavior where the drone would attempt to obtain its IP from Docker DNS. Instead, the argument to `--ip` may now be a hostname and will be attempt to be resolved, so you can pass `--ip=host.docker.internal` to achieve the old behavior.

The cleanup loop runs every minute (not currently configurable).

This will close #300.

### DIFF
--- a/plane/src/drone/docker/mod.rs
+++ b/plane/src/drone/docker/mod.rs
@@ -11,12 +11,13 @@ use crate::{
 use anyhow::Result;
 use bollard::{
     auth::DockerCredentials,
-    container::StatsOptions,
+    container::{ListContainersOptions, StatsOptions},
     errors::Error,
     service::{EventMessage, HostConfigLogConfig},
     system::EventsOptions,
     Docker,
 };
+use chrono::{DateTime, LocalResult, TimeZone, Utc};
 use std::sync::atomic::{AtomicU64, Ordering};
 use thiserror::Error;
 use tokio_stream::{Stream, StreamExt};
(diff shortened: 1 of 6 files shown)


plane/src/main.rs
```diff
@@ -79,6 +81,17 @@ enum Command {
...
+        /// Automatically prune stopped images.
+        /// This prunes *all* unused container images, not just ones that Plane has loaded, so it is disabled by default.
+        #[clap(long)]
+        auto_prune: bool,
+
+        /// Minimum age (in seconds) of backends to prune.
+        /// By default, all stopped backends are pruned, but you can set this to a positive number of seconds to prune
+        /// only backends that were created more than this many seconds ago.
+        #[clap(long, default_value = "0")]
+        cleanup_min_age_seconds: i32,
```
rolyatmax, line 94: I think it's easy to get confused over what these two flags are meant to do. `auto_prune` applies only to pruning images and `cleanup_min_age_seconds` applies only to pruning containers. I wonder if there's a different name we could give these (maybe `auto_prune_images` and/or `prune_containers_min_age_secs`). I think updating the comments to emphasize that one is for images and the other for containers would also help.

plane/src/drone/docker/mod.rs
```diff
@@ -187,6 +188,82 @@ impl PlaneDocker {
...
+                tracing::error!(?e, container_id, "Error removing container.");
+            }
+        }
+
+        if skipped > 0 {
+            tracing::info!(skipped, "Skipped pruning of {} containers.", skipped);
+        }
+
+        if auto_prune {
+            match self.docker.prune_images::<&str>(None).await {
```
rolyatmax, line 257: Would `auto_prune: true` have us lose the image that we are periodically pulling and cost cold start times? I'm guessing this is a feature we'd want to leave off in Jamsocket?

paulgb: Yes, I think that's right on both counts.

rolyatmax (approved):

PR #645, opened 2024-03-04
Support docker volume mounts

### LINKED ISSUE #586: Add volume mounts to Plane
Design doc here: https://docs.google.com/document/d/1KHkHQ823vzVDBwg5iyF6CW-pMjqNO_3LhmF1v_iNK_0/edit

michaelsilver:
Fixes #586

Design doc here: https://docs.google.com/document/d/1KHkHQ823vzVDBwg5iyF6CW-pMjqNO_3LhmF1v_iNK_0/edit

Notes:
* Drone has an optional `--mount_base` command line arg. This must be an absolute path.
* Spawn requests have an optional `mount` field in the executable config. This can be a boolean or a string (details in the doc). If it's a string, it must be a _relative path_.
* The admin CLI allows you to create a connect request only an optional string `mount` (boolean is not supported).

# Test Plan 
Tested the API and CLI using the demo detailed below. 
- [x] Mounts a custom directory (screenshot is `docs` from under the plane repo)
- [x] Mounts the "key directory" when a mount is `true`, rather than a string
- [x] Creates a directory when it doesn't exist

Screenshot 2024-02-29 at 15 20 19

Did extensive unit testing to test all possible types of `mount_base` and `mount` to ensure the correct docker mounting config is generated (in the binds field).

# Known limitations
Specifying a relative path for `mount_base` or an absolute path for `mount` will lead to a failure to mount. We do not do any validation of inputs to be friendly to the user about this.

Furthermore, a failure to mount will result in failure to spawn. The connect request will still return 200 OK but backend status will show that the backend spawn didn't succeed. There will be an error in the drone logs.

# Security
The biggest security issue to consider is escaping the `mount_base`. The unit tests should be comprehensive in showing this shouldn't be possible.

A failure to mount leads to an ERROR in the drone logs, which details the attempted mount request as follows:
```
Mount path {}, which we canonicalized as {}, is not under the specified mount base {}.
```
This will allow us to keep an eye on potentially malicious requests.

# Demo:
In this demo, we create a custom nginx image that serves any files monuted at `/plane-data`, and use this image to demo successful mounting.

`Dockerfile`:
```
FROM nginx
COPY nginx.conf /etc/nginx/nginx.conf
EXPOSE 8080
```

`nginx.conf`:
```
events {
}

http {
    server {
        listen 8080;
        server_name localhost;

        location / {
            root /plane-data;
            autoindex on;
        }
    }
}
```

Build the docker image:
```
$ docker build -t custom-nginx .
```

Then in different terminals run
```
$ ./dev/drone.sh --mount-base '/mnt/my-nfs'
$ ./dev/proxy.sh
```

Then issue a connect request (using the newly built `custom-nginx` image):
```
$ ./dev/cli.sh connect --cluster 'localhost:9090' --mount 'some-folder' --image 'custom-nginx'
```

Finally, connect to the returned URL, and you should see nginx serving an index of `/mnt/my-nfs/some-folder`

### DIFF
--- a/plane/src/drone/docker/commands.rs
+++ b/plane/src/drone/docker/commands.rs
@@ -2,7 +2,7 @@ use super::{types::ContainerId, PlaneDocker};
 use crate::{
     names::BackendName,
     protocol::AcquiredKey,
-    types::{BearerToken, ExecutorConfig},
+    types::{BearerToken, ExecutorConfig, Mount},
 };
 use anyhow::Result;
 use bollard::{
@@ -11,10 +11,16 @@ use bollard::{
     Docker,
 };
 use futures_util::StreamExt;
-use std::{collections::HashMap, time::Duration};
+use std::{
+    collections::HashMap,
+    path::{Component, Path, PathBuf},
+    time::Duration,
+};
 
 /// Port inside the container to expose.
 const CONTAINER_PORT: u16 = 8080;
+/// Base directory for any data mounted in the container from the host
+const PLANE_DATA_DIR: &str = "/plane-data";
 
 pub async fn pull_image(
     docker: &Docker,
@@ -113,6 +119,44 @@ pub async fn get_port(docker: &Docker, container_id: &ContainerId) -> Result<u16
     Err(anyhow::anyhow!("Failed to get port after retries."))
 }
 
+// Canonicalize a path without talking to the filesystem.
+//
+// Copy-pasted this function from Cargo.
+// https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
+//
+// The issue is Path's canonicalize function contacts the filesystem and checks
+// if the path exists. Instead, we effectively want a string-operation.
+//
+// We use this function before passing a directory to Docker's "Binds" field,
+// which works similarly to "Mounts", but Docker also creates the directory if
+// it does not exist.
+pub fn normalize_path(path: &Path) -> PathBuf {
+    let mut components = path.components().peekable();
+    let mut ret = if let Some(c @ Component::Prefix(..)) = components.peek().cloned() {
+        components.next();
+        PathBuf::from(c.as_os_str())
+    } else {
+        PathBuf::new()
+    };
+
+    for component in components {
+        match component {
+            Component::Prefix(..) => unreachable!(),
+            Component::RootDir => {
+                ret.push(component.as_os_str());
+            }
+            Component::CurDir => {}
+            Component::ParentDir => {
+                ret.pop();
+            }
+            Component::Normal(c) => {
+                ret.push(c);
+            }
+        }
+    }
+    ret
+}
+
 pub fn get_container_config_from_executor_config(
     backend_id: &BackendName,
     exec_config: ExecutorConfig,
@@ -120,6 +164,7 @@ pub fn get_container_config_from_executor_config(
     key: Option<&AcquiredKey>,
     static_token: Option<&BearerToken>,
     log_config: Option<&HostConfigLogConfig>,
+    mount_base: Option<&PathBuf>,
 ) -> Result<bollard::container::Config<String>> {
     let mut env = exec_config.env;
     env.insert("PORT".to_string(), CONTAINER_PORT.to_string());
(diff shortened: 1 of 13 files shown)


plane/src/drone/docker/commands.rs
```diff
@@ -146,6 +191,40 @@ pub fn get_container_config_from_executor_config(
...
+                    base.to_string_lossy()
+                ));
+            }
+            Some(vec![format!(
+                "{}:{}",
+                canonical_mount_path.to_string_lossy(),
+                PLANE_DATA_DIR
+            )])
+        }
+        _ => None,
```
paulgb, line 225: Even if we don't return an error to the user, we should at least stick a `tracing::warn!` in here if `mount_base` is `None` but `exec_config.mount` is Some.

plane/src/drone/docker/commands.rs
```diff
@@ -146,6 +191,40 @@ pub fn get_container_config_from_executor_config(
...
+                    if let Some(key) = key {
+                        base.join(&key.key.name)
+                    } else {
+                        return Err(anyhow::anyhow!(
+                            "Key is required for Bool(true) mount option"
+                        ));
+                    }
+                }
+                Mount::Path(path) => base.join(path),
+                _ => return Err(anyhow::anyhow!("Invalid mount option")),
```
paulgb, line 208: Since this is impossible (if the case is `Mount::Bool(false)` we would have fallen through on the first case of the outer `match`, let's use `unreachable!()` here instead.

If you do that, I would also use `Mount::Bool(false)` as the case here instead of `_` because it will be easier for a reader to understand _why_ it is unreachable.

plane/src/drone/docker/commands.rs
```diff
@@ -113,13 +119,52 @@ pub async fn get_port(docker: &Docker, container_id: &ContainerId) -> Result<u16
...
+// Copy-pasted this function from Cargo.
+// https://github.com/rust-lang/cargo/blob/fede83ccf973457de319ba6fa0e36ead454d2e20/src/cargo/util/paths.rs#L61
+//
+// The issue is Path's canonicalize function contacts the filesystem and checks
+// if the path exists. Instead, we effectively want a string-operation.
+//
+// We use this function before passing a directory to Docker's "Binds" field,
+// which works similarly to "Mounts", but Docker also creates the directory if
+// it does not exist.
+pub fn normalize_path(path: &Path) -> PathBuf {
```
paulgb, line 133: Instead of normalizing the path, thoughts on just validating that the portion included in the spawn request is valid? You could take the same approach of iterating over `components()`, and just fail a URL if it includes anything other than `Normal` and `Prefix` parts.

I'm trying to think of an example where someone would want to use `../` for their own mount, and it's hard to think of one.

plane/src/drone/docker/commands.rs
```diff
@@ -231,3 +313,175 @@ pub async fn run_container(
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::UNIX_EPOCH;
+
```
paulgb, line 320: nit, but I like to remove newlines in a `use` list so that rustfmt can organize the imports (rustfmt assumes that gaps are intentional so treats them as boundaries that it won't cross)

paulgb (approved):
Nice, this is in pretty good shape! Left a few comments.

plane/src/drone/docker/commands.rs
michaelsilver: Sounds good, makes sense! 

> I'm trying to think of an example where someone would want to use ../ for their own mount, and it's hard to think of one.

Just to confirm, you're saying here that a spawn with a `mount` like `"custom-mount/../different-folder/././subfolder/../../actual-folder"` (which technically would be a subfolder of `mount_base`) is ok to not allow? I think that's reasonable -- I also can't think of a scenario where someone would do that when they could do a much simpler direct path (`"actual-folder"`). 

Calling this out because I'm removing this test case.

plane/src/drone/docker/commands.rs
```diff
@@ -231,3 +291,152 @@ pub async fn run_container(
...
+    #[should_panic(expected = "not under the mount base")]
+    fn test_mount_path_invalid_parent() {
+        let mount_base = "/mnt/my-nfs";
+        let mount = Some(Mount::Path(PathBuf::from("..")));
+
+        let _config = get_container_config_from_mount(mount_base, mount)
+            .unwrap()
+            .host_config
+            .unwrap()
+            .binds;
```
paulgb, line 414: I like to reserve `should_panic` for cases where the implementation under test itself panics, rather than have the test code itself promote an `Err` into a `panic` with `unwrap()`.

In this case, something like this should work:

```rust
let err = get_container_config_from_mount(..).unwrap_err();
assert!(err.to_string(...).contains("not under the mount base"));
```

plane/src/types/mod.rs
```diff
@@ -168,6 +168,14 @@ impl From<DockerRegistryAuth> for DockerCredentials {
     }
 }
 
+// A spawn requestor can provide a mount parameter, which can be a string or a boolean.
+#[derive(Debug, Clone, Serialize, Deserialize, valuable::Valuable)]
+#[serde(untagged)]
+pub enum Mount {
+    Bool(bool),
+    Path(PathBuf),
+}
```
paulgb, line 177: Glad you figured out the clean way to do this with serde!

paulgb (commented):
Looks good! One test-related nit and then I think it's good to merge.

michaelsilver:
Addressed your `should_panic` comment!

paulgb (approved):

PR #650, opened 2024-03-06
mark-backend-lost db cli command

rolyatmax:
This adds a command to the db-cli to mark backends as Lost in cases where a failure mode has left the controller believing a backend is running when it no longer is. We will ideally address these failure modes with bugfixes and reconciliation loops, but while we do that, this will be a necessary tool for making sure the controller's state is correct.

### DIFF
--- a/plane/src/bin/db-cli.rs
+++ b/plane/src/bin/db-cli.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::println_empty_string)]
+
 use clap::{Parser, Subcommand};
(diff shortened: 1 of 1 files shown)


rolyatmax:
Tested in staging and everything works!

paulgb (approved):
LGTM, but there are a couple clippy warnings that would be good to clean up (not so much that I agree with clippy here as that appeasing it will prevent us from seeing these again -- you could also silence this clippy check globally if you prefer)

PR #658, opened 2024-03-11
fix quickstart image on mac

rolyatmax:
This fixes the Quickstart image on a mac, but breaks it on Linux.

[This PR](https://github.com/drifting-in-space/plane/pull/607) fixed that by relying on the admittedly awkward behavior of the previous `--ip` flag which [was only used](https://github.com/drifting-in-space/plane/pull/642/files#diff-a2f5b466f90757cf3ddcd106024fc1492f9ad5966124478fcb335c6364fb9a5cL200-L206) if `host.docker.internal` didn't resolve to anything.

@paulgb Any ideas on a good way to get this passing in the right IP/host for both Mac and Linux?

### DIFF
--- a/docker/quickstart/README.md
+++ b/docker/quickstart/README.md
@@ -15,6 +15,7 @@ docker run \
     -p 8080:8080 \
     -p 9090:9090 \
     -v /var/run/docker.sock:/var/run/docker.sock \
+    --add-host=host.docker.internal:host-gateway \
     plane/quickstart
 ```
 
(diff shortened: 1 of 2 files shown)


paulgb:
If we pass this to docker:

`--add-host=host.docker.internal:host-gateway`

It will set `host.docker.internal`, so it will work on Mac/Linux.

paulgb:
But I'm not opposed to a `--docker-internal-host` flag that has that behavior, so that we don't need to have the user add another flag?

rolyatmax:
From the perspective of the Jamsocket dev cli, I'm fine with just using the `--add-host` flag assuming that works.

rolyatmax:
@paulgb Pushed an update to the Quickstart README and tested on both linux and mac. Let me know if all that looks good to you, then we can land, rebuild, and I can update the Jamsocket dev cli.

paulgb (approved):

paulgb:
LGTM!

PR #667, opened 2024-03-15
Use pgpool for emit

paulgb:
Instead of acquiring a connection, just pass the pgpool.


rolyatmax (approved):
LGTM! Curious: what do we expect the effect to be here?

paulgb:
> LGTM! Curious: what do we expect the effect to be here?

Probably nothing, this just keeps it consistent with how we use it elsewhere.

PR #672, opened 2024-03-15
Another log nit

paulgb:


rolyatmax (approved):
might need a cargo fmt

paulgb:
> might need a cargo fmt

looks like it was just a github action issue, it never actually ran the check. Running today passes.

PR #679, opened 2024-03-18
Support subdomain validation

### LINKED ISSUE #619: Support subdomains
We already obtain a wildcard certificates, so we can serve traffic on `*.cluster.domain.com`. Currently, we do this indiscriminately: `foo.cluster.domain.com/connection-token/` is treated the same as `bar.cluster.domain.com/connection-token/`.

Instead, we should allow backends to specify an (optional) subdomain with their spawn config payload. If a backend is spawned for that connection request, the backend would be associated with that subdomain, and **only accessing it with that subdomain would work**.

When a request is made, the connection token would still be used to locate the appropriate backend, but the backend located would include information about the subdomain. The proxy would then compare the actual subdomain with the expected one, and fail the request (as unauthorized, or maybe bad request?) if the subdomain does not match.

Subdomains are useful when we want to serve things out of a backend that require host isolation, in particular cookies.

michaelsilver:
Adds an optional `subdomain` field to the `spawn_config`; when present, the returned URL for that session backend includes the subdomain.

Ensures that the (optional) subdomain associated with the request's backend matches the request's subdomain, as extracted from the host header.

Demo:
```bash
$ ./dev/cli.sh connect --cluster 'localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Created backend: t0tnaig8ahngha
URL: http://localhost:9090/2NcB1xFZ4ORQrZst2fnpTaXFyg5DHA3BaZSFxjeRS0U/

$ curl -Is http://localhost:9090/2NcB1xFZ4ORQrZst2fnpTaXFyg5DHA3BaZSFxjeRS0U/ | head -n 1
HTTP/1.1 200 OK
$ curl -Is http://sub.localhost:9090/2NcB1xFZ4ORQrZst2fnpTaXFyg5DHA3BaZSFxjeRS0U/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is http://sub.sub.localhost:9090/2NcB1xFZ4ORQrZst2fnpTaXFyg5DHA3BaZSFxjeRS0U/ | head -n 1
HTTP/1.1 401 Unauthorized

$ ./dev/cli.sh connect --cluster 'localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four' --subdomain 'subdomain'
Created backend: c81rmo5dd1f0k0
URL: http://subdomain.localhost:9090/BO25Mv9LkeB-uoZvTWpB-hoqoCrr2dkxyZYH24VayDE/

$ curl -Is http://subdomain.localhost:9090/BO25Mv9LkeB-uoZvTWpB-hoqoCrr2dkxyZYH24VayDE/ | head -n 1
HTTP/1.1 200 OK
$ curl -Is http://different.localhost:9090/BO25Mv9LkeB-uoZvTWpB-hoqoCrr2dkxyZYH24VayDE/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is http://localhost:9090/BO25Mv9LkeB-uoZvTWpB-hoqoCrr2dkxyZYH24VayDE/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is http://sub.subdomain.localhost:9090/BO25Mv9LkeB-uoZvTWpB-hoqoCrr2dkxyZYH24VayDE/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is http://subdomain.sub.localhost:9090/BO25Mv9LkeB-uoZvTWpB-hoqoCrr2dkxyZYH24VayDE/ | head -n 1
HTTP/1.1 401 Unauthorized

$ ./dev/cli.sh connect --cluster 'localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four' --subdomain 'sub.sub'
2024-03-19T22:53:38.484127Z ERROR plane::client: Got error response from API server. url="http://localhost:8080/ctrl/connect"
API returned error status:
    Message: Invalid subdomain provided.
    HTTP status: 400 Bad Request
    Error ID for correlating with logs: gfiimuob4ru8nx
```

Fixes #619


michaelsilver:
I still need to add tests and give this a final once-over, but putting this up early to get feedback on:
1. whether this is the correct behavior
2. whether the `host` header parsing approach seems reasonable.

Thanks!

michaelsilver:
Update from IRL convo: 
1. subdomain behavior -- correct, except we want to fail spawn requests if they include a sub-sub (etc.) domain

plane/schema/migrations/20240318145858_support-subdomains.sql
```diff
@@ -0,0 +1,4 @@
+alter table backend add column subdomain varchar(255);
```
paulgb, line 1: Since #680 also included a migration, this file will need to be renamed to come after it. Changing the filename to a current timestamp (really any timestamp after `20240319133519`) will do the trick.

michaelsilver: Thanks for the heads up!

michaelsilver: huh just rebased, ran all the tests, everything seems fine without changing this... am I missing something or does that mean we're good as is?

michaelsilver:
Ready for review! Ensured that sub-subdomains are rejected as invalid spawn requests + added tests (tests only cover URL format, not the proxy itself -- supporting proxy in plane-tests is left as future work).

plane/schema/migrations/20240318145858_support-subdomains.sql
paulgb: Tests/running locally should work because they start with a new database and run the migrations in order. I'm worried about applying to a production database where we are adding a migration to an existing stack out of order. I misspoke when I said it _needs_ to be in order (it used to be the case but it's [been fixed](https://github.com/launchbadge/sqlx/issues/869)), but I still think it's good if we try to keep migrations in the migrations folder in the same order that they are actually run in production, in order to ensure that the order of migrations in the code matches the order they were run against prod databases.

plane/src/database/connect.rs
```diff
@@ -304,6 +310,12 @@ async fn attempt_connect(
         .or(default_cluster)
         .ok_or(ConnectError::NoClusterProvided)?;
 
+    if let Some(subdomain) = &spawn_config.subdomain {
+        if subdomain.contains('.') {
+            return Err(ConnectError::InvalidSubdomain);
+        }
```
paulgb, line 316: Subdomains can consist of `a-z`, `0-9`, and `-` but not as the first or last char. How about this:

```rust
let valid_subdomain = s.chars().all(|c| c.is_alphanumeric() || c == '-')
    && !s.starts_with('-') && !s.ends_with('-')
if !valid_subdomain {...}
```

plane/src/proxy/proxy_service.rs
```diff
@@ -143,6 +143,14 @@ impl RequestHandler {
                 .body(hyper::Body::empty())?);
         };
 
+        // If the host header doesn't parse (non-ascii) or doesn't match the backend's stored subdomain, return an error
+        if !matches!(request_rewriter.get_subdomain(), Ok(request_subdomain) if request_subdomain == route_info.subdomain.as_deref())
```
paulgb, line 147: I find this a bit hard to parse, but if we land #684 first we can do something like:

```rust
let request_subdomain = request_rewriter.get_subdomain().map_err(|_| ProxyError::InvalidSubdomain)?;
if request_subdomain != route_info.subdomain.as_deref() {
    return Err(ProxyError::InvalidSubdomain);
}
```

plane/src/proxy/rewriter.rs
```diff
@@ -61,6 +62,19 @@ impl RequestRewriter {
...
+    }
+
+    pub fn get_subdomain(&self) -> Result<Option<&str>, ToStrError> {
+        Ok(self
+            .get_host_header()
+            .map(|h| h.to_str())
+            .transpose()?
+            .and_then(|h| h.rsplit_once('.'))
+            .map(|(before_domain_part, _)| before_domain_part))
+    }
```
paulgb, line 76: I think we want to thread through the cluster name so that we have it here, and then remove it as a suffix and return whatever's left. Otherwise, if we have a cluster like `p.jamsocket.net` without a subdomain this will still return `Ok(Some("p"))`.

paulgb (commented):
Nice work! Left some comments.

plane/src/proxy/rewriter.rs
michaelsilver: Fixed! Tried to leverage idiomatic Rust as much as possible -- let me know if you have any feedback! 

Setup:
```diff
$ git diff dev/
diff --git a/dev/controller.sh b/dev/controller.sh
index 6b471c3..f543f4f 100755
--- a/dev/controller.sh
+++ b/dev/controller.sh
@@ -3,5 +3,5 @@
 cargo run -- \
     controller \
     --db postgres://postgres@localhost \
-    --default-cluster localhost:9090 \
+    --default-cluster p.localhost:9090 \
     "$@"
diff --git a/dev/drone.sh b/dev/drone.sh
index 89c902a..a99c899 100755
--- a/dev/drone.sh
+++ b/dev/drone.sh
@@ -3,6 +3,6 @@
 cargo run -- \
     drone \
     --controller-url ws://localhost:8080/ \
-    --cluster "localhost:9090" \
+    --cluster "p.localhost:9090" \
     --db "drone.sqlite" \
     "$@"
diff --git a/dev/proxy.sh b/dev/proxy.sh
index 48b50bf..db51fc1 100755
--- a/dev/proxy.sh
+++ b/dev/proxy.sh
@@ -3,5 +3,5 @@
 cargo run -- \
     proxy \
     --controller-url ws://localhost:8080/ \
-    --cluster "localhost:9090" \
+    --cluster "p.localhost:9090" \
     "$@"
```

Demo:
```shell
$ ./dev/cli.sh connect --cluster 'p.localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four' --subdomain 'subdomain'
Created backend: qevn9013knwmgo
URL: http://subdomain.p.localhost:9090/btaUziMNMxuO-ujV7TwyxfrAb7uoJPOzSrQewdYNjjk/

$ curl -Is http://subdomain.p.localhost:9090/btaUziMNMxuO-ujV7TwyxfrAb7uoJPOzSrQewdYNjjk/
HTTP/1.1 200 OK
$ curl -Is http://p.localhost:9090/btaUziMNMxuO-ujV7TwyxfrAb7uoJPOzSrQewdYNjjk/
HTTP/1.1 401 Unauthorized
$ curl -Is http://subdomainp.localhost:9090/btaUziMNMxuO-ujV7TwyxfrAb7uoJPOzSrQewdYNjjk/
HTTP/1.1 400 Bad Request
$ curl -Is http://localhost:9090/btaUziMNMxuO-ujV7TwyxfrAb7uoJPOzSrQewdYNjjk/
HTTP/1.1 400 Bad Request
$ curl -Is http://subdomain.localhost:9090/btaUziMNMxuO-ujV7TwyxfrAb7uoJPOzSrQewdYNjjk/
HTTP/1.1 400 Bad Request

$ ./dev/cli.sh connect --cluster 'p.localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Created backend: vsyu98ie6eud66
URL: http://p.localhost:9090/vQHaeFhv_OV6b-R0u0xAN5hMpCRTdYPbKS4Zozc9aUs/

$ curl -Is http://p.localhost:9090/vQHaeFhv_OV6b-R0u0xAN5hMpCRTdYPbKS4Zozc9aUs/
HTTP/1.1 200 OK
$ curl -Is http://sub.p.localhost:9090/vQHaeFhv_OV6b-R0u0xAN5hMpCRTdYPbKS4Zozc9aUs/
HTTP/1.1 401 Unauthorized
$ curl -Is http://subp.localhost:9090/vQHaeFhv_OV6b-R0u0xAN5hMpCRTdYPbKS4Zozc9aUs/
HTTP/1.1 400 Bad Request

File: plane/src/proxy/rewriter.rs [hunk 5]
```diff
@@ -61,6 +77,32 @@ impl RequestRewriter {
...
+            .map(|h| {
+                h.strip_suffix(cluster.as_str())
+                    .ok_or(RequestRewriterError::InvalidHostHeader)
+            })
+            .transpose()? // doesn't end in cluster name
+            .filter(|s| !s.is_empty()) // filter out here if there's no subdomain
+            .map(|s| {
+                s.strip_suffix('.')
+                    .ok_or(RequestRewriterError::InvalidHostHeader)
+            })
```
paulgb, line 101: I think the order here should be flipped, we want to check if the subdomain is empty after removing the `.`. I guess it doesn't really matter since the client would have to do something intentionally weird to send a host header of the form `.{cluster}`, and we'll still reject the request, but assuming that it's the ~same code complexity to handle that case by rejecting the domain I think it's worth doing that.

plane/src/proxy/rewriter.rs
```diff
@@ -61,6 +77,32 @@ impl RequestRewriter {
...
+            .transpose()?; // the remaining string must be a subdomain ending in dot
```
paulgb, line 102: I think this comment is outdated? We already strip the dot.

plane/src/types/mod.rs
```diff
@@ -66,6 +70,14 @@ impl FromStr for ClusterName {
...
 }
 
+impl TryFrom<String> for ClusterName {
+    type Error = &'static str;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        Self::from_str(&value)
+    }
+}
+
```
paulgb, line 80: is there a reason you can't just call `ClusterName::from_str` in `database/backend.rs`?

plane/src/proxy/rewriter.rs
```diff
@@ -16,6 +20,18 @@ const PATH_PREFIX_HEADER: &str = "x-verified-path";
...
+pub enum RequestRewriterError {
+    #[error("Invalid `host` header")]
+    InvalidHostHeader,
+}
+
+impl From<hyper::header::ToStrError> for RequestRewriterError {
+    fn from(_: hyper::header::ToStrError) -> Self {
+        RequestRewriterError::InvalidHostHeader
+    }
+}
```
paulgb, line 33: Yes, this is a good and idiomatic use of `From`!

You may have noticed that `?` will do a conversion when there is a `From` implementation, so something like:

```rust
fn blah() -> Result<(), ToStrError> {
    let c: Result<(), RequestRewriterError> = blahblah();
    c?;
    Ok(())
}
```

will work thanks to this implementation.

plane/src/proxy/proxy_service.rs
```diff
@@ -181,6 +196,10 @@ impl RequestHandler {
             return Err(ProxyError::InvalidConnectionToken);
         };
 
+        if request_rewriter.get_subdomain(&route_info.cluster)? != route_info.subdomain.as_deref() {
+            return Err(ProxyError::InvalidSubdomain);
+        }
```
paulgb, line 201: let's add a `tracing::warn!` here for now, we can remove it later if it gets noisy

paulgb (commented):
Looks good! Left a couple more comments, but I think this is almost ready to ship.

plane/src/types/mod.rs
michaelsilver: Not really, just syntactic sugar and seemed more uniform with the other types' `try_from` calls. I'll delete the function and call `from_str(&var)`

michaelsilver:
Introduced a custom `Subdomain` type to better leverage Rust's type system and do the validation more idiomatically.

Demo showing this still works:
```shell
$ ./dev/cli.sh connect --cluster 'p.localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four' --subdomain 'subdomain'
Created backend: qejkle5d79h1hp
URL: http://subdomain.p.localhost:9090/krP-NMAEIn_ZqUmcFHGpDGVRzeDNL3NARIXgtxaYmj8/

$ curl -Is http://subdomain.p.localhost:9090/krP-NMAEIn_ZqUmcFHGpDGVRzeDNL3NARIXgtxaYmj8/ | head -n 1
HTTP/1.1 200 OK
$ curl -Is http://sub.p.localhost:9090/krP-NMAEIn_ZqUmcFHGpDGVRzeDNL3NARIXgtxaYmj8/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is http://p.localhost:9090/krP-NMAEIn_ZqUmcFHGpDGVRzeDNL3NARIXgtxaYmj8/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is http://localhost:9090/krP-NMAEIn_ZqUmcFHGpDGVRzeDNL3NARIXgtxaYmj8/ | head -n 1
HTTP/1.1 400 Bad Request

$ ./dev/cli.sh connect --cluster 'p.localhost:9090' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'
Created backend: qlk1a8togfvla9
URL: http://p.localhost:9090/ypy4y_mrhC9SjhqUjYMMiJstFTe4ngAbHRbJSPxBhTQ/

$ curl -Is  http://p.localhost:9090/ypy4y_mrhC9SjhqUjYMMiJstFTe4ngAbHRbJSPxBhTQ/ | head -n 1
HTTP/1.1 200 OK
$ curl -Is  http://sub.p.localhost:9090/ypy4y_mrhC9SjhqUjYMMiJstFTe4ngAbHRbJSPxBhTQ/ | head -n 1
HTTP/1.1 401 Unauthorized
$ curl -Is  http://localhost:9090/ypy4y_mrhC9SjhqUjYMMiJstFTe4ngAbHRbJSPxBhTQ/ | head -n 1
HTTP/1.1 400 Bad Request
```

paulgb (approved):

PR #680, opened 2024-03-19
Fix controller race

paulgb:
Normally, a drone will cleanly cut from a controller and be marked offline before connecting to another controller. However, there is technically a race condition where if a dropped connection is not written to the DB by controller A before the drone reconnects to controller B, controller A will mark the drone as offline after controller B has written it online, causing the drone to appear to be permanently offline according to the database.

This fixes the problem by only marking a drone as offline if it is connected to the current controller.


rolyatmax (approved):
Good catch!

PR #682, opened 2024-03-19
Fix unwrap and satisfy clippy

paulgb:


plane/src/database/cluster.rs
```diff
@@ -14,7 +13,7 @@ impl<'a> ClusterDatabase<'a> {
         Self { pool }
     }
 
-    pub async fn cluster_state(&self, cluster: &ClusterName) -> anyhow::Result<ClusterState> {
+    pub async fn cluster_state(&self, cluster: &ClusterName) -> sqlx::Result<ClusterState> {
```
rolyatmax, line 16: Was this a clippy-motivated change?

rolyatmax (approved):
LGTM

plane/src/database/cluster.rs
paulgb: We implement `or_internal_error(...)` on a `sqlx::Result` but not an `anyhow::Result`.

It's sort of clippy-motivated because clippy pointed out the unwrap that got introduced, and `or_internal_error` is how we typically convert errors to response messages.

PR #684, opened 2024-03-19
Add CORS to errors served by proxy

paulgb:
The desired behavior here is:

- Add CORS headers if returning an error _from the proxy itself_
- Don't muck with headers if returning an error from the backend itself
- If the request is an `OPTIONS` request and results in an error, add CORS headers and send a 200-level response


michaelsilver (approved):
Awesome!

PR #685, opened 2024-03-20
update idx_backend_drone_id

rolyatmax:
I've noticed that the query for selecting a drone for scheduling can be quite slow due to an expensive sequence scan of the backend table. I also noticed that the index we have for that query includes the cluster, but the query doesn't. Would this updated index help speed this query up?

### DIFF
--- a/plane/schema/derived_schema.sql
+++ b/plane/schema/derived_schema.sql
@@ -912,7 +912,7 @@ CREATE INDEX idx_backend_action_pending ON public.backend_action USING btree (dr
 -- Name: idx_backend_drone_id; Type: INDEX; Schema: public; Owner: postgres
 --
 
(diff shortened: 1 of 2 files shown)


paulgb:
Nice catch! I think this should help. The current index would also apply if we specify the `cluster` and not the `drone_id` (since it's a btree index and cluster comes first) but I don't see any queries we have currently that do that, so I think this is a good change.

paulgb (approved):

michaelsilver (commented):
Heads up — it looks like the query we make uses [Terminated as uppercase](https://github.com/drifting-in-space/plane/blob/main/plane/src/database/drone.rs#L109) where the index is lowercase.

rolyatmax:
> Heads up — it looks like the query we make uses [Terminated as uppercase](https://github.com/drifting-in-space/plane/blob/main/plane/src/database/drone.rs#L109) where the index is lowercase.

Oh! Nice catch!

PR #686, opened 2024-03-20
Correct queries filtering on backend.last_status [part of DIS-1824]

michaelsilver:
At some point backend status changed from capitalized to all lowercase. The BackendStatus type implements the Display trait and ensures under the hood that all lowercase is used. We've used this in our SQL queries instead of string literals, but missed a couple spots.

Most notably, this caused `pick_drone_for_spawn` to incorrectly filter on (capitalized) `last_status != 'Terminated'`, which is not only incorrect, but potentially costly in performance because our index is built on a lowercase `last_status != 'terminated'`:
```sql
create index idx_backend_drone_id on backend(drone_id) where last_status != 'terminated';
```


rolyatmax (approved):
Good catch!

PR #697, opened 2024-04-22
Return GONE on terminated backend

paulgb:
- `route_info_for_token` returns `None` if the backend is not ready
- adds `MessageToProxy::BackendRemoved` variant. This will be used for the controller to notify drones when a backend has been terminated, and tells it to invalidate any connection tokens associated with that backend. This is not currently sent from the controller yet, but will be in a future PR.
- Modifies handling of `InvalidConnectionToken` to return `GONE` instead of `FORBIDDEN`.

Fixes DIS-1891

### DIFF
--- a/plane/src/proxy/proxy_service.rs
+++ b/plane/src/proxy/proxy_service.rs
@@ -117,9 +117,10 @@ impl RequestHandler {
             Ok(response) => Ok(response),
             Err(err) => {
                 let (status_code, body) = match err {
-                    ProxyError::InvalidConnectionToken => {
-                        (hyper::StatusCode::FORBIDDEN, "Forbidden")
-                    }
+                    ProxyError::InvalidConnectionToken => (
+                        hyper::StatusCode::GONE,
+                        "The backend is no longer available or the connection token is invalid.",
+                    ),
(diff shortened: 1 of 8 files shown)


rolyatmax (approved):

plane/src/proxy/proxy_service.rs
```diff
@@ -117,9 +117,10 @@ impl RequestHandler {
             Ok(response) => Ok(response),
             Err(err) => {
                 let (status_code, body) = match err {
-                    ProxyError::InvalidConnectionToken => {
-                        (hyper::StatusCode::FORBIDDEN, "Forbidden")
-                    }
+                    ProxyError::InvalidConnectionToken => (
+                        hyper::StatusCode::GONE,
+                        "The backend is no longer available or the connection token is invalid.",
```
rolyatmax, line 122: Is there any way to differentiate between these two cases?

paulgb: Unfortunately not reliably, because we only keep connection tokens around until they expire and we would need to keep them around indefinitely. We could encode the expiration time in the token, I suppose, and then for (supposedly) unexpired tokens we would know that it was invalid if we did not find it.

PR #705, opened 2024-04-24
Add DronePoolName type

michaelsilver:
Also remove pool from docs -- pool was mentioned but under the wrong place. We don't elaborate on
other spawn config options, so removing it entirely.

### DIFF
--- a/plane/src/types/mod.rs
+++ b/plane/src/types/mod.rs
@@ -70,6 +70,53 @@ impl FromStr for ClusterName {
     }
 }
 
+#[derive(
+    Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, valuable::Valuable,
+)]
+pub struct DronePoolName(String);
+
+impl DronePoolName {
+    pub fn is_default(&self) -> bool {
+        self == &Self::default()
+    }
+
+    pub fn as_str(&self) -> &str {
+        &self.0
+    }
+}
+
+impl From<String> for DronePoolName {
+    fn from(s: String) -> Self {
+        DronePoolName(s)
+    }
+}
+
+impl From<&str> for DronePoolName {
+    fn from(s: &str) -> Self {
+        DronePoolName(s.to_string())
+    }
+}
+
+impl From<&Option<String>> for DronePoolName {
+    fn from(opt: &Option<String>) -> Self {
+        DronePoolName(opt.as_deref().unwrap_or_default().to_string())
+    }
+}
+
+impl Display for DronePoolName {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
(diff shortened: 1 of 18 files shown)


plane/src/types/mod.rs
```diff
@@ -70,6 +70,53 @@ impl FromStr for ClusterName {
...
+    }
+}
+
+impl From<&str> for DronePoolName {
+    fn from(s: &str) -> Self {
+        DronePoolName(s.to_string())
+    }
+}
+
+impl From<&Option<String>> for DronePoolName {
```
michaelsilver, line 100: This is only used in the platform. Does it make sense to keep here or only implement in the platform?

plane/src/types/mod.rs
```diff
@@ -70,6 +70,53 @@ impl FromStr for ClusterName {
...
 }
 
+#[derive(
+    Default, Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash, valuable::Valuable,
+)]
+pub struct DronePoolName(String);
+
+impl DronePoolName {
+    pub fn is_default(&self) -> bool {
+        self == &Self::default()
```
paulgb, line 80: I was curious if this would involve an allocation since it creates a new string under the hood, but it turns out it does not! https://doc.rust-lang.org/beta/alloc/string/struct.String.html#method.new

(Not that I encourage that sort of optimization anyway, I was just curious)

plane/src/types/mod.rs
paulgb: A trait `impl` has to be in the same module as either the `trait` or the type being implemented for (`DronePoolName` here), to avoid ambiguous impls. That said, it might be more explicit just to implement `From<String>`, and the call site call `optional_pool.map(DronePoolName::from).unwrap_or_default()`.

(`From<&str>` would also work, I have a slight preference for `From<String>` because `From<String>` will compile to a no-op whereas `From<&str>` will allocate and copy the string even if the original is never used again.)

plane/src/types/mod.rs
```diff
@@ -219,7 +266,7 @@ pub struct SpawnConfig {
 
     /// The drone pool to use for the connect request.
     #[serde(default)]
-    pub pool: Option<String>,
+    pub pool: Option<DronePoolName>,
```
paulgb, line 269: It would probably simplify things to make this just `DronePoolName`, since we already have `#[serde(default)]` and we don't care about the distinction between this being `None` and `Some(DronePoolName::default())`.

paulgb (commented):
Looks good, a couple nitty commends (and one no-action comment)

plane/src/types/mod.rs
michaelsilver: Oooh, great idea! Vastly simplifies a bunch of stuff and makes the code a lot cleaner :)

paulgb (approved):
LGTM!

PR #710, opened 2024-05-01
Fix Mac IP issue

### LINKED ISSUE #704: The backend is no longer available or the connection token is invalid error.
Hello, thank you for creating such a wonderful product.

I've been trying out Plane, but I encountered the following error: "The backend is no longer available or the connection token is invalid."

Here are the steps I followed:

1. git clone https://github.com/drifting-in-space/plane.git
1. cd plane
1. docker compose -f docker/docker-compose.yml up
1. docker/cli.sh connect --cluster 'localhost:9090' --key 'my-first-backend' --image 'ghcr.io/drifting-in-space/demo-image-drop-four'

Initially, I was able to successfully make requests several times, and I confirmed that the Docker containers were launching for each request. However, after a few days, I attempted again and encountered this error.

Additionally, I am seeing the following log in the Docker compose logs:

plane::drone::backend_manager: Updating backend state backend_id=BackendName("jqpwjvwmtfkerb") state={"status": "waiting", "address": ("127.0.0.1:61539",)}

When I access 127.0.0.1:61539, I can see the Drop Four page.

Could you please help me understand what might be going wrong?

paulgb:
This appears to fix #704

- [x] Test that this doesn't break things on Linux

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.9"
 services:
   controller:
     container_name: plane-controller
@@ -22,7 +21,7 @@ services:
     volumes:
       - /var/run/docker.sock:/var/run/docker.sock
     command:
-      "drone --controller-url ws://plane-controller:8080 --cluster 'localhost:9090'"
+      "drone --controller-url ws://plane-controller:8080 --cluster 'localhost:9090' --ip='host.docker.internal'"
     networks:
       - plane-dev
     extra_hosts:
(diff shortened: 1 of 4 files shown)


michaelsilver (approved):

docker/docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: "3.9"
```
paulgb, line 1: this was obsoleted and now results in a warning if present

Issue #712, opened 2024-05-04
Make Runtimes Generic


paulgb:
- [x] Make the executor configuration generic (#713)
- [x] Turn executor config into an enum (#717)
- [x] Make the executor itself generic
  - [x] Move pruning loop into `PlaneDocker` (#718)
  - [x] Move readiness check into `PlaneDocker` (#728)
  - [x] Move Docker construction into `PlaneDocker` (#719)
  - [x] Rename `PlaneDocker` to `DockerRuntime` (#720)
  - [x] Make `ContainerId` an implementation detail of `DockerRuntime` (#721)
  - [x] Generalize pull (#722)
  - [x] Implement `Runtime` trait (#723)
  - [x] Move metrics loop into docker side (#724)
  - [x] Replace `DockerRuntime` with generic `Runtime` in the main loop
    - [x] Make `BackendManager` generic over runtime (#727)
  - [x] Refactor codebase to create a module for runtimes
- [x] Send backend ID as an HTTP header to the backend (so that multiple backends can share a port) (#714)
- [x] Spec out a Drone <> Executor wire protocol (#766)
- [x] Implement an executor that uses the wire protocol over domain sockets (#766, #780)
- [ ] Create a mock executor and use it to add integration tests

PR #718, opened 2024-05-06
Move `prune` into PlaneDocker executor

paulgb:
`PlaneDocker` exposes a method `prune` which is run on a cleanup loop. The specifics of this cleanup are an implementation detail that only the Docker executor cares about. Since we want to move towards `PlaneDocker` implementing a generic, executor-agnostic trait, we want to eliminate this.

The approach here is to move the pruning loop into the PlaneExecutor itself. The loop starts when the executor is constructed, and runs until it is dropped.

In order to facilitate this, the configuration for pruning has to be moved into the executor config instead of the general drone config, which this PR also does.

Part of #712

### DIFF
--- a/plane/plane-tests/tests/common/test_env.rs
+++ b/plane/plane-tests/tests/common/test_env.rs
@@ -117,6 +117,8 @@ impl TestEnvironment {
             runtime: None,
             log_config: None,
             mount_base: mount_base.map(|p| p.to_owned()),
+            auto_prune: Some(false),
+            cleanup_min_age: Some(Duration::zero()),
         };
 
         #[allow(deprecated)] // `docker_config` field is deprecated.
@@ -126,8 +128,8 @@ impl TestEnvironment {
             ip: IpAddr::V4(Ipv4Addr::LOCALHOST),
             db_path: Some(self.scratch_dir.join("drone.db")),
             pool: pool.clone(),
-            auto_prune: false,
-            cleanup_min_age: Duration::zero(),
+            auto_prune: None,
+            cleanup_min_age: None,
             executor_config: Some(ExecutorConfig::Docker(docker_config)),
             docker_config: None,
             controller_url: controller.url().clone(),
--- a/plane/src/drone/command.rs
+++ b/plane/src/drone/command.rs
@@ -67,20 +67,22 @@ impl DroneOpts {
             .map(|s| serde_json::from_str(&s))
             .transpose()?;
 
+        let cleanup_min_age =
+            Duration::try_seconds(self.auto_prune_containers_older_than_seconds as i64)
+                .expect("valid duration");
+
         let docker_config = PlaneDockerConfig {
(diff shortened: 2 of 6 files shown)


plane/plane-tests/tests/common/test_env.rs
```diff
@@ -126,8 +128,8 @@ impl TestEnvironment {
             ip: IpAddr::V4(Ipv4Addr::LOCALHOST),
             db_path: Some(self.scratch_dir.join("drone.db")),
             pool: pool.clone(),
-            auto_prune: false,
-            cleanup_min_age: Duration::zero(),
+            auto_prune: None,
+            cleanup_min_age: None,
```
paulgb, line 132: these are deprecated; while they are still around, the values inside the docker config will take precedence if specified.

plane/src/drone/command.rs
```diff
@@ -67,30 +67,32 @@ impl DroneOpts {
...
             cluster: self.cluster.clone(),
             ip,
             db_path: self.db,
             pool: self.pool,
-            auto_prune: self.auto_prune_images,
-            cleanup_min_age,
-            docker_config: None,
+            auto_prune: None,      // deprecated
+            cleanup_min_age: None, // deprecated
+            docker_config: None,   // deprecated
```
paulgb, line 95: all replaced by the line below

plane/src/drone/mod.rs
```diff
@@ -178,11 +178,19 @@ impl Drone {
...
+                docker_config.cleanup_min_age =
+                    docker_config.cleanup_min_age.or(config.cleanup_min_age);
+
                 PlaneDocker::new(Docker::connect_with_local_defaults()?, docker_config).await?
             }
-            (None, Some(ExecutorConfig::Docker(docker_config))) => {
+            (None, Some(ExecutorConfig::Docker(mut docker_config))) => {
+                docker_config.auto_prune = docker_config.auto_prune.or(config.auto_prune);
+                docker_config.cleanup_min_age =
+                    docker_config.cleanup_min_age.or(config.cleanup_min_age);
```
paulgb, line 192: this is a little messy, but will be gone once we remove the now-deprecated `auto_prune` and `cleanup_min_age` from the global drone config.

plane/src/serialization.rs
```diff
@@ -17,3 +17,31 @@ pub mod serialize_duration_as_seconds {
...
+    {
+        let seconds: Option<i64> = Deserialize::deserialize(deserializer)?;
+        match seconds {
+            Some(seconds) => Ok(Some(
+                Duration::try_seconds(seconds).expect("valid duration"),
+            )),
+            None => Ok(None),
+        }
+    }
+}
```
paulgb, line 47: this is only needed during the transition period where the prune configuration is allowed at either the drone and executor level.

rolyatmax (approved):

PR #721, opened 2024-05-07
Remove ContainerId from runtime interface

paulgb:
`ContainerId` is an implementation detail specific to Docker, but we pass it across the runtime interface a lot. This changes it so that we only ever pass `BackendName`s around, and derive the container ID from these names where necessary.

Part of #712

### DIFF
--- a/plane/src/drone/docker/mod.rs
+++ b/plane/src/drone/docker/mod.rs
@@ -41,7 +41,7 @@ pub mod types;
 const PLANE_DOCKER_LABEL: &str = "dev.plane.backend";
 
 fn backend_id_to_container_id(backend_id: &BackendName) -> ContainerId {
-    ContainerId::from(format!("plane-{}", backend_id.to_string()))
+    ContainerId::from(format!("plane-{}", backend_id))
 }
 
 #[derive(Clone, Debug, Default, Serialize, Deserialize)]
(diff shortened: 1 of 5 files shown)


plane/src/drone/docker/mod.rs
```diff
@@ -40,6 +40,10 @@ pub mod types;
 /// The existence of this label is used to determine whether a container is managed by Plane.
 const PLANE_DOCKER_LABEL: &str = "dev.plane.backend";
 
+fn backend_id_to_container_id(backend_id: &BackendName) -> ContainerId {
+    ContainerId::from(format!("plane-{}", backend_id.to_string()))
+}
```
michaelsilver, line 45: nit: listen to clippy

michaelsilver: Can we use `From` and/or `into` for this to leverage Rust typing more?

michaelsilver (approved):
LGTM, left a nit + one optional suggestion to keep the codebase clean. Lmk if this is the sort of thing we'd block a Plane PR on or not, still calibrating.

plane/src/drone/docker/mod.rs
paulgb: I actually think it's a good thing for the conversion to be explicit here. If it was just a `From` conversion, I would expect it to be just re-packing the same string, not modifying it in an application-specific way.

PR #724, opened 2024-05-09
Refactor metrics out of the executor

paulgb:
This refactors metrics collection so that it lives entirely on the side of the runtime. This way, docker-specific things are confined to the Docker runtime.

The way it works is:

- `metrics()` on the `Runtime` trait is replaced with a `metrics_callback()` method, which can be used to pass in a callback that is called when metrics are available
- Container "start" events are added to the list of the events that the Docker runtime listens for (nb. these are currently only received if the events stream is being polled -- decoupling this is added as a todo)
- When a container start event is received, we pull the backend ID from the container metadata and start a tokio task which open a metrics stream for that container. When metrics are received, we convert them to a metrics message and invoke the callback.

### DIFF
--- a/plane/src/drone/docker/mod.rs
+++ b/plane/src/drone/docker/mod.rs
@@ -4,6 +4,7 @@ use self::{
 };
 use crate::{
     database::backend::BackendMetricsMessage,
+    drone::{docker::metrics::metrics_loop, runtime::Runtime},
     heartbeat_consts::KILL_AFTER_SOFT_TERMINATE_SECONDS,
     names::BackendName,
     protocol::AcquiredKey,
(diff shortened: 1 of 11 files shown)


plane/src/drone/docker/mod.rs
```diff
@@ -56,11 +54,106 @@ pub struct DockerRuntimeConfig {
...
+        until: None,
+        filters: vec![
+            ("type", vec!["container"]),
+            ("event", vec!["die", "stop", "start"]),
+            ("label", vec![PLANE_DOCKER_LABEL]),
+        ]
+        .into_iter()
+        .collect(),
+    };
+    //docker.events(Some(options)).filter_map(|e| {
```
rolyatmax, line 84: I'm guessing this was left in unintentionally?

rolyatmax (approved):

plane/src/drone/docker/mod.rs
paulgb: No actually. Removed.

PR #728, opened 2024-05-13
Move backend wait to runtime

paulgb:
Currently, the behavior to check if a backend is ready lives outside of the runtime, and depends on polling the port. An alternate runtime may use the same port for all backends (enabled by #714), so we need to allow the runtime to implement the readiness check.

For now, we still thread back the address through the executor and pass it to the readiness check. This is still exposing more of an implementation detail than I would like, but it will require some refactoring of the Docker executor to remove it.


rolyatmax (approved):
LGTM

PR #730, opened 2024-05-13
More sophisticated cleanup

paulgb:
The backend action cleanup did not actually work before, because it would violate foreign key constraints. (We haven't enabled it yet, which is why we didn't see failures).

This changes the cleanup to a transaction that finds all backends that are candidates for deletion and deletes rows of other tables that have foreign keys to them first.


plane/src/database/backend.rs
```diff
@@ -455,21 +455,79 @@ impl<'a> BackendDatabase<'a> {
...
+
+        txn.commit().await?;
+
+        if backend_deleted > 0 {
+            tracing::info!(
+                token_deleted,
+                backend_action_deleted,
+                backend_state_deleted,
+                backend_deleted,
+                "Finished cleanup."
```
rolyatmax, line 528: Even if no backends were deleted, it seems useful to log this so the operator at least knows that it ran.

rolyatmax (approved):
One small comment, otherwise LGTM

PR #734, opened 2024-05-17
Clean up events

paulgb:
This keeps the event table from growing unbounded if `min_age_days` is set on the controller.


rolyatmax (approved):
LGTM

Something to maybe warn plane users about - "turning on" the `min_age_days` arg could result in some potentially very expensive queries and have performance / uptime implications for the DB.

PR #736, opened 2024-05-30
dont write to routes cache if None

rolyatmax:


rolyatmax:
Built the quickstart image and pulled into the Jamsocket dev CLI to test locally. Can confirm this fixes the issue.

PR #737, opened 2024-05-30
Make drones globally unique, removing dependance on cluster

michaelsilver:
As a step in doing the same change on the jamsocket side of things.

Caveat about database migration -- this isn't done in the "ideal" way:
* Code changes and migration changes are in the same PR. This is because it's hard to rework the "on-conflict" logic -- the constraint and on-conflict code are directly tied to one another, and it's invalid to have one without the other.
* The ideal way to add a constraint like this is to `ADD INDEX CONCURRENTLY` which can only be done outside of a transaction. SQLx does not currently support non-tx migrations as of their latest release (although the change landed in `main`). The result of doing the migration this way is the table will be locked up while the constraint is validated. The table is small so this shouldn't be an issue.


plane/src/database/node.rs
```diff
@@ -80,7 +80,7 @@ impl<'a> NodeDatabase<'a> {
                 last_connection_start_time
             )
             values ($1, $2, $3, $4, $5, $6, $7, now())
-            on conflict (cluster, name) do update set
+            on conflict (name) do update set
```
rolyatmax, line 83: When there's a conflict, do we need to set `cluster` now, too?

michaelsilver:
We observed that we have duplicate `name` values in the `node` table, and upon further inspection found that the culprit is `AcmeDnsServer` type nodes, which all have NULL `cluster` values.

Some helpful commands for reference:
```sql
select name, count(*) from node group by name having count(*) > 1;
```
and
```sql
WITH DuplicateNames AS (
    SELECT name
    FROM public.node
    GROUP BY name
    HAVING COUNT(*) > 1
)
SELECT n.*
FROM public.node n
JOIN DuplicateNames d ON n.name = d.name;
```

This boils down to a couple of things:
1. `cluster` is a nullable column
1. AcmeDnsServer node types are registered in the database [explicitly with a NULL cluster value](https://github.com/jamsocket/plane/blob/b792b34909c7d9ccff4cab16637c92eef95f377a/plane/src/controller/dns.rs#L21).

For now, we're going to rely on the fact that logically we expect drones to be globally unique. We see this is the case on the Jamsocket side.

In the future, we can address the tech debt here in the following way:
1. Make `cluster` not null
1. Update the AcmeDnsServer registration code to include the cluster. This would possibly require adding the cluster info to `MessageToDns`. 

At the moment we do not have helper tools in Plane to do data migrations, and fixing this would potentially require a data migration to remove duplicate nodes with null cluster values.

PR #738, opened 2024-05-31
allow requests to proxy to not have a trailing slash

rolyatmax:
Requests to `$CLUSTER/$CONNECTION_TOKEN` fail because they don't have a trailing slash. This PR fixes that by proxying to `/` anyway. (When using a static token, we don't strip the token from the path before proxying. In those cases, this proxies to `/$CONNECTION_TOKEN` (no trailing slash).

This is important because some servers (e.g. Nextjs) will redirect to remove a trailing slash. When this happens, the connection through proxy fails, returning a `400 Bad Request`.


michaelsilver (approved):
LGTM, can you leave a comment explaining why we're doing this?

PR #739, opened 2024-06-03
Kill all non-Terminated backends on drone start

michaelsilver:
Fixes the following bug:
* drone process restarts
* drone executor's `apply_action` receives a `BackendAction::Terminate` for a backend that's been running since before the restart
* the `self.backends` map does not contain the backend's manager anymore because the process was restarted
* the drone assumes the backend has been terminated and the backend continues running

Tested locally: running `docker ps` shows that all previously created backends are terminated upon drone agent restart.

Creating a unit test for this might involve a side quest because terminate currently hangs when the backend can't be found. This is a pre-existing issue outside of the scope of this bug, and I think the code here is sufficiently straightforward to merge without a dedicated test.

# Demo
In separate terminals:
```shell
$ ./dev/drone.sh
$ ./dev/proxy.sh
```

Spawn a bunch:
```shell
$ ./dev/cli.sh connect --cluster 'localhost:9090' --image 'ghcr.io/jamsocket/demo-image-drop-four'
$ ./dev/cli.sh connect --cluster 'localhost:9090' --image 'ghcr.io/jamsocket/demo-image-drop-four'
```

Check what's running in docker:
```shell
$ docker ps
```

Stop the drone process, then restart it, observe that backends get killed:
```
$ ./dev/drone.sh 
...
INFO plane::drone::executor: Terminating preexisting backends backends=[(BackendName("21p3dk3zvlm4jl"), Ready { address: Some(BackendAddr(127.0.0.1:53635)) })...
```

Verify all containers have been killed in docker:
```shell
$ docker ps
```

This is idempotent: stopping and restarting the drone multiple times works without issue.

### DIFF
--- a/plane/src/drone/executor.rs
+++ b/plane/src/drone/executor.rs
@@ -4,7 +4,7 @@ use crate::{
     names::BackendName,
     protocol::{BackendAction, BackendEventId, BackendStateMessage},
     types::BackendState,
-    util::GuardHandle,
+    util::{ExponentialBackoff, GuardHandle},
 };
 use anyhow::Result;
 use dashmap::DashMap;
@@ -24,8 +24,15 @@ pub struct Executor<R: Runtime> {
 }
 
 impl<R: Runtime> Executor<R> {
-    pub fn new(runtime: Arc<R>, state_store: StateStore, ip: IpAddr) -> Self {
+    pub async fn new(runtime: Arc<R>, state_store: StateStore, ip: IpAddr) -> Self {
         let backends: Arc<DashMap<BackendName, Arc<BackendManager<R>>>> = Arc::default();
+        let state_store = Arc::new(Mutex::new(state_store));
+
+        if let Err(err) =
+            Self::terminate_preexisting_backends(runtime.clone(), state_store.clone()).await
+        {
+            tracing::error!(?err, "Error terminating preexisting backends.");
+        }
 
         let backend_event_listener = {
             let docker = runtime.clone();
@@ -53,11 +60,55 @@ impl<R: Runtime> Executor<R> {
 
         Self {
             runtime,
-            state_store: Arc::new(Mutex::new(state_store)),
+            state_store,
             backends,
             ip,
             _backend_event_listener: backend_event_listener,
         }
+    }
+
+    // On restart, we want to terminate all existing backends and start fresh.
+    // This prevents bugs where an agent restart leaves the drone unable to
+    // terminate old backends.
+    async fn terminate_preexisting_backends(
+        runtime: Arc<R>,
+        state_store: Arc<Mutex<StateStore>>,
+    ) -> Result<()> {
+        let backends = state_store.lock().unwrap().active_backends()?;
+        let mut tasks = vec![];
+        for backend_id in backends {
+            let runtime_clone = runtime.clone();
+            tasks.push(tokio::spawn(async move {
+                let mut backoff = ExponentialBackoff::default();
+                let mut success = false;
+                for attempt in 1..=10 {
+                    match runtime_clone.terminate(&backend_id, true).await {
+                        Ok(()) => {
+                            success = true;
+                            break;
+                        }
+                        Err(err) => {
+                            tracing::error!(
+                                ?err,
+                                ?backend_id,
+                                ?attempt,
+                                "Attempt failed to terminate backend"
+                            );
+                            backoff.wait().await;
+                        }
+                    }
+                }
+                if !success {
+                    tracing::error!(?backend_id, "Failed to terminate backend after 10 attempts");
+                }
+            }));
+        }
+
+        for task in tasks {
+            task.await?
+        }
+
+        Ok(())
     }
 
     pub fn register_listener<F>(&self, listener: F) -> Result<()>
--- a/plane/src/drone/executor.rs
+++ b/plane/src/drone/executor.rs
@@ -3,7 +3,7 @@ use crate::{
     drone::runtime::Runtime,
     names::BackendName,
     protocol::{BackendAction, BackendEventId, BackendStateMessage},
-    types::BackendState,
+    types::{BackendState, TerminationKind, TerminationReason},
     util::{ExponentialBackoff, GuardHandle},
 };
 use anyhow::Result;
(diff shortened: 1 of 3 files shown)


michaelsilver:
Converting to draft: need to add logic to update the backend status once terminated so this operation it idempotent

plane/src/drone/executor.rs
```diff
@@ -53,13 +61,77 @@ impl<R: Runtime> Executor<R> {
...
+                            );
+                            backoff.wait().await;
+                        }
+                    }
+                }
+                if success {
+                    state_store
+                        .lock()
+                        .expect("State store lock poisoned.")
+                        .register_event(&backend_id, &state.to_terminated(Some(137)), Utc::now())
```
rolyatmax, line 115: I think we should call `to_terminating()` first with `TerminationKind::Hard` and `TerminationReason::KeyExpired`. Otherwise, those values won't be included in the final `Terminated` state.

rolyatmax: I'm not sure if we want to provide an exit code here that wasn't actually returned from the process. Do hard-terminated backends _always_ exit with a `137`?

plane/src/drone/state_store.rs
```diff
@@ -201,6 +201,31 @@ impl StateStore {
...
+        let mut rows = stmt.query([])?;
+        let mut active_backends = Vec::new();
+
+        while let Some(row) = rows.next()? {
+            let id: String = row.get(0)?;
+            let state_json: String = row.get(1)?;
+            let state: BackendState = serde_json::from_str(&state_json)?;
+
+            if !matches!(state, BackendState::Terminated { .. }) {
+                active_backends.push((BackendName::try_from(id)?, state));
```
rolyatmax, line 223: Re: the `?`s in the loop - I'd suggest that we log an error and `continue` the loop instead of dropping completely out of the function.

plane/src/drone/executor.rs
rolyatmax: Also, do you know if these state transitions end up getting persisted to the sqlite db?

plane/src/drone/executor.rs
```diff
@@ -53,13 +61,77 @@ impl<R: Runtime> Executor<R> {
...
+                let mut backoff = ExponentialBackoff::default();
+                let mut success = false;
+                for attempt in 1..=10 {
+                    match runtime_clone.terminate(&backend_id, true).await {
+                        Ok(()) => {
+                            success = true;
+                            break;
+                        }
+                        Err(err) => {
+                            tracing::error!(
```
rolyatmax, line 101: If the container is not found, I'm assuming that results in an `Err`? We should probably also count that as a success and update the backend's state to `Terminated`.

plane/src/drone/executor.rs
```diff
@@ -53,13 +61,77 @@ impl<R: Runtime> Executor<R> {
...
+            .lock()
+            .expect("State store lock poisoned.")
+            .active_backends()?;
+
+        tracing::info!(?backends, "Terminating preexisting backends");
+
+        let mut tasks = vec![];
+        let state_store_clone = state_store.clone();
+        for (backend_id, state) in backends {
+            let runtime_clone = runtime.clone();
```
rolyatmax, line 88: Nit: elsewhere when we `clone()` to avoid ownership bugs, we often just use the same var name. Doing so doesn't change the behavior here, does it?

plane/src/drone/executor.rs
paulgb: If we don't have an exit code for a backend, we should use `None`; that's why it's an option (hard-terminated backends don't have any exit code by definition)

plane/src/drone/executor.rs
paulgb: It doesn't change the behavior; it just shadows that name within the block. I like that approach because it prevents you from referring to the wrong clone by mistake (since you literally can't refer to it later in the block) and can move code in/out of the block without renaming the var.

paulgb:
My original intention for using sqlite as the state-of-truth for drone state is that if the drone process is restarted, we can continue the state machine for each backend where it left off. This was never fully implemented, but it should solve the same problems that this does, with the added advantage that if a drone process is restarted (e.g. it OOMs from a bug and Docker restarts it), running backends keep working as normal.

I'm curious if that approach was considered?

plane/src/drone/executor.rs
michaelsilver: Yep sorry just missed this one when switching styles, will fix

plane/src/drone/executor.rs
michaelsilver: Great feedback, addressing it all. Re: do these state transitions get persisted in the sqlite db -- yes, that's what  `state_store.reigster_event()` does.

michaelsilver:
> I'm curious if that approach was considered?

Discussed IRL; in summary, yes was considered extensively on Friday with Taylor and ruled out in favor of simply terminating.

plane/src/drone/executor.rs
```diff
@@ -24,8 +25,14 @@ pub struct Executor<R: Runtime> {
...
 impl<R: Runtime> Executor<R> {
-    pub fn new(runtime: Arc<R>, state_store: StateStore, ip: IpAddr) -> Self {
+    pub async fn new(runtime: Arc<R>, state_store: StateStore, ip: IpAddr) -> Self {
         let backends: Arc<DashMap<BackendName, Arc<BackendManager<R>>>> = Arc::default();
+        let state_store = Arc::new(Mutex::new(state_store));
+
+        Self::terminate_preexisting_backends(runtime.clone(), state_store.clone())
+            .await
+            .context("Failed to terminate all preexisting backends! Locks may be violated, Drone aborting startup.")
+            .unwrap();
```
michaelsilver, line 35: I'm ignoring this clippy comment -- the way I'm doing it with `.context().unwrap()` is a clean way to wrap the original error with additional context then panic. Open to other suggestions! Ultimately, I think Clippy simply doesn't understand the `.context()` and is simply reacting to the `.unwrap()`. A `.expect('bla')` by contrast will not include the original error.

paulgb: I think the "right thing" to do here would be to change the `Executor` interface to allow this to return a `Result`, but I'm fine with this for now until we know what `Err` type we would even want to use here.

I am surprised that `.expect()` doesn't include the error though, when I try it in the playground it does print the error: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=554e311a39e24d719c045ca0b6cd91db

plane/src/drone/executor.rs
```diff
@@ -53,13 +61,94 @@ impl<R: Runtime> Executor<R> {
...
+                            "Failed to register backend termination for backend {:?}",
+                            backend_id
+                        )
+                    });
+            }));
+        }
+
+        for task in tasks {
+            task.await?
+        }
```
paulgb, line 147: I think this will actually run the tasks in series rather than parallel. The reason for this is that tasks are not scheduled on the executor until they are awaited, so only one task will get executor time at a time. (This is admittedly kind of unintuitive!)

Instead, `futures_util::join_all(task)` should work here. ([ref](https://docs.rs/futures-util/0.3.21/futures_util/future/fn.join_all.html)). We already have `futures_util` as a dependency.

paulgb: Actually, @rolyatmax pointed out that the `tokio::spawn` here _does_ put the task on the executor, so I was wrong. That said, I think you could avoid the `tokio::spawn` call entirely by just putting the futures in the `tasks` vec and using `join_all`. I'll leave it up to you though.

michaelsilver:
@paulgb addressed your comments!

paulgb (approved):

PR #740, opened 2024-06-04
Don't overwrite CORS headers

paulgb:
If the response already has CORS headers set, we shouldn't overwrite them. This also tidies up how we do CORS header setting, by moving the headers into a const.

### DIFF
--- a/plane/src/proxy/proxy_service.rs
+++ b/plane/src/proxy/proxy_service.rs
@@ -29,10 +29,16 @@ use url::Url;
 
 const PLANE_BACKEND_ID_HEADER: &str = "x-plane-backend-id";
 
-const DEFAULT_CORS_HEADERS: &[(&str, str)] = &[
+const DEFAULT_CORS_HEADERS: &[(&str, &str)] = &[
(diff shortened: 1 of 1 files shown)


michaelsilver (approved):

rolyatmax (approved):

plane/src/proxy/proxy_service.rs
```diff
@@ -324,26 +337,15 @@ impl RequestHandler {
...
-                .parse()
-                .expect("Valid header value."),
-        );
-        headers.insert(
-            "Access-Control-Allow-Credentials",
-            "true".parse().expect("Valid header value."),
-        );
+
+        for (key, value) in DEFAULT_CORS_HEADERS {
+            if !headers.contains_key(*key) {
```
paulgb, line 342: Note that `key` here is received as an [`AsHeaderName`](https://docs.rs/http/latest/http/header/trait.AsHeaderName.html) implementation, so it is case-normalized before it is looked up in the `HeaderMap`. So we will do the right thing even if the header name from upstream has a different casing.

PR #741, opened 2024-06-04
Add docker network support to spawn config

paulgb:
This adds a way to pass a network name in a spawn config. This is mostly useful for development.


rolyatmax (approved):
This works

PR #743, opened 2024-06-11
Add Typed Unix Domain socket library

michaelsilver:
As part of our efforts to support executors that run in a separate process (see https://github.com/jamsocket/plane/issues/712), introduce a library for communication over unix domain sockets.

I wrote this library because I could not find any mature Rust library for sending typed messages over unix sockets. Luckily, most of the "heavy lifting" is done by stitching together existing mature libraries in the ecosystem.

The goals of this library were to support:
* Typed messages of any length, so long as they are serializable by `serde` (for this reason we chose streaming sockets over datagram sockets)
* Newline-delimited JSON as the wire protocol (for easy interop with other languages)
* Async code (and thus concurrent message sending/receiving) 
* The specific usecase of request/response messaging
* Ad-hoc messages client to server, or server to client
* Safe client or server restarts

Non goals of this library:
* Multiple clients connecting to a server
* High message throughput
* Unreliable transport over unix sockets (["as on most UNIX implementations, UNIX domain datagram sockets are always reliable"](https://man7.org/linux/man-pages/man7/unix.7.html))

For a peek into how I anticipate using this library to implement a generic executor, take a look at [this commit](https://github.com/jamsocket/plane/commit/f3443b447db1d3d7ff513d286161a9bed76d4614). This will be fully implemented in a subsequent PR. The general contour is all async functions like `spawn`, `terminate`, etc. will use `client.send_request()` (and await the response) and streaming functions will use `client.subscribe_events()`. See [runtime/mod.rs](https://github.com/jamsocket/plane/blob/2e2cbe4e8b9eebe2048452d657e2f0197e62a64b/plane/src/drone/runtime/mod.rs) for details. 

This was a fun project -- unix streaming sockets simply send binary over the wire and are very low level primitive, so achieving correct concurrency control and typing was nontrivial! Chapter 57 of _The Linux Programming Interface_ is a good brief primer on unix sockets for anyone unfamiliar. 

Note to reviewers: this is my first project leveraging Tokio directly, so any and all feedback welcome, and apologies if anything is unconventional!


plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,219 @@
...
+    response_map: Arc<DashMap<Uuid, oneshot::Sender<ResponseType>>>,
+    event_tx: broadcast::Sender<ServerMessageType>,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync>>
+where
+    ClientMessageType:
+        Send + Sync + 'static + Clone + Debug + Serialize + for<'de> Deserialize<'de>,
+    ServerMessageType:
+        Send + Sync + 'static + Clone + Debug + Serialize + for<'de> Deserialize<'de>,
+    RequestType: Send + Sync + 'static + Clone + Debug + Serialize + for<'de> Deserialize<'de>,
+    ResponseType: Send + Sync + 'static + Clone + Debug + Serialize + for<'de> Deserialize<'de>,
```
paulgb, line 133: I gather that `ClientMessageType` and `ServerMessageType` are for "event" messages and `RequestType` / `ResponseType` are for request/response messages. I am inclined to collapse them all into `ClientMessageType` and `ServerMessageType`, and wrap them all in a struct like `IDedMessage` except that makes the ID optional instead of required.

You would still have `send_message` and `send_request`, where `send_message` would give an ID of `None` and `send_request` would set an ID.

My thinking here is that presumably there will be different types of requests/responses, and there's no way at the rust type / serde level to ensure that the response we get back is actually the type we expect given the request we sent, so the code that handles a `ResponseType` response on the client side still needs to extract the correct type of response.

Given that we need to have that complexity live outside of the `TypedUnixSocket` code anyway, I think this would simplify the code here quite a bit.

paulgb (commented):
I'll dig in some more tomorrow, but left a few comments for now. So far I like the approach and decisions you've made.

plane/src/typed_unix_socket/server.rs
```diff
@@ -0,0 +1,220 @@
...
+                        tracing::error!("Error reading line: {}", e);
+                    }
+                }
+            }
+        }
+    });
+
+    // Task to handle sending responses
+    let send_task = tokio::spawn(async move {
+        loop {
```
paulgb, line 190: I think this loop never stops, even if the connection has reconnected?

I am tempted to just break from the loop if we encounter an error; reconnecting in that case doesn't sound like a bad idea even if it may have been recoverable.

plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,209 @@
...
+                        tracing::error!("Error reading line: {}", e);
+                    }
+                }
+            }
+        }
+    });
+
+    // Task to handle sending messages
+    let send_task = tokio::spawn(async move {
+        loop {
```
paulgb, line 179: I don't think this loop ever terminates, even after a reconnect (left a similar comment on the server with a suggestion)

paulgb (commented):
A couple of comments. I have a good grasp on this code now, I think it's pretty close to ready.

michaelsilver:
@paulgb I believe I've addressed all your concerns, take another look and happy to chat in office tomorrow!

plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,247 @@
...
+    event_tx: broadcast::Sender<MessageToClient>,
+    shutdown_tx: watch::Sender<()>,
+}
+
+impl<MessageToServer, MessageToClient> TypedUnixSocketClient<MessageToServer, MessageToClient>
+where
+    MessageToServer: Send + Sync + 'static + Clone + Debug + Serialize + for<'de> Deserialize<'de>,
+    MessageToClient: Send + Sync + 'static + Clone + Debug + Serialize + for<'de> Deserialize<'de>,
+{
+    pub async fn new<P: AsRef<Path>>(socket_path: P) -> Result<Self, Error> {
```
rolyatmax, line 31: Looks like this function only returns `Ok`s. Do we need the `Result` type?

plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,247 @@
...
+                            )
+                            .await
+                            .unwrap_or_else(|e| {
+                                tracing::error!("Error handling connection: {}", e);
+                            });
+                            break;
+                        }
+                        Err(e) => {
+                            tracing::error!("Error connecting to server: {}", e);
+                            backoff.wait().await;
```
rolyatmax, line 63: What happens if it never connects? This loop will never break, and these values will never be dropped. Maybe the `backoff` could wait for a maximum amount of time, after which it returns an `Err`?

plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,247 @@
...
+            message: request,
+        };
+
+        self.response_map.insert(id, response_tx);
+
+        // Wait until there is at least one subscriber
+        let mut backoff = get_quick_backoff();
+        while self.tx.receiver_count() == 0 {
+            tracing::info!("Waiting for a subscriber...");
+            backoff.wait().await;
```
rolyatmax, line 92: Another spot where we might want to return an error after some period of time, so this doesn't loop forever.

plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,247 @@
...
+        let response_map = Arc::clone(&response_map);
+        let mut shutdown_rx = shutdown_rx.clone();
+        async move {
+            loop {
+                tokio::select! {
+                    _ = shutdown_rx.changed() => {
+                        tracing::info!("Shutting down receive task");
+                        break;
+                    }
+                    result = lines.next_line() => {
```
rolyatmax, line 160: Should maybe have a timeout? Feel free to ignore if y'all have already considered this and decided against it.

plane/src/typed_unix_socket/client.rs
```diff
@@ -0,0 +1,247 @@
...
+                                    }
+                                    WrappedMessage { id: None, message } => {
+                                        if let Err(e) = event_tx.send(message) {
+                                            tracing::error!("Error sending event: {}", e);
+                                        }
+                                    }
+                                }
+                            }
+                            Ok(None) => {
+                                tracing::error!("Connection closed by server");
```
rolyatmax, line 192: Should this `break`?

paulgb: +1

Also, if you break here, I don't think you need the whole `shutdown` part, it seems like you are just reimplementing the connection closure mechanism that already exists on a channel.

plane/src/typed_unix_socket/server.rs
```diff
@@ -0,0 +1,263 @@
...
+        let _ = self.shutdown_server_tx.send(());
+
+        // Signal the handler to shut down
+        let _ = self.shutdown_handler_tx.send(());
+
+        if let Err(e) = fs::remove_file(&self.socket_path) {
+            tracing::warn!("Failed to remove socket file: {}", e);
+        }
+    }
+}
```
paulgb, line 154: The drop/clone semantics here are inconsistent, creating the footgun you encountered while testing. In general I would expect clone/drop to behave such that the code in the block is essentially a no-op as far as `a` is concerned:

```rust
let a = Something::new();

{
    let b = a.clone()
}
```

paulgb (commented):
I'm going to have another pass at this but a few of these comments are actionable now so I wanted to get them out.

michaelsilver:
Ready for a final review! 
* Removed unnecessary retry loops
* Added timeouts to remaining retry loops
* In service of testing: changed drop to shutdown, use cancellation tokens
* Otherwise server/client will shutdown due to their peer disconnecting, and that will be handled by tokio. No more channels everywhere!

plane/src/typed_unix_socket/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        let mut handles = vec![];
+        for i in 0..10 {
+            let client = client.clone();
+            handles.push(spawn(async move {
+                let request = format!("Request {}", i);
+                let response = client.send_request(request).await.unwrap();
+                assert_eq!(response, format!("Response to Request {}", i));
+            }));
+        }
```
paulgb, line 151: I don't think this is a meaningfully different test from this:

```rust
for i in 0..10 {
    let request = format!("Request {}", i);
    let response = client.send_request(request).await.unwrap();
    assert_eq!(response, format!("Response to Request {}", i));
}
```

_Technically_, the runtime is allowed (as a matter of what it guarantees) to poll each task in arbitrary order, but it won't do that in practice, which means that we aren't actually testing any concurrency with this, and it could give us a false sense of confidence.

If you want to test that everything works if responses come out of order, consider having the server sleep, i.e something like this:

```rust
spawn(async move {
    let mut request_rx = server.subscribe_requests();
    while let Ok(request) = request_rx.recv().await {
        let req_num: i32 = request.parse().unwrap();
        tokio::sleep(Duration::from_millis(req_num * 100).await;
        let response = format!("Response to {}", request.message);
        server.send_response(&request, response).await.unwrap();
    }
});

let mut handles = vec![];
for i in 0..10 {
    let request = format!("{}", i);
    handles.push(async move {
        client.send_request(request).await.unwrap();
    });
}

let result = futures::future::join_all(handles);

for i in 0..10 {
    assert_eq!(result[i], format!("Response to Request {}", i));
}
```

paulgb: I tried this out in #748. If it LGTY, let's do the same for other concurrency tests, and then we are good to merge.

michaelsilver: Awesome, just merged that in and I'll take care of adding this for the rest -- thanks for catching this!

plane/src/typed_unix_socket/mod.rs
```diff
@@ -15,10 +15,10 @@ pub struct WrappedMessage<T> {
 
 fn get_quick_backoff() -> ExponentialBackoff {
     ExponentialBackoff::new(
-        Duration::try_milliseconds(10).expect("duration is always valid"),
-        Duration::try_milliseconds(100).expect("duration is always valid"),
+        Duration::milliseconds(10),
```
paulgb, line 18: TIL that `Duration::milliseconds` was [un-deprecated](https://github.com/chronotope/chrono/releases/tag/v0.4.36), nice!

plane/src/typed_unix_socket/mod.rs
```diff
@@ -153,7 +150,7 @@ mod tests {
             handles.push(async move { client.send_request(request).await.unwrap() });
         }
 
-        let result = futures::future::join_all(handles).await;
+        let result = futures_util::future::join_all(handles).await;
```
paulgb, line 153: ah cool we didn't need the extra dep after all!

plane/src/typed_unix_socket/mod.rs
```diff
@@ -30,10 +30,7 @@ mod tests {
     use futures_util::future::join_all;
     use std::collections::HashSet;
     use std::env;
-    use std::ops::Deref;
     use std::path::PathBuf;
-    use std::sync::Arc;
-    use std::time::Duration;
     use tokio::spawn;
```
paulgb, line 34: nit: since you've added some `tokio::` to the spawns below, maybe do it for all of them for consistency? I'm not super opinionated on whether we use `tokio::spawn` or `spawn` as long as it's consistent within a file, but I lean towards `tokio::spawn` everywhere for clarity.

paulgb (approved):
Looks good to me! One actionable comment.

PR #744, opened 2024-06-12
Suggested changes on top of #743

paulgb:


plane/src/typed_unix_socket/mod.rs
```diff
@@ -1,30 +1,17 @@
...
 pub mod client;
 pub mod server;
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct IDedMessage<T> {
-    id: Uuid,
+pub struct WrappedMessage<T> {
+    id: Option<String>,
     message: T,
 }
```
paulgb, line 13: The semantics here are: if ID is given, this is a request/response; otherwise this is an event.

This replaces both `IDedMessage` and `Wrapped{Client,Server}MessageType`.

plane/src/typed_unix_socket/mod.rs
```diff
@@ -38,25 +25,25 @@ fn get_quick_backoff() -> ExponentialBackoff {
...
     use futures_util::future::join_all;
     use std::collections::HashSet;
     use std::env;
     use std::path::PathBuf;
     use tokio::spawn;
-    use uuid::Uuid;
 
     fn create_temp_socket_path() -> PathBuf {
-        env::temp_dir().join(format!("test_socket_{}", Uuid::new_v4())) // ensure random file name
+        env::temp_dir().join(format!("test_socket_{}", random_string())) // ensure random file name
```
paulgb, line 36: `random_string()` over `random_token()` is important here because a socket path can only be max 108 bytes ([ref](https://man7.org/linux/man-pages/man7/unix.7.html)) and it is borderline with `random_token()` which is longer.

Fortunately, there is absolutely zero chance of collision here because paths are already scoped by test.

paulgb (commented):
Add some inline comments.

plane/src/typed_unix_socket/client.rs
```diff
@@ -9,32 +10,22 @@ use tokio::{
     net::UnixStream,
     sync::{broadcast, oneshot},
 };
-use uuid::Uuid;
 
 #[derive(Clone)]
-pub struct TypedUnixSocketClient<ClientMessageType, ServerMessageType, RequestType, ResponseType>
+pub struct TypedUnixSocketClient<MessageToServer, MessageToClient>
```
paulgb, line 15: I find `{Client,Server}MessageType` ambiguous because it's not clear whether a client message is a message sent by or to the client. I like the convention `MessageToX` or `MessageFromX` to make the distinction obvious.

plane/src/typed_unix_socket/server.rs
```diff
@@ -80,43 +70,39 @@ where
...
-        request: IDedMessage<RequestType>,
-        response: ResponseType,
-    ) -> Result<(), Error> {
-        let response_msg =
-            WrappedServerMessageType::<ResponseType, ServerMessageType>::Response(IDedMessage {
-                id: request.id,
-                message: response,
-            });
-
+    pub async fn send_response(&self, id: String, response: MessageToClient) -> Result<(), Error> {
```
michaelsilver, line 84: One thing I was intentional about with the API design is I'd like the details of the ID to be as hidden as possible from the calling code. As such, it'd be nice if `send_response` could accept the request and response like before, and internally take care of extracting/reusing the ID.

plane/src/typed_unix_socket/mod.rs
michaelsilver: > The semantics here are: if ID is given, this is a request/response; otherwise this is an event.

Nit: can we add a comment about this?

plane/src/typed_unix_socket/mod.rs
michaelsilver: Nit: seems like a good idea to put something like this in a comment.

plane/src/typed_unix_socket/mod.rs
```diff
@@ -65,7 +52,10 @@ mod tests {
             let mut request_rx = server.subscribe_requests();
             while let Ok(request) = request_rx.recv().await {
                 let response = "Hello, client!".to_string();
-                server.send_response(request, response).await.unwrap();
+                server
+                    .send_response(request.id.unwrap(), response)
```
michaelsilver, line 56: (This function call will be cleaner once we change `send_response` to simply take `(request, response)`)

plane/src/lib.rs
```diff
@@ -24,6 +24,7 @@ pub mod proxy;
 pub mod serialization;
 pub mod signals;
 pub mod typed_socket;
+#[allow(unused)] // for now, to disable clippy noise
```
michaelsilver, line 27: Good call

plane/src/typed_unix_socket/mod.rs
```diff
@@ -1,30 +1,17 @@
...
 use crate::util::ExponentialBackoff;
 
 pub mod client;
 pub mod server;
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
-pub struct IDedMessage<T> {
-    id: Uuid,
+pub struct WrappedMessage<T> {
+    id: Option<String>,
```
michaelsilver, line 11: Minor nit: I'm borderline on this one, but thoughts on creating a type (even if it's just a `String` alias) for the ID?

michaelsilver: It's less important once we change the signature of `send_response` but that was largely what motivated this suggestion -- something like `RandomID` is more meaningful of a type than `String`.

michaelsilver (approved):
Nice, appreciate it! Left a few small comments -- up to you if you want to address and merge or merge and leave it to me to tweak. Either way I want to unblock your larger review of the library so feel free to merge without requesting another review.

plane/src/typed_unix_socket/mod.rs
paulgb: My general preference is for functions to take just the data they need to do what they need to do. I find that this makes them easier to test and more versatile when refactoring, and the actual data flow is more accurately represented by the function signatures.

I'm inclined to keep this to ID, but use a newtype as you suggested to take a `MessageId` or similar as the first param instead of a `String`.

paulgb: Just I read your rationale below and I see that side of it too. I do like the idea of hiding the implementation detail of the ID, so I'm fine with that. It does mean that if we want to use the contents of the original message we will need to clone it though, since we won't be able to pass the request to `send_response` if we have already taken the message field from it.

michaelsilver (approved):

plane/src/typed_unix_socket/mod.rs
michaelsilver: Sounds good, I'll pick one of these options and finalize it in the parent PR!

PR #745, opened 2024-06-14
Stop metrics loop when backend is terminated (DIS-2123)

paulgb:


rolyatmax (approved):
LGTM!

As we talked about offline, there's potential to use this same code to solve this problem, too: https://linear.app/drifting/issue/DIS-2169/fix-backend-not-found-when-terminating-error-loops-on-drones

PR #746, opened 2024-06-14
Mark backend as terminated when lost

paulgb:
In the weird case where the Plane controller asks a drone to terminate a backend that has already been removed from Docker, we currently do nothing but don't do anything to tell the controller to mark the backend as lost. This changes the behavior to do that, via the state store.

### DIFF
--- a/plane/src/drone/executor.rs
+++ b/plane/src/drone/executor.rs
@@ -107,7 +107,7 @@ impl<R: Runtime> Executor<R> {
                 let mut success = false;
                 for attempt in 1..=10 {
                     match runtime.terminate(&backend_id, true).await {
-                        Ok(()) => {
+                        Ok(_) => {
                             success = true;
                             break;
                         }
@@ -211,6 +211,9 @@ impl<R: Runtime> Executor<R> {
                     // else we can deadlock.
                     let Some(manager) = self.backends.get(backend_id) else {
                         tracing::warn!(backend_id = backend_id.as_value(), "Backend not found when handling terminate action (assumed terminated).");
+
+                        // Terminate will only return an error on a docker error, not if the backend is already terminated.
+                        self.runtime.terminate(backend_id, true).await?;
 
                         self.state_store
                             .lock()
--- a/plane/src/drone/executor.rs
+++ b/plane/src/drone/executor.rs
@@ -3,7 +3,7 @@ use crate::{
     drone::runtime::Runtime,
     names::BackendName,
     protocol::{BackendAction, BackendEventId, BackendStateMessage},
-    types::{BackendState, TerminationKind, TerminationReason},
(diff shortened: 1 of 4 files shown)


plane/src/drone/executor.rs
```diff
@@ -211,6 +211,21 @@ impl<R: Runtime> Executor<R> {
                     // else we can deadlock.
                     let Some(manager) = self.backends.get(backend_id) else {
                         tracing::warn!(backend_id = backend_id.as_value(), "Backend not found when handling terminate action (assumed terminated).");
+
+                        self.state_store
```
rolyatmax, line 215: Here we're just assuming the backend is no longer running if the `backend_id` is not in `self.backends`. This isn't a _terrible_ assumption, since the drone is supposed to look for and terminate any backends that were stored in the sqlite table. But there are a few spots where this could break (if a backend doesn't get written to the table before the process dies or if the restarted drone agent was unable to send a termination message to the container). 

If it's possible, I think it would be better if we could determine for sure if a container is no longer running (by inspecting and seeing if docker says it's still running) before we assert that it is `Terminated`. Is that possible?

paulgb: Good point. I took the approach of just _always_ attempting to terminate the backend, and changing the semantics of the runtime to not error if a backend was already terminated or doesn't exist.

plane/src/drone/runtime/docker/mod.rs
```diff
@@ -230,10 +230,10 @@ impl Runtime for DockerRuntime {
                 "Container not found, assuming it was already terminated."
             );
         } else {
-            return result.map_err(|e| e.into());
+            return Ok(false);
```
rolyatmax, line 233: From offline convo: this should stay the same and the `return Ok(false)` should move into the prev block

plane/src/drone/executor.rs
```diff
@@ -212,6 +212,9 @@ impl<R: Runtime> Executor<R> {
                     let Some(manager) = self.backends.get(backend_id) else {
                         tracing::warn!(backend_id = backend_id.as_value(), "Backend not found when handling terminate action (assumed terminated).");
 
+                        // Terminate will only return an error on a docker error, not if the backend is already terminated.
+                        self.runtime.terminate(backend_id, true).await?;
```
rolyatmax, line 216: Do we want to log anything based on whether this returned `Ok(true)` or `Ok(false)`?

rolyatmax (approved):
Once the below changes that we talked through in person are made this is good to land

plane/src/drone/executor.rs
paulgb: The `Ok(false)` case already logs a warning for the Docker runtime, I think that will be sufficient for now.

PR #748, opened 2024-06-17
Change concurrency test of #743

paulgb:
This refactors the `concurrent_requests_responses` test to force requests to be fulfilled out-of-order.


michaelsilver (approved):
Only comment is we don't need `arc` but I'll take care of removing that, thanks!

PR #755, opened 2024-06-25
Create a BackendStatus -> numeric mapping

paulgb:
The backend status lifecycle has a strict ordering. Backends can move forward in the lifecycle, but not backwards.

On the Rust side, `BackendStatus` is `PartialOrd`, so we can do natural comparisons between backend statuses as they relate to this order. However, on the database side, they are just strings.

We would like to add constraints at the database level, such as "a backend status can never go backwards". In the database, statuses are represented as strings. We could re-implement the comparison logic in the database as a function over strings, but I don't love having the logic live in both Rust and the database.

The alternative implemented here is to create a function from `BackendStatus` -> `i32` which produces a number that reflects the natural order of statuses.

No Plane behavior is changed in this PR. The next step is to do a migration that stores these numbers alongside the status.

### DIFF
--- a/plane/src/types/backend_state.rs
+++ b/plane/src/types/backend_state.rs
@@ -40,6 +40,8 @@ impl BackendStatus {
     /// Returns an integer representation of the status. This is meant to be an
     /// opaque value that can be used for determining if one status comes before
     /// another in the backend lifecycle.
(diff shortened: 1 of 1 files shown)


rolyatmax (approved):
LGTM!

PR #758, opened 2024-06-25
Populate and enforce last_status_number column of backend

paulgb:
This bails from updating the backend status if the existing status is greater or equal to the new status.

If the existing `last_status_number` is `NULL`, this will always succeed.


plane/src/database/backend.rs
```diff
@@ -181,29 +181,53 @@ impl<'a> BackendDatabase<'a> {
...
         state: BackendState,
-    ) -> sqlx::Result<()> {
+    ) -> sqlx::Result<bool> {
         let mut txn = self.db.pool.begin().await?;
 
         emit_with_key(&mut *txn, &backend.to_string(), &state).await?;
 
-        sqlx::query!(
+        let last_status = state.status();
+        let last_status_number = last_status.as_int();
```
rolyatmax, line 190: nit: can we call these vars `new_status` and `new_status_number` or `next_status`?

plane/src/database/backend.rs
```diff
@@ -181,29 +181,53 @@ impl<'a> BackendDatabase<'a> {
...
+            last_status.to_string(),
+            last_status_number,
             state.address().map(|d| d.0.to_string()),
             serde_json::to_value(&state).expect("BackendState should always be JSON-serializable."),
         )
         .execute(&mut *txn)
         .await?;
 
+        if result.rows_affected() == 0 {
+            let current_status = sqlx::query!(
```
rolyatmax, line 214: nit: can we call this `result`?

plane/src/database/backend.rs
```diff
@@ -181,29 +181,53 @@ impl<'a> BackendDatabase<'a> {
...
+                select last_status
+                from backend
+                where id = $1
+                "#,
+                backend.to_string(),
+            )
+            .fetch_optional(&mut *txn)
+            .await?;
+
+            let current_status = current_status.map(|r| r.last_status);
```
rolyatmax, line 225: nit: can we all this `last_status`?

rolyatmax (approved):
Code looks good - sorry for the nits, but it looks like the terms `last_status`, `current_status`, and `new_status` are being used in slightly confusing ways. Mind if we stick to the two terms `last_status` and `new_status` (or `next_status`) while in that function?

plane/src/database/backend.rs
```diff
@@ -181,29 +181,53 @@ impl<'a> BackendDatabase<'a> {
...
+
+            tracing::warn!(current_status, new_status=%last_status, "Not updating backend status");
```
rolyatmax, line 227: Can we add the backend_id here too?

plane/src/database/backend.rs
```diff
@@ -180,43 +180,69 @@ impl<'a> BackendDatabase<'a> {
...
+                where id = $1
+                "#,
+                backend.to_string(),
+            )
+            .fetch_optional(&mut *txn)
+            .await?;
+
+            let last_status = result.map(|r| r.last_status);
+
+            tracing::warn!(last_status, new_status=%new_status, "Not updating backend status");
```
rolyatmax, line 228: Can we add `backend` name here?

plane/src/database/connect.rs
```diff
@@ -156,6 +161,17 @@ async fn create_backend_with_key(
         }
     };
 
+    sqlx::query!(
+        r#"
+        insert into backend_state (backend_id, state, created_at)
```
rolyatmax, line 166: I think we should use the same timestamp here as we use for `last_status_time` above

paulgb: Because they are both in the same transaction, they both refer to the same timestamp ([doc](https://www.postgresql.org/docs/9.3/functions-datetime.html#FUNCTIONS-DATETIME-CURRENT))

> `now()` is a traditional PostgreSQL equivalent to `transaction_timestamp()`.

rolyatmax: TIL

rolyatmax (approved):

PR #761, opened 2024-06-25
Emit scheduled event in connect transaction

paulgb:
This refactors the code that:
- updates the `backend_state` table
- emits the `BackendState` event

Into one helper function that takes a transaction to execute in. This code path is then used both when updating state on the drone, and when creating the initial `Scheduled` state on the server.

The `last_status_time`, `backend_state.created_at`, and the event's timestamp will all be the time of the transaction start.


plane/src/database/subscribe.rs
```diff
@@ -352,11 +352,11 @@ impl EventSubscriptionManager {
     }
 }
 
-pub async fn emit_impl<'c, T, E>(db: E, key: Option<&str>, payload: &T) -> Result<(), sqlx::Error>
-where
-    T: NotificationPayload,
-    E: PgExecutor<'c>,
-{
+pub async fn emit_impl<T: NotificationPayload>(
```
paulgb, line 355: These changes were made following [this comment](https://github.com/launchbadge/sqlx/discussions/1136#discussioncomment-566896) about using sqlx.

plane/src/database/backend.rs
```diff
@@ -435,7 +424,10 @@ impl<'a> BackendDatabase<'a> {
     }
 
     pub async fn publish_metrics(&self, metrics: BackendMetricsMessage) -> sqlx::Result<()> {
-        emit_ephemeral_with_key(&self.db.pool, &metrics.backend_id.to_string(), &metrics).await
+        let mut txn = self.db.pool.begin().await?;
+        emit_ephemeral_with_key(&mut txn, &metrics.backend_id.to_string(), &metrics).await?;
+        txn.commit().await?;
+        Ok(())
```
paulgb, line 430: The `emit_*` functions now take a `PgConnection`. The easiest way to get that from a pool is to create a transaction.

plane/src/database/backend.rs
```diff
@@ -585,3 +577,25 @@ impl BackendRow {
...
+        insert into backend_state (backend_id, state)
+        values ($1, $2)
+        "#,
+        backend.to_string(),
+        serde_json::to_value(&new_state).expect("BackendState should always be JSON-serializable."),
+    )
+    .execute(&mut *txn)
+    .await?;
+
+    emit_with_key(txn, &backend.to_string(), new_state).await?;
```
paulgb, line 598: note: this is what forced the change from using a generic sqlx `Executor` to passing a `&mut PgConnection`. If we have an `Executor`, once we pass it to `execute(...)` we have moved it, so we can't use it for `emit_with_key`. Changing everything to references fixes this.

rolyatmax (approved):

PR #766, opened 2024-06-26
Socket based executor client

michaelsilver:
No logic, just sends/receives stuff over a unix socket (as a client) and expects a server to do all the thinking.

Warning: completely untested -- was going to continue with implementing the server side and see how that goes.

### DIFF
--- a/plane/src/drone/runtime/unix_socket/mod.rs
+++ b/plane/src/drone/runtime/unix_socket/mod.rs
@@ -1,0 +1,145 @@
+use super::{
+    docker::{SpawnResult, TerminateEvent},
+    Runtime,
+};
+use crate::{
+    database::backend::BackendMetricsMessage,
+    names::BackendName,
+    protocol::AcquiredKey,
+    typed_unix_socket::client::TypedUnixSocketClient,
+    types::{backend_state::BackendError, BearerToken, DockerExecutorConfig},
+};
+use anyhow::{Error, Result};
+use serde::{Deserialize, Serialize};
+use std::net::SocketAddr;
+use tokio_stream::{Stream, StreamExt};
+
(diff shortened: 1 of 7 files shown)


plane/src/drone/runtime/unix_socket/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+        address: SocketAddr,
+    ) -> Result<(), BackendError> {
+        let response = self
+            .client
+            .send_request(MessageToServer::WaitForBackend(backend.clone(), address))
+            .await
+            .expect("Failed to send request");
+        match response {
+            MessageToClient::WaitForBackendResult(Ok(())) => Ok(()),
+            MessageToClient::WaitForBackendResult(Err(e)) => Err(e),
```
paulgb, line 134: If this is intentional to be explicit I'm fine with it, but why not:

```rust
MessageToClient::WaitForBackendResult(v) => v,
```

to cover both cases?

plane/src/drone/runtime/unix_socket/mod.rs
```diff
@@ -0,0 +1,145 @@
...
+                backend_id.clone(),
+                executable,
+                acquired_key.cloned(),
+                static_token.cloned(),
+            ))
+            .await?;
+        match response {
+            MessageToClient::SpawnResult(Ok(result)) => Ok(result),
+            MessageToClient::SpawnResult(Err(e)) => Err(Error::msg(e)),
+            _ => unreachable!("Unexpected response from server"),
```
paulgb, line 84: we should just warn on these rather than panicking

paulgb (approved):
lgtm aside from a couple comments!

paulgb (approved):

PR #768, opened 2024-06-26
query for drones from drone pool with a duration

rolyatmax:
Updates the `get_drones_for_pool()` to take a duration instead of only returning drones from the last 45 seconds. This flexibility is useful when wanting to see all the drones that ran over a longer period of time.

### DIFF
--- a/plane/src/database/drone.rs
+++ b/plane/src/database/drone.rs
@@ -95,6 +95,8 @@ impl<'a> DroneDatabase<'a> {
         Ok(result.pool.into())
     }
 
+    /// Returns a list of drones that are ready and have been seen in the last `seen_in_last` duration.
+    /// Limits the result to the most recent 100 drones.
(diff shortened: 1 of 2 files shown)


plane/src/database/drone.rs
```diff
@@ -125,11 +129,11 @@ impl<'a> DroneDatabase<'a> {
                 and pool = $3
                 and last_local_time is not null
                 and last_connection_start_time is not null
-            order by drone.id
+            order by drone.id desc
+            limit 100
```
paulgb, line 133: nit: since this is hard-coded, maybe add a comment to the function that mentions that the first 100 by ID are returned?

paulgb (approved):
LGTM, one nit

plane/src/database/drone.rs
rolyatmax: Good call

PR #769, opened 2024-06-27
Fix proxy panic

paulgb:
We have occasionally seen a panic in the proxy where the receiver falls behind. This fixes that by replacing the broadcast channel with a watch channel, since we don't need to keep multiple values around anyway.

### DIFF
--- a/plane/src/proxy/proxy_service.rs
+++ b/plane/src/proxy/proxy_service.rs
@@ -203,9 +203,17 @@ impl RequestHandler {
                 return Err(ProxyError::MissingHostHeader);
             };
 
+            let host = match host.parse() {
+                Ok(host) => host,
+                Err(err) => {
+                    tracing::warn!(?err, ?host, "Invalid host header.");
+                    return Err(ProxyError::BadRequest);
+                }
+            };
+
             let mut uri_parts = req.uri().clone().into_parts();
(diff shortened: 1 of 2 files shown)


rolyatmax (approved):

rolyatmax:
Looks like there are a bunch of panics in the proxy_service too:

https://app.datadoghq.com/logs?query=service%3Aplane%20host%3Apx-%2A%20-status%3Ainfo%20%22panicked%22%20&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1719333112485&to_ts=1719505912485&live=true

paulgb:
> Looks like there are a bunch of panics in the proxy_service too:
> 
> https://app.datadoghq.com/logs?query=service%3Aplane%20host%3Apx-%2A%20-status%3Ainfo%20%22panicked%22%20&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1719333112485&to_ts=1719505912485&live=true

Good catch, added a fix for that too.

PR #772, opened 2024-07-02
Proxy waits on (scheduled/loading/starting) backend to become ready

paulgb:
Follows design doc here: https://docs.google.com/document/d/1KpZMyz8E4a7pk_OaIqQKLWWKiLo4SdRhe1MkUy6EyLE/edit

The general approach:
- Introduces [`PartialRouteInfo`](https://github.com/jamsocket/plane/pull/772/files#diff-0abc0d77e47c302d2447792802d4fb6300e2ee6ba14b99cb00016febdc5109e3R639-R660), which is a `RouteInfo` without an address. The address is not known until the scheduled drone sets it when the backend is in the `Starting` state, but everything else in the `RouteInfo` is _always_ available on a valid (non-expired) connection token. A `PartialRouteInfo` [can be upgraded](https://github.com/jamsocket/plane/pull/772/files#diff-0abc0d77e47c302d2447792802d4fb6300e2ee6ba14b99cb00016febdc5109e3R649-R659) to a regular `RouteInfo` once the address is known.
- Introduces [`RouteInfoResult`](https://github.com/jamsocket/plane/pull/772/files#diff-0abc0d77e47c302d2447792802d4fb6300e2ee6ba14b99cb00016febdc5109e3R633-R637) which is an enum corresponding to the three possible states of a backend:
  1. Not found (connection token expired or never existed)
  2. Waiting (corresponds to scheduled/loading/starting/waiting states)
  3. Ready
- Refactors [`route_info_for_token`](https://github.com/jamsocket/plane/pull/772/files#diff-0abc0d77e47c302d2447792802d4fb6300e2ee6ba14b99cb00016febdc5109e3R362) (and relatedly [`route_info_for_static_token`](https://github.com/jamsocket/plane/pull/772/files#diff-0abc0d77e47c302d2447792802d4fb6300e2ee6ba14b99cb00016febdc5109e3R293) to return a `RouteInfoResult`. This also means modifying the query to return a result even if the backend is not ready.

In the controller:
- Adds [`handle_route_info_request`](https://github.com/jamsocket/plane/pull/772/files#diff-79359c831293eaad21a8f2cf49450fc80fa43b0888bb741868703fce8c0d2e80R25), which uses `route_info_for_token` and then, if the result is a `RouteInfoResult::Waiting`, [spawns a tokio task to listen for backend changes](https://github.com/jamsocket/plane/pull/772/files#diff-79359c831293eaad21a8f2cf49450fc80fa43b0888bb741868703fce8c0d2e80R73).

### DIFF
--- a/plane/src/controller/proxy.rs
+++ b/plane/src/controller/proxy.rs
@@ -1,14 +1,17 @@
 use super::{error::ApiErrorKind, Controller};
 use crate::{
     controller::error::IntoApiError,
-    database::subscribe::{Notification, Subscription},
-    names::BackendName,
+    database::{
+        backend::RouteInfoResult,
+        subscribe::{Notification, Subscription},
+    },
+    names::{BackendName, Name},
     protocol::{
         CertManagerRequest, CertManagerResponse, MessageFromProxy, MessageToProxy,
         RouteInfoRequest, RouteInfoResponse,
     },
     typed_socket::{server::new_server, TypedSocket},
-    types::{BackendState, ClusterName, NodeId},
+    types::{BackendState, BearerToken, ClusterName, NodeId},
 };
 use axum::{
     extract::{ws::WebSocket, ConnectInfo, Path, State, WebSocketUpgrade},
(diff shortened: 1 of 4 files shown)


plane/src/controller/proxy.rs
```diff
@@ -19,31 +22,135 @@ use std::net::{IpAddr, SocketAddr};
...
+            let mut sub: Subscription<BackendState> =
+                controller.db.subscribe_with_key(backend_id.as_str());
+
+            // There is a race condition if the status updated between when our last query hit and when we started the
+            // subscription. It's a bit hacky, but for now we will just issue the query again.
+            // We can't start the subscription first to avoid repeating the query, because we need to know the backend
+            // ID to start the subscription.
+            if let Ok(RouteInfoResult::Ready(route_info)) =
+                controller.db.backend().route_info_for_token(&token).await
+            {
```
rolyatmax, line 58: It's also possible for the backend to have terminated in which case we should also drop out.

plane/src/controller/proxy.rs
```diff
@@ -19,31 +22,135 @@ use std::net::{IpAddr, SocketAddr};
...
+                    .await
+                {
+                    tracing::error!(?err, "Error sending route info response to proxy.");
+                }
+                return;
+            }
+
+            let socket = socket.sender(MessageToProxy::RouteInfoResponse);
+            tokio::spawn(async move {
+                while let Some(result) = sub.next().await {
```
rolyatmax, line 74: I think it would be a good idea to have a timeout here to prevent hanging forever and creating a memory leak (an issue we've had to track down and fix elsewhere)

plane/src/controller/proxy.rs
```diff
@@ -19,31 +22,135 @@ use std::net::{IpAddr, SocketAddr};
...
+                    let Notification { payload, .. } = result;
+
+                    match payload {
+                        BackendState::Ready { address } => {
```
rolyatmax, line 78: Why is `address` an `Option` again?

plane/src/controller/proxy.rs
paulgb: This will also catch that case (`ready` here means that the RouteInfoResult is ready, not that the backend itself is, so if the backend is terminated a "terminated" RouteInfoResult will be ready). The overlap with backend states is confusing, so I renamed Waiting/Ready it to Pending/Available.

rolyatmax: It looks like `route_info_for_token()` returns `Ok(RouteInfoResult::NotFound)` if the backend is terminated or terminating, though.

paulgb: Ah good catch, fixed.

plane/src/controller/proxy.rs
paulgb: No good reason! I opened #776 to refactor to require it.

rolyatmax (approved):

PR #776, opened 2024-07-08
Make ready address required

paulgb:
It doesn't make sense to have a ready backend without an address. This avoids that situation by passing the address through `to_ready`.

`BackendState` is sent across the network, so changes to it need to be version-safe. This change is version-safe:

- If a **writer** is on this version and a **reader** is on a prior version, nothing bad can happen. The reader continues to accept `null` a the value, but the writer will never produce that value.
- If a **reader** is on this version and a **writer** is on a prior version, there is technically a code path that produces a `null` here, but:
  - It's an error code path, and it logs a warning message
  - I could not find any instances of this in our logs

### DIFF
--- a/plane/src/drone/backend_manager.rs
+++ b/plane/src/drone/backend_manager.rs
@@ -178,7 +178,7 @@ impl<R: Runtime> BackendManager<R> {
                             TerminationReason::StartupTimeout,
                         )
                     } else {
-                        state.to_ready()
+                        state.to_ready(address)
                     }
                 })
             }
--- a/plane/src/drone/state_store.rs
+++ b/plane/src/drone/state_store.rs
@@ -258,7 +258,7 @@ mod test {
             .register_event(
                 &backend_id,
                 &BackendState::Ready {
-                    address: Some(dummy_addr()),
+                    address: dummy_addr(),
                 },
                 Utc::now(),
             )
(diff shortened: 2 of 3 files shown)


rolyatmax (approved):
Is it possible for there to be Ready states stored in the DB (on the `backend` or `backend_state` tables) without addresses? Will reading those states out of the DB after this lands cause runtime errors (either for us or for open source users of the project?)

paulgb:
> Is it possible for there to be Ready states stored in the DB (on the `backend` or `backend_state` tables) without addresses? Will reading those states out of the DB after this lands cause runtime errors (either for us or for open source users of the project?)

It's impossible for the current (pre-PR) code to produce a `None` here. The only call to `to_ready` is [here](https://github.com/jamsocket/plane/blob/f2d5e79266e67f282e6fa11eb6d0403c1b542536/plane/src/drone/backend_manager.rs#L181). It is called on `state`, and it is called within the branch of a `match state` that matches `BackendState::Waiting`.

Additionally, if it were to happen, a warn log is emitted. I haven't found any instances of in datadog. I also ran this query which returned no results:

```sql
SELECT * FROM "public"."backend_state" WHERE ("state"::TEXT LIKE '%"status": "ready"%') AND ("state"::TEXT LIKE '%null%') LIMIT 300 OFFSET 0;
```

rolyatmax:
Sounds good!

PR #778, opened 2024-07-09
Don't terminate if we already are

paulgb:
This prevents us from going from hard-terminating to soft-terminating, and from sending multiple SIGINTs.

### DIFF
--- a/plane/src/drone/backend_manager.rs
+++ b/plane/src/drone/backend_manager.rs
@@ -262,6 +262,18 @@ impl<R: Runtime> BackendManager<R> {
             .expect("State lock is poisoned")
             .state
             .clone();
+
+        if let BackendState::Terminating { termination, .. } = state {
+            if termination == TerminationKind::Hard {
+                // We are already hard-terminating, don't change the kind.
+                return Ok(());
+            }
+            if termination == TerminationKind::Soft && kind == TerminationKind::Soft {
+                // We are already soft-terminating, don't change the kind.
+                return Ok(());
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


plane/src/drone/backend_manager.rs
```diff
@@ -262,6 +262,18 @@ impl<R: Runtime> BackendManager<R> {
...
+
+        if let BackendState::Terminating { termination, .. } = state {
+            if termination == TerminationKind::Hard {
+                // We are already hard-terminating, don't change the kind.
+                return Ok(());
+            }
+            if termination == TerminationKind::Soft && kind == TerminationKind::Soft {
+                // We are already soft-terminating, don't change the kind.
+                return Ok(());
+            }
```
rolyatmax, line 274: We currently already have a check in `to_terminating` that makes sure a hard terminate can't turn into a soft terminate. Maybe it makes sense to protect against _sending_ more than one soft terminate, but I'm worried about dropping out here on a hard terminate. One reason is that the `set_state` call is one spot where we can actually detect that a hard terminating backend no longer exists and mark it as terminated. If a backend reaches a hard terminating state but we miss the event from docker to mark it as terminated, then I don't see a way the drone can ever mark the backend as terminated if we're dropping out here.

paulgb: Hmm good point, I'll mark this as a draft and have another think through it.

rolyatmax:
Have we observed a user-facing problem that we're trying to fix with this PR? Have we seen several sigints being sent to the same backend?

PR #779, opened 2024-07-09
Introduce SocketPath type with copy-safe drop

michaelsilver:
We want `server` to be copy safe: we do this successfully by wrapping the `_loop_task` in an `Arc` `GuardHandle`. However, `server`'s `Drop` will attempt to delete the `socket_path` file on every dropped copy -- which isn't the behavior we want. We only want that file deleted when the last copy is dropped. We achieve this similar to `_loop_task`: by creating a new `SocketPath` type (with its own `Drop` impl) and wrapping this type in an `Arc`.

### DIFF
--- a/plane/src/typed_unix_socket/mod.rs
+++ b/plane/src/typed_unix_socket/mod.rs
@@ -1,6 +1,8 @@
 use crate::util::ExponentialBackoff;
 use chrono::Duration;
 use serde::{Deserialize, Serialize};
+use std::fs;
+use std::path::PathBuf;
 
 pub mod client;
 pub mod server;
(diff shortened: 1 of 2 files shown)


paulgb (commented):
Good catch. How about wrapping up the behavior into a `struct SocketPath(PathBuf)`, and making `socket_path` an `Arc<SocketPath>`? That way you can avoid implementing `Drop` on `TypedUnixSocketServer`, and instead rely on a much simpler `Drop` implementation on `SocketPath`.

(You wouldn't need to check the reference count, since `Drop` on `SocketPath` will only be called after the last reference is dropped.)

michaelsilver:
Yep that's a much better approach, will do -- thanks!

michaelsilver:
Done -- ready for another look!

paulgb (approved):
Much simpler, love it!

PR #780, opened 2024-07-09
Allow drone to run as unix domain socket executor

michaelsilver:
Example usage:
```bash
$ ./dev/drone.sh --executor-socket "/tmp/tmp.sock"
```

This includes work from @paulgb to remove generics (#774) in favor of `Box<dyn ...>>` to enable dynamic dispatch.

### DIFF
--- a/plane/src/drone/command.rs
+++ b/plane/src/drone/command.rs
@@ -1,3 +1,4 @@
+use super::{runtime::unix_socket::UnixSocketRuntimeConfig, ExecutorConfig};
 use crate::{
     drone::{runtime::docker::DockerRuntimeConfig, DroneConfig},
     names::{DroneName, OrRandom},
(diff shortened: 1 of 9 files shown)


plane/src/drone/command.rs
```diff
@@ -10,7 +10,7 @@ use clap::Parser;
 use std::{net::IpAddr, path::PathBuf};
 use url::Url;
```
paulgb, line 12: nit: rm newline

plane/src/drone/command.rs
```diff
@@ -34,6 +34,10 @@ pub struct DroneOpts {
     #[clap(long)]
     docker_runtime: Option<String>,
 
+    // Optional path to a unix socket for connecting to an external executor.
```
paulgb, line 37: If you add a third `/`, this comment will be parsed as a docstring and be printed when running `plane --help`

paulgb (approved):
Looks good! Left a few comments.

plane/src/drone/command.rs
michaelsilver: sorry about this happening a lot, VSCode seems to do this automatically -- have you found a way to turn this off or are you just diligent about it?

PR #781, opened 2024-07-09
get_drones_for_pool() should return all drones + add index on drone(last_heartbeat)

rolyatmax:


michaelsilver (approved):

plane/schema/derived_schema.sql
```diff
@@ -952,6 +952,13 @@ CREATE INDEX idx_backend_static_token ON public.backend USING btree (static_toke
 CREATE UNIQUE INDEX idx_cluster_name ON public.node USING btree (cluster, name);
 
 
+--
+-- Name: idx_drone_last_heartbeat; Type: INDEX; Schema: public; Owner: postgres
+--
+
+CREATE INDEX idx_drone_last_heartbeat ON public.drone USING btree (last_heartbeat);
```
paulgb, line 959: This is a write-heavy column; I'm a bit concerned about having an index on it at scale because we'll be updating the index every drone heartbeat.

PR #783, opened 2024-07-10
Add batching to cleanup loop and clean up backend_key also

paulgb:
- Adds batching (default size 100).
- Sets the interval to once every 3 minutes (per controller) instead of once every 15 minutes
- Cleans up `backend_key` table too (as long as the keys are expired. This should always be the case, but we fail-safe by not deleting a non-expired key.)

### DIFF
--- a/plane/src/controller/command.rs
+++ b/plane/src/controller/command.rs
@@ -48,6 +48,7 @@ impl ControllerOpts {
             controller_url,
             default_cluster: self.default_cluster,
             cleanup_min_age_days: self.cleanup_min_age_days,
+            cleanup_batch_size: 100,
         })
     }
 }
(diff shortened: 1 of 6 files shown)


plane/src/controller/command.rs
```diff
@@ -48,6 +48,7 @@ impl ControllerOpts {
             controller_url,
             default_cluster: self.default_cluster,
             cleanup_min_age_days: self.cleanup_min_age_days,
+            cleanup_batch_size: None,
```
rolyatmax, line 51: I feel like it doesn't make sense to store this value on the ControllerConfig and then thread it through so many functions if ultimately it can _only_ be `None`. Maybe we just have `run_cleanup_loop()` call `run_cleanup(&db, min_age_days, None)`? (And the db cli would continue to call `run_cleanup(&db, min_age_days, cleanup_batch_size)`.) Thoughts?

paulgb: The `None` here only applies to the simplified option set available when configuring Plane via the CLI. When run in production, we configure with a `.json` file which can express the full range of `ControllerConfig` functionality.

rolyatmax: Ohhh, gotcha gotcha

rolyatmax (approved):

PR #784, opened 2024-07-11
Satisfy clippy

michaelsilver:
The only nontrivial change was switching from `std::sync::Mutex` to `tokio::sync::Mutex` because the mutex was held across calls to `.await`, following the [official guidance from Tokio](https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex-and-tokiosyncmutex).

### DIFF
--- a/plane/plane-tests/tests/common/test_env.rs
+++ b/plane/plane-tests/tests/common/test_env.rs
@@ -27,7 +27,7 @@ const TEST_CLUSTER: &str = "plane.test";
 #[derive(Clone)]
 pub struct TestEnvironment {
     pub scratch_dir: PathBuf,
-    db: Arc<Mutex<Option<DevDatabase>>>,
+    db: Arc<tokio::sync::Mutex<Option<DevDatabase>>>, // held across await, see: https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex-and-tokiosyncmutex
     drop_futures: Arc<Mutex<Vec<Arc<dyn AsyncDrop>>>>,
     log_subscription: Arc<Mutex<Option<LogSubscription>>>,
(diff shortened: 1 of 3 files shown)


plane/plane-tests/tests/common/test_env.rs
```diff
@@ -16,8 +16,9 @@ use plane::{
 use std::{
     net::{IpAddr, Ipv4Addr},
     path::{Path, PathBuf},
-    sync::{Arc, Mutex},
+    sync::Arc,
 };
+use tokio::sync::Mutex;
```
paulgb, line 21: switching the import makes all mutexes async, but it looks like only `db_lock` is held across an await, is that right?

paulgb (commented):
How did you run clippy so that you get these lint warnings? I'm trying to reproduce the errors here but it comes up clean on main.

paulgb (changes requested):
forgot to request changes

michaelsilver:
> How did you run clippy so that you get these lint warnings? I'm trying to reproduce the errors here but it comes up clean on main.

`cargo clippy --fix`

From help docs:
```
--fix                    Automatically apply lint suggestions. This flag implies --no-deps and --all-targets
```

So the problem is you were running `cargo clippy` when instead you should run `cargo clippy --all-targets`

plane/plane-tests/tests/common/test_env.rs
paulgb: confirmed, let's just use `tokio::sync::Mutex` on 30 and keep the standard mutex import for the rest.

michaelsilver: Fixed

paulgb (approved):

PR #787, opened 2024-07-18
Export entity_name macro for external use

michaelsilver:


plane/src/names.rs
```diff
@@ -3,7 +3,7 @@ use clap::error::ErrorKind;
 use serde::{Deserialize, Serialize};
 use std::fmt::{Debug, Display};
 
-const MAX_NAME_LENGTH: usize = 45;
+pub const MAX_NAME_LENGTH: usize = 45;
```
paulgb, line 6: Now that I look at it, I’m confused why this works (even before your changes). If this is scoped to the file, why can we refer to it in a macro?

I’ll play a bit with this code tomorrow just to make sure I understand it.

paulgb (approved):

PR #789, opened 2024-07-20
Handle removed backend

paulgb:
When a keepalive request fails because the backend has been removed, we should tell the proxy that the backend has been removed.

Fixes DIS-2106

### DIFF
--- a/plane/src/controller/proxy.rs
+++ b/plane/src/controller/proxy.rs
@@ -178,7 +178,11 @@ pub async fn handle_message_from_proxy(
             if let Err(err) = controller.db.backend().update_keepalive(&backend_id).await {
                 tracing::error!(?err, ?backend_id, ?node_id, "Error updating keepalive");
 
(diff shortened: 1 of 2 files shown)


plane/src/controller/proxy.rs
```diff
@@ -176,7 +176,13 @@ pub async fn handle_message_from_proxy(
         }
         MessageFromProxy::KeepAlive(backend_id) => {
             if let Err(err) = controller.db.backend().update_keepalive(&backend_id).await {
-                tracing::error!(?err, ?backend_id, "Error updating keepalive");
+                tracing::error!(?err, ?backend_id, ?node_id, "Error updating keepalive");
+
+                socket
```
rolyatmax, line 181: Maybe we should only send this if the error is `sqlx::Error::RowNotFound`?

paulgb: Good call. Updated.

rolyatmax (approved):

PR #791, opened 2024-07-22
Split `Terminating` into (soft) `Terminating` and `HardTerminating`

paulgb:
Currently, we use `Terminating` to represent both soft and hard terminating. If a backend is hard-terminated while it is being soft-terminated (which is allowed), a backend goes from a `Terminating` state to another `Terminating` state. This is the only allowed state transition that does not move forward in the state sequence.

The change in this PR is to split soft- and hard- `Terminating` states into their own separate states. Soft-termination keeps the name `Terminating`, and hard termination gets `HardTerminating`. The field `termination` in `Terminating` remains, but is deprecated, and can only be set to `Soft` in the constructor.

### DIFF
--- a/plane/src/types/backend_state.rs
+++ b/plane/src/types/backend_state.rs
@@ -6,7 +6,6 @@ use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::{fmt::Display, net::SocketAddr};
-use valuable::Valuable;
 
 #[derive(Clone, Copy, Serialize, Deserialize, Debug, PartialEq, PartialOrd)]
 #[serde(rename_all = "lowercase")]
@@ -32,6 +31,10 @@ pub enum BackendStatus {
     /// Proxies should stop sending traffic to it, but we should not yet release the key.
     Terminating,
 
+    /// The backend has been sent a SIGKILL, either because the user sent a hard termination
+    /// request or the lock was past the hard-termination deadline.
+    HardTerminating,
+
     /// The backend has exited or been swept.
     Terminated,
 }
@@ -50,6 +53,7 @@ impl BackendStatus {
             BackendStatus::Waiting => 40,
             BackendStatus::Ready => 50,
             BackendStatus::Terminating => 60,
+            BackendStatus::HardTerminating => 65,
             BackendStatus::Terminated => 70,
         }
     }
@@ -64,6 +68,7 @@ impl valuable::Valuable for BackendStatus {
             BackendStatus::Waiting => valuable::Value::String("waiting"),
             BackendStatus::Ready => valuable::Value::String("ready"),
             BackendStatus::Terminating => valuable::Value::String("terminating"),
+            BackendStatus::HardTerminating => valuable::Value::String("hard-terminating"),
             BackendStatus::Terminated => valuable::Value::String("terminated"),
         }
     }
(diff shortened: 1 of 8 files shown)


plane/src/types/backend_state.rs
```diff
@@ -393,7 +427,8 @@ impl BackendStatusStreamEntry {
 
         let termination_kind = match state {
             BackendState::Terminated { termination, .. } => termination,
-            BackendState::Terminating { termination, .. } => Some(termination),
+            BackendState::Terminating { .. } => Some(TerminationKind::Soft),
+            BackendState::HardTerminating { .. } => Some(TerminationKind::Hard),
```
rolyatmax, line 431: Do we need to pull out the `termination_reason` for `HardTerminating` a few lines above, too?

plane/src/types/backend_state.rs
```diff
@@ -273,50 +304,53 @@ impl BackendState {
...
                 last_status: self.status(),
-                termination,
                 reason,
             },
         }
     }
 
+    pub fn to_terminating(&self, reason: TerminationReason) -> BackendState {
+        if self.status() >= BackendStatus::Terminating {
+            tracing::warn!(?reason, state=?self, "to_hard_terminating called on backend in later state.");
```
rolyatmax, line 327: I think this log message is incorrect. It should say `to_terminating called...`

rolyatmax:
I think we need to update this line, too: https://github.com/jamsocket/plane/blob/main/plane/src/controller/proxy.rs#L128

plane/src/types/backend_state.rs
```diff
@@ -273,50 +304,53 @@ impl BackendState {
...
+            },
+            _ => BackendState::HardTerminating {
                 last_status: self.status(),
-                termination,
                 reason,
             },
         }
     }
 
+    pub fn to_terminating(&self, reason: TerminationReason) -> BackendState {
```
rolyatmax, line 325: Nit: maybe put `to_terminating` before `to_hard_terminating` so all these status methods are in order

rolyatmax:
Should update https://github.com/jamsocket/plane/blob/main/docs/pages/concepts/backend-lifecycle.mdx

rolyatmax:
Also https://github.com/jamsocket/plane/blob/main/plane/src/database/backend.rs#L320 and https://github.com/jamsocket/plane/blob/main/plane/src/database/backend.rs#L395

rolyatmax:
One thought: there will be a small window of time during a deploy where it'll be possible for p2 to receive unparseable BackendState update messages. That's only for messages that have a HardTerminating BackendStatus, though, which I'm guessing are rare. If we really wanted to avoid this, we could put the type changes in their own commit that comes first in the PR, then update plane version to that specific commit, deploy, then update to the latest commit and deploy again.

paulgb:
> One thought: there will be a small window of time during a deploy where it'll be possible for p2 to receive unparseable BackendState update messages. That's only for messages that have a HardTerminating BackendStatus, though, which I'm guessing are rare. If we really wanted to avoid this, we could put the type changes in their own commit that comes first in the PR, then update plane version to that specific commit, deploy, then update to the latest commit and deploy again. It's kind of a hassle though, and might not really be worth it.

Terminating statuses are only created on the drone, so as long as we deploy the controller before the drones we should be fine.

rolyatmax (approved):

PR #793, opened 2024-07-24
Migrate off storing `BackendName` in Docker labels

michaelsilver:
Instead of storing a label like:
 - `"dev.plane.backend"`: `"xjbgb03bhfxqgp"`

We will now store a label like:
 - `"dev.plane.backend"`: `"true"`

Critically, this allows us to continue relying on
`"dev.plane.backend"` as an _identifier_ for plane backends, while deprecating the reliance on that label for carrying the `BackendName`.

The only place we relied on the `BackendName` carried by this label was in the events loop, where as it turns out docker already applies a `"name"` attribute to events.

Otherwise, introduce some helper functions to idiomatically convert between `ContainerId` and `BackendName`.


rolyatmax (approved):
Nice! Happy you were able to make plane work with just static labels!

paulgb (approved):

paulgb (changes requested):
LGTM but let's discuss before landing re. some other implications.

paulgb (approved):
Thanks for the log collector fix.

PR #799, opened 2024-08-06
Handle `BackendError::Other` in `wait_for_backend()`

michaelsilver:
https://github.com/jamsocket/plane/pull/766 introduced `BackendError::Other` as a way to handle misc errors that may occur in the `wait_for_backend()` executor function. However, the PR failed to handle that error type in `BackendManager`. As a result, spawns would succeed in spite of that error.

Introduce `TerminationReason::ErrorWaiting` and hard terminate any backend that encounters `BackendError::Other` in the `wait_for_backend()` step, logging the specifics. (Use a `match` statement to ensure future additions to the error struct get caught in this codepath during compilation.)


paulgb (approved):
Good catch on switching to `match` to avoid it happening again!

PR #800, opened 2024-08-06
Improve typed unix socket logging

michaelsilver:
Helpful to see the message in context of the error.

### DIFF
--- a/plane/src/typed_unix_socket/client.rs
+++ b/plane/src/typed_unix_socket/client.rs
@@ -119,7 +119,7 @@ async fn connect<P: AsRef<Path>>(socket_path: P) -> UnixStream {
                 return stream;
             }
             Err(e) => {
-                tracing::error!("Error connecting to server: {}", e);
+                tracing::error!(%e, "Error connecting to server.");
                 backoff.wait().await;
             }
         }
@@ -156,7 +156,7 @@ where
                         {
                             Ok(msg) => msg,
                             Err(e) => {
-                                tracing::error!(?line, "Error deserializing message: {}", e);
+                                tracing::error!(%e, ?line, "Error deserializing message.");
                                 continue;
                             }
                         };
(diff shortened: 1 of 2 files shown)


plane/src/typed_unix_socket/client.rs
```diff
@@ -163,16 +166,21 @@ where
...
+                                        ?message,
+                                        "No sender found for response ID."
+                                    );
                                 }
                             }
                             WrappedMessage { id: None, message } => {
-                                if let Err(e) = event_tx.send(message) {
-                                    tracing::error!("Error sending event: {}", e);
+                                if let Err(e) = event_tx.send(message.clone()) {
+                                    tracing::error!(?message, "Error sending event: {}", e);
```
paulgb, line 183: message is a datadog [reserved attribute](https://docs.datadoghq.com/logs/log_configuration/attributes_naming_convention/#reserved-attributes) so we should call it something else.

Also, rather than splicing `e` into the error message, I prefer to use structured fields like this:

```rust
tracing::error!(message=?message, ?e, "Error sending event");
```

plane/src/typed_unix_socket/server.rs
```diff
@@ -142,19 +142,19 @@ where
...
-                                    tracing::error!("Error sending request: {}", e);
+                                if let Err(e) = request_tx.send(msg.clone()) {
+                                    tracing::error!(?msg, "Error sending request: {}", e);
                                 }
                             }
                             WrappedMessage { id: None, message } => {
-                                if let Err(e) = event_tx.send(message) {
-                                    tracing::error!("Error sending event: {}", e);
+                                if let Err(e) = event_tx.send(message.clone()) {
+                                    tracing::error!(?message, "Error sending event: {}", e);
```
paulgb, line 157: same as above (`message` is reserved)

paulgb (changes requested):
Just a nit, otherwise lgtm.

michaelsilver:
@paulgb ready for another look. Sorry for the delay here, I wanted to figure out how to make the errors print using `Display` instead of `Debug` and stumbled upon `%` while working on other stuff. It turns out the `Display` versions of these errors tend to be more helpful!

paulgb (approved):

PR #802, opened 2024-08-13
Rename TerminationReason::ErrorWaiting => InternalError

michaelsilver:
Renaming error status introduced in https://github.com/jamsocket/plane/pull/799 so that it's more user friendly.


rolyatmax (approved):
I think this makes sense - thanks!

PR #804, opened 2024-08-29
Allow Plane to delegate auth

paulgb:
Adds `forward_auth` config option (name borrowed from [Caddy](https://caddyserver.com/docs/caddyfile/directives/forward_auth)). If provided, requests to `/ctrl/*` are forwarded to that URL first (after stripping the body). If that URL returns a 200, Plane proceeds; otherwise they are rejected.


rolyatmax (approved):
LGTM!

PR #808, opened 2024-09-10
Add health endpoint

paulgb:
We currently use `/ctrl/status` as a health check endpoint. The introduction of auth delegation (#804) applies, when enabled, to all `/ctrl` endpoints, including the health check. That means that the health checker needs to have credentials.

Instead, this adds a separate health check endpoint `/pub/health` which is part of the `/pub` path, which is not credentialed. This does not include the version information returned by the `/ctrl/status` endpoint, but does a health check.


rolyatmax (approved):
LGTM - failing test though

PR #810, opened 2024-09-10
Improve typed socket connection robustness

paulgb:
What we know:

- Sometimes, when there are network issues, some (but not all) drones fail to reconnect. In the logs, it shows up as `failed to send heartbeat` (from [heartbeat.rs](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/drone/heartbeat.rs#L19)), with the `err` value of `Disconnected` ([typed_socket](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/typed_socket/mod.rs#L45)).
- When we “send” websocket messages, we are actually sending messages to a queue that gets picked up by the websocket event loop asynchronously. Disconnected [actually means that the message queue is full](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/typed_socket/mod.rs#L51), which *implies* we are disconnected, but is not immediate.
- On a `TypedSocket`, `send()` [uses Sender::send](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/typed_socket/mod.rs#L70-L76), while `TypedSocketSender`'s `send()` uses [try_send](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/typed_socket/mod.rs#L92).
- In tokio, [`send`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Sender.html#method.send) on a full channel blocks until the channel has capacity. `try_send` returns immediately if the channel is full.

My leading theory is that when the network is interrupted abruptly, upon reconnecting to the controller, the controller sends a bunch of messages, causing a deadlock:

- the `new_client` loop is stalled waiting for capacity in send_to_client, which won't happen until `socket.recv()` [is called](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/drone/mod.rs#L96-L99) in the main drone event loop
- the main drone event loop is waiting on a call to `socket.send()` when [acking an action](https://github.com/jamsocket/plane/blob/afc9b7f0786f69770fb9fe4b9731fde566dc793d/plane/src/drone/mod.rs#L142)

This PR introduces several changes, which should improve the robustness of reconnects:

- Instead of handling messages from the controller directly in the drone event loop, they are sent to separate tasks. This means that nothing can get in the way of the drone event loop's ability to call `socket.recv()`.
- `TypedSocket::send` now uses `Sender::try_send`, for consistency with `TypedSocketSender::send`. Note that as a result of moving handling out of the drone event loop, messages sent from the drone loop now use `TypedSocketSender` instead of `TypedSocket` anyway, so those messages would now use `try_send` regardless.
- As a result of using `try_send`, `TypedSocket::send` no longer needs to be async, so that's removed.
- `TypedSocketError::Disconnected` is renamed to `TypedSocketError::Clogged`, to better reflect what the issue is.

### DIFF
--- a/plane/src/typed_socket/mod.rs
+++ b/plane/src/typed_socket/mod.rs
@@ -42,13 +42,13 @@ pub enum TypedSocketError {
     #[error("Socket closed")]
     Closed,
     #[error("Socket disconnected")]
-    Disconnected,
+    Clogged,
 }
 
 impl<A> From<TrySendError<A>> for TypedSocketError {
     fn from(e: TrySendError<A>) -> Self {
         match e {
-            TrySendError::Full(_) => Self::Disconnected,
+            TrySendError::Full(_) => Self::Clogged,
             TrySendError::Closed(_) => Self::Closed,
         }
     }
--- a/plane/src/typed_socket/mod.rs
+++ b/plane/src/typed_socket/mod.rs
@@ -67,12 +67,10 @@ impl<A: Debug> TypedSocketSender<A> {
 }
 
 impl<T: ChannelMessage> TypedSocket<T> {
-    pub async fn send(&mut self, message: T) -> Result<(), PlaneClientError> {
-        if let Err(e) = self.send.send(SocketAction::Send(message)).await {
-            tracing::error!(?e, "Failed to send message on websocket. Closing receiver.");
-            self.recv.close();
-            return Err(PlaneClientError::SendFailed);
-        }
+    pub fn send(&mut self, message: T) -> Result<(), PlaneClientError> {
+        self.send
+            .try_send(SocketAction::Send(message))
+            .map_err(|_| PlaneClientError::SendFailed)?;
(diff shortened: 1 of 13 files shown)


plane/src/typed_socket/mod.rs
```diff
@@ -42,13 +42,13 @@ pub enum TypedSocketError {
     #[error("Socket closed")]
     Closed,
     #[error("Socket disconnected")]
```
rolyatmax, line 44: "Socket clogged"?

rolyatmax (approved):
LGTM

PR #828, opened 2024-10-16
Require minimum token length when parsing connection token from request

paulgb:
Sets a minimum token length when parsing, mostly to avoid vulnerability scanners from clogging the LRU cache. (Note that this is _not_ protection against someone intentionally clogging the LRU cache).

image

This does not change the length of generated tokens, which is currently about 43 bytes (varies slightly depending on the number of leading zeros in the number it encodes).


rolyatmax (approved):
Good idea!

PR #838, opened 2024-11-05
drone failed to mark backend as terminated

rolyatmax:
We've observed failure modes where a container exit is not detected. In those cases, we expect that a termination request will see that the container is no longer running due to docker returning an error when calling `stop_container`/`kill_container`. Docker _doesn't_ error when calling `stop_container()` on an already stopped container, however. This has caused termination loops that never fully succeed, leaving the drone thinking the container is still running, and re-renewing its key over and over. This PR fixes this issue by checking the container's status before calling `terminate()` on it.

This still doesn't address these two issues, however:
(1) we aren't detecting the container exit in some cases which causes this failure mode in the first place
(2) we should make `terminating` loops eventually resort to `hard-terminating`

### DIFF
--- a/plane/src/drone/executor.rs
+++ b/plane/src/drone/executor.rs
@@ -177,8 +177,8 @@ impl Executor {
                 let callback = {
                     let state_store = self.state_store.clone();
                     let backend_id = backend_id.clone();
-                    let timestamp = chrono::Utc::now();
                     move |state: &BackendState| {
+                        let timestamp = chrono::Utc::now();
                         state_store
                             .lock()
                             .expect("State store lock poisoned.")
--- a/plane/src/drone/runtime/docker/mod.rs
+++ b/plane/src/drone/runtime/docker/mod.rs
@@ -194,6 +194,37 @@ impl Runtime for DockerRuntime {
     async fn terminate(&self, backend_id: &BackendName, hard: bool) -> Result<bool, anyhow::Error> {
         let container_id: ContainerId = backend_id.into();
 
+        // check if container is no longer running, since stop_container() returns Ok(()) even when the container is already gone
+        match self
+            .docker
+            .inspect_container(&container_id.to_string(), None)
+            .await
+        {
+            Ok(details) => {
+                if let Some(state) = details.state {
+                    if !state.running.unwrap_or(false) {
+                        tracing::warn!(
+                            %container_id,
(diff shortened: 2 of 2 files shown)


plane/src/drone/executor.rs
```diff
@@ -177,8 +177,8 @@ impl Executor {
                 let callback = {
                     let state_store = self.state_store.clone();
                     let backend_id = backend_id.clone();
-                    let timestamp = chrono::Utc::now();
                     move |state: &BackendState| {
+                        let timestamp = chrono::Utc::now();
```
rolyatmax, line 181: Assuming this is a bug causing all events stored in the state store to have the timestamp of when this backend was created, not when the callback was called.

plane/src/drone/runtime/docker/mod.rs
```diff
@@ -220,16 +251,6 @@ impl Runtime for DockerRuntime {
                 );
                 Ok(false)
             }
-            Err(bollard::errors::Error::DockerResponseServerError {
-                status_code: 404, ..
-            }) => {
```
rolyatmax, line 225: Moving this check up to the inspect container call because we expect to catch a 404 there first.

plane/src/drone/runtime/docker/mod.rs
```diff
@@ -194,6 +194,37 @@ impl Runtime for DockerRuntime {
...
+                if let Some(state) = details.state {
+                    if !state.running.unwrap_or(false) {
+                        tracing::warn!(
+                            %container_id,
+                            %backend_id,
+                            "Container could not be terminated, because it is not running."
+                        );
+                        return Ok(false);
+                    }
+                }
```
michaelsilver, line 213: I'm not a fan of nested `if` or `if lets`, find this hard to read. 

How about instead:
```Rust
if !details.state.and_then(|s| s.running).unwrap_or(false) {
    tracing::warn!(
        %container_id,
        %backend_id,
        "Container could not be terminated, because it is not running."
    );
    return Ok(false);
}
```

michaelsilver (approved):
LGTM; left one style nit

plane/src/drone/runtime/docker/mod.rs
rolyatmax: Updated!

paulgb (approved):

PR #858, opened 2025-01-06
fix mem leak - drop the JoinSets in the dynamic proxy connection loops

rolyatmax:
Ran locally and watched memory usage with `htop` while making a stream of requests to the proxy. Before this change, memory continued to increase with each request. After this change, memory rose slightly but then plateaued indefinitely.

### DIFF
--- a/dynamic-proxy/src/server.rs
+++ b/dynamic-proxy/src/server.rs
@@ -31,12 +31,7 @@ pub struct SimpleHttpServer {
     graceful_shutdown: Option<GracefulShutdown>,
 }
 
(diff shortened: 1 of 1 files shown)


dynamic-proxy/src/server.rs
```diff
@@ -36,14 +36,13 @@ async fn listen_loop<S>(
     listener: TcpListener,
     service: S,
     graceful_shutdown: GracefulShutdown,
-) -> JoinSet<()>
+) -> ()
```
paulgb, line 39: nit: we can omit the `-> ()` here

paulgb (approved):
LGTM, thanks for this! one nitpick which is not blocking.

